PR #483, opened 2013-01-17
allow get_class to import from modules not present in oscar

marekbrzoska:
@gissues:{"order":85.71428571428571,"status":"backlog"}

### DIFF
--- a/oscar/core/loading.py
+++ b/oscar/core/loading.py
@@ -13,11 +13,11 @@ class ClassNotFoundError(Exception):
     pass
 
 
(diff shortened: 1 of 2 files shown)


v1kku:
@codeinthehole @mbertheau Was there a reason why this feature was dropped. There are some use cases where this could be useful. I've done the same work [on this commit](https://github.com/v1kku/django-oscar/commit/f4b5e4b7f525d19e232c6886e3f970791b2e6fb3) and was wondering whether there was/is any reason not to have this in oscar?

PR #856, opened 2013-09-27
Improve product image handling

maikhoepfel:
There's currently a few issues with the way the dashboard handles product images.

https://github.com/tangentlabs/django-oscar/blob/master/oscar/templates/oscar/partials/image_input_widget.html
- [ ] If no image is set, an empty `` is rendered. It's not visible as it's overlaid by both a button and the <input> field, but at least in Chrome changes the mouse cursor
- [x] The inline(!) javascript uses relative addressing via `.next()` to hide or show the upload button
- [ ] Semantically, the button is never used, and merely for display purposes. Maybe one could remove it, and instead have an always visible `` which looks something like this
  button
- [ ] As everything else has an immediate reaction, it would be nice if deleting pictures would be instant as well. Or at least the image becoming grayed out when clicking 'Delete'
- [x] The input field is, at least in Chrome, 100 px to the right. That causes the mouse cursor to not display as expected in the left half of the area. This is caused by absolute positioning and some not widely supported transforming: https://github.com/tangentlabs/django-oscar/blob/master/oscar/static/oscar/less/dashboard.less#L771
- [ ] It would be nice if images could be dragged & dropped. Seems easy enough: http://djangosnippets.org/snippets/1053/
- [ ] ImageField validation errors need looking into, as per @mbertheau's comment below
- [ ] Uploaded images aren't saved if other validation fails, as reported on the [mailing list](https://groups.google.com/d/msgid/django-oscar/a1bf22d9-d674-43c9-a88b-3d6382a8c174%40googlegroups.com) and in https://github.com/tangentlabs/django-oscar/pull/1126
- [ ] Allow associating images with child products (as mentioned in https://github.com/tangentlabs/django-oscar/issues/675#issuecomment-48340506)

Possible enhancements:
- Allow image cropping with e.g. [django-image-cropping](https://github.com/jonasundderwolf/django-image-cropping)

### DIFF
--- a/oscar/static/oscar/css/dashboard.css
+++ b/oscar/static/oscar/css/dashboard.css
@@ -2306,10 +2306,7 @@ caption > i,
   opacity: 0;
 }
 .upload-image .input-field input {
-  position: absolute;
-  border-width: 0 0 100px 200px;
-  filter: alpha(opacity=0);
-  -moz-transform: translate(-100px, 0) scale(1);
+  border-width: 0;
   direction: ltr;
   cursor: pointer;
   width: 200px;
--- a/oscar/static/oscar/less/dashboard.less
+++ b/oscar/static/oscar/less/dashboard.less
@@ -769,10 +769,7 @@ caption,
     opacity: 0;
   }
   .input-field input {
-    position: absolute;
-    border-width: 0 0 100px 200px;
-    filter: alpha(opacity=0);
-    -moz-transform: translate(-100px, 0) scale(1);
+    border-width: 0;
     direction: ltr;
     cursor: pointer;
     width: 200px;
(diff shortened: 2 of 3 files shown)


aef-:
I'd like to integrate https://github.com/flowjs/flow.js. This will fix the following on your list:

* No Image is set, empty image is shown
* Semantically, the button is never used, and merely for display purposes. Maybe one could remove it, and instead have an always visible  which looks something like this 
* It would be nice if images could be dragged & dropped

And add the ability to drag and drop folders (which is why I need to integrate this) so that a user no longer has to save 4 images and wait for 4 more input boxes to appear.

ImageField error validation, deletion and drag and drop for ordering can be explored.

But I'm asking here to see whether this would be pulled in or if you guys have another direction you'd rather take.

PR #1252, opened 2014-03-12
Don't overwrite bootstrap error, success and information alert colors

mbertheau:
http://getbootstrap.com/2.3.2/components.html#alerts

git blame doesn't reveal any useful information as to why this was added in
the first place.

(Edit: added screenshots)

Before:
before
After:
after

### DIFF
--- a/oscar/static/oscar/css/dashboard.css
+++ b/oscar/static/oscar/css/dashboard.css
@@ -3325,11 +3325,6 @@ caption > i,
 .promotions .well-info {
   background-color: #ffffff;
 }
-.alert {
-  background-color: #fff49f;
-  border-color: #ede181;
-  color: #333;
-}
 #order-graph {
   position: relative;
   margin-bottom: 5em;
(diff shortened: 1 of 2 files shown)


john-parton:
I would rather have the default bootstrap colors. I know I can override it in my app, but what is the rationale behind stripping the color?

The regression is super easy to fix:
Remove:
```css
.alert {
  background-color: #fff49f;
  border-color: #ede181;
  color: #333;
}
```
and add:
```css
.alert .btn {
  color: #333;
}
```

mbertheau:
(The changes are now on https://github.com/machtfit/django-oscar/commits/css-changes. I hope I get around to some dashboard CSS cleanup in or after #1576.)

john-parton:
`master` doesn't need the customization to the alert anymore, because the buttons have a color manually applied with a bootstrap `btn` class.

screenshot from 2015-10-20 16 05 04

PR #1351, opened 2014-06-03
Restructure docs

codeinthehole:
*This is a work-in-progress*

I think the left-hand nav for Oscar's docs could be improved considerably (and the docs as a whole). At the moment, it just stacks up the various TOCs from `index.rst`. We can improve this by using a `:hidden:` TOC to hand-pick the appropriate links and just use normal lists of links for the current TOCs.

Not quite sure on what the right nav structure should be. Maybe something like:

- Is Oscar right for me? (description of what Oscar is, design decisions)
- Getting started (oscar sample sites, installing locally)
- Customisation (eg how to customise apps, URLs, models, views etc)
- Topics (move all the recipes into a more coherent tree structure)
- Reference (settings, signals, template tags, management commands)
- Deploying, upgrading etc
- Support
- Upgrading 
- Extensions
- Contributing
- Releases


SalahAdDin:
Why did you close this?

sasha0:
Pull request was outdated and labeled accordingly. Let's not start new unrelated discussions in the closed pull requests.

Thank you.

PR #1366, opened 2014-06-11
Docker!

codeinthehole:
☢☢☢ *This is a work-in-progress* ☢☢☢

This PR adds support for creating a docker container for Oscar's sandbox. Running `docker/build.sh` should create a container tagged `oscar_latest` that has the sandbox installed. You can then run Django's server within the container using:
```bash
$ docker run -it -p 8000:8000 /code/sites/sandbox/manage.py runserver 0.0.0.0:8000
```
and browse it on your host machine. OSX users will have to forward port 8000 of the boot2docker VM too:
```bash
$ boot2docker ssh -L 8000:localhost:8000
```
Double port-forwarding! Yay

I'm relatively new to Docker so feedback is welcome.


knicholes:
As per the [docker best practices](https://docs.docker.com/articles/dockerfile_best-practices/#run-only-one-process-per-container) it'd be nice if you separated each process into its own container then linked them all together using docker-compose.

john-parton:
I agree with @knicholes. It's strange seeing so much stuff crammed into one container.

My app is built with docker. This isn't it exactly, obviously, but an okay starting guide:

```yml
# docker-compose.yml
db:
  image: postgres 
  volumes:
    - /var/lib/postgresql/pillowcentral:/var/lib/postgresql/data
memcached:
  image: memcached
elasticsearch:
  image: elasticsearch
nginx:
  build: deploy/nginx
  ports:
    - "8051:80"
    - "8052:443"
  volumes:
    - /var/www//example.com:/var/www/example.com
  links:
    - oscar
oscar:
  build: .
  command: gunicorn --config=gunicorn.py project.wsgi
  volumes:
    - /var/www/example.com:/var/www/example.com # for collectstatic and saving media
  links:
    - db
    - memcached
    - elasticsearch
```

```Dockerfile
# Dockerfile
FROM python:2.7
ENV PYTHONUNBUFFERED 1

RUN mkdir /code
WORKDIR /code

COPY requirements.txt /code/

RUN pip install -r requirements.txt

COPY code /code

RUN python manage.py collectstatic --noinput
```

codeinthehole:
@knicholes @john-parton Yes, I'm sure you're right. This PR was really me learning Docker back in 2014. Things have moved on since then but this PR hasn't.

I'm really sure it's worth having an official images for Oscar. It was a fun exercise but I'm not convinced they are that useful in general.

sasha0:
@john-parton @knicholes mb we could finalize this PR?

john-parton:
I actually have a dockerized oscar app running on top of Amazon's EC2 container service. It was a massive pain to set up, and there are some caveats, but it seems pretty solid so far.

mvantellingen:
I've just merged a simple implementation of docker for the sandbox via #1915 and closing this one. Please let me know if you don't agree with me :-)

PR #1298, opened 2014-06-12
Allow non-linear prices

maikhoepfel:
Oscar defaults implementation makes assumption that the price is computed as:

```
line_price = quantity * unit price
```

That is not always a case. For example, you might want to implement bulk prices:
- 1-99 items... unit price of X
- 100-999 items... unit price of Y

This could be easily implemented by adding additional methods to Price object:
- `price_for_quantity(quantity)` (effective one)
- `price_excl_tax_for_quantity(quantity)`
- `price_incl_tax_for_quantity(quantity)`

Default implementation could reflect the current one, so it is backwards compatible.

### DIFF
--- a/oscar/apps/analytics/receivers.py
+++ b/oscar/apps/analytics/receivers.py
@@ -51,26 +51,28 @@ def _record_products_in_order(order):
 
 
 def _record_user_order(user, order):
-    try:
-        record = UserRecord.objects.filter(user=user)
-        affected = record.update(
-            num_orders=F('num_orders') + 1,
-            num_order_lines=F('num_order_lines') + order.num_lines,
-            num_order_items=F('num_order_items') + order.num_items,
-            total_spent=F('total_spent') + order.total_incl_tax,
-            date_last_order=order.date_placed)
-        if not affected:
-            UserRecord.objects.create(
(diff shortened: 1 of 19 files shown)


mvantellingen:
@maikhoepfel mind if I continue your work on this? Currently implementing this in a a new project based on this PR. If that works out fine i'll try to update this PR with lessons learned

john-parton:
@vovanbo Why does your model have a `ForeignKey` to `Product`? Isn't the product accessible with `price_for_quantity.stockrecord.product`?

Why is `price_excl_tax` nullable?

I'm just curious as to why you made those design decisions. Thanks for supplying your model.

sasha0:
@mvantellingen could we now use your solution to finalize this pull request? Thanks.

mvantellingen:
No sorry. It isn't in any state now to just backport. Besides that, and this might sound a bit harsh, but it isn't something I myself currently need. So unless someone actually pays me to implement this it isn't going to happen anytime soon by me.

solarissmoke:
I'm closing this as there has been no progress since 2014, and there are enough conflicts that you'd probably want to start again anyway. If anyone is willing/able to work on this then please make a fresh PR.

Issue #1398, opened 2014-07-07
Applying multiple benefits per basket line


mbertheau:
(#1414 is the corresponding PR)
## Requirements

We need multiple benefits per basket line for our implementation. Two offers we want to combine are:
1. a usually time-based offer like "If Germany wins, next day 15% off all products"
2. a custom offer where every user gets a 50% discount up to 150 Euros per year paid by his employer

On our site it is so common that people buy only quantity 1 of one product that we will skip the UI
part of the basket functionality. Both offers should apply to the same product. For example if a
product costs 100 Euros and Germany won yesterday (which it did!) the first offer is active and the
price is 85 Euros. Then the other offer applies and the user gets a discount of 50% on the remaining
price. The end price is then 42,50 Euros.
### Additional requirements
- https://groups.google.com/d/msg/django-oscar/83-KLHQwSfE/CQkgQvzxwU0J
- https://groups.google.com/forum/#!topic/django-oscar/wQLsQqK-qF8
## Current situation in Oscar

To implement offers oscar has the concept of conditions and benefits. An example condition is "buy
two products from category A". An example benefit is "get 50% off a product from category B".

Oscar takes care to consume the products mentioned in the condition and in the benefit so that
1. the same two products from category A don't allow a discount for a second product from category B
2. the two products from category A can't be discounted
3. the same product from category B that was discounted cannot be discounted again
4. the product from category B cannot be used as a condition for another benefit

It does so by using `Line._affected_quantity` to track what quantity of the line was used in
conditions (`Line.consume()`) or benefits (`Line.discount()`). Both methods use the same counter.

When testing a condition oscar looks in the basket whether there are lines that weren't used in
another condition or discounted by a benefit by comparing `Line.quantity` and
`Line._affected_quantity`.
### Shortcomings
- Mixing the tracking of condition-consumed and benefit-consumed items leads to problems such as
  https://groups.google.com/forum/#!topic/django-oscar/vceEpy889O4.
- `CountCondition` and `CoverageCondition` consume lines inconsistently. `CountCondition` discards
  lines with a zero price, `CoverageCondition` does not.
## The plan

In order to fix the shortcomings of the current implementation and also realize the new requirements
I intend to
1. (Refactoring) Separately track which line items were used in which condition and which were used
   in which benefit. Details on semantics below.
   1. Use an extra data structure instead of basket lines to track that (anticipating
      https://groups.google.com/forum/#!topic/django-oscar/nr5QFw5tPII).
[...]

PR #1414, opened 2014-07-16
Refactor offers

mbertheau:
This is my offer refactoring according to the revised plan in #1398. It builds on the so far unmerged changes from #1392. It is not ready for merge into oscar because all conditions and all but one benefit are ripped out. It can serve as a starting point to implement the missing condition and benefit types on the refactored base code.

Some commits however can probably be cherry-picked into master:
- c7a54e0
- f53d1fa
- be20ee5
- b1ecd3e
- e8f7b1d


Mohl:
Is there any progress on this?

The problem is that neither the documentation nor the inline docstrings indicate that only one offer can be applied to a single line (rather the opposite is the case), which sucks, because you either know the code (which I do now) or your customers find it out the hard way (which they did).

sasha0:
Looks interesting as a starting point for discounting improvement, however quite outdated, has conflicts and requires more work to be done until this refactoring is complete. Nonetheless, we might use this approach during next offer revision and include multiple benefits per line to the list of improvements.

PR #1423, opened 2014-07-23
Checkout flow refactor

mbertheau:
Our oscar implementation sells only non-corporal products (=no shipping), customers buy only one product at a time (=no basket with several products in different quantities) and there's only one payment method (=no payment method selection or data input). Currently when a customer buys a product, she has to click "continue" a number of times for no good reason (for our use case) and some automatic redirects for shipping shipping method and address and the like are also involved. I'd like to streamline this so that the process is shorter:

1. product-detail: click "buy"
2. basket: can add voucher; click "pay"
3. redirect to payment provider

That'll probably involve separating the checkout steps from the checkout views. I'm working on that right now.

### DIFF
--- a/TODO.branch
+++ b/TODO.branch
@@ -1,0 +1,21 @@
+Changes
+* any checkout step if requested directly will be executed; previously every
+  view checked pre_conditions and if need be redirected elsewhere
+* checkout:index is now always the right entry point to the checkout from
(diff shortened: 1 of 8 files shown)


sasha0:
Quite outdated and not active for a while. Please re-open when it's relevant again.

Thanks.

SalahAdDin:
@mbertheau What do you think know about this?

PR #1552, opened 2014-11-06
Explicitly pass the billing address around checkout

### LINKED ISSUE #1508: build_submission doesn't include billing address
As far as I can tell, the submission dict should include the billing address. It's easily added here: https://github.com/tangentlabs/django-oscar/blob/master/oscar/apps/checkout/session.py#L252

codeinthehole:
This changeset adjusts `build_submission` to include the billing address.
This also requires changes to a few method signatures to handle the
billing address explicitly.

This only works when the billing address is stored in the session (as
that's where it is looked up from when building the submission data in
`build_submission`). In many cases (such as Oscar's demo site), we don't
store the billing address in the session - instead, it is captured via
the payment-details forms. Hence, the billing address will still need to
be explicitly set within the submission data.

Fixes #1508

### DIFF
--- a/docs/source/releases/v0.8.rst
+++ b/docs/source/releases/v0.8.rst
@@ -24,7 +24,7 @@ two versions of Django, support for Django 1.5 has been dropped.
 
 This release also adds full Python 3.3 and 3.4 support. But due to South
 not supporting Python 3, Python 3 support is only supported in combination
-with Django 1.7 and it's new migration framework.
+with Django 1.7 and its new migration framework.
 
 If you're using ``pysolr`` for search, you need to upgrade to version 3.2 or
 later.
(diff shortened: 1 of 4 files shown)


boyander:
This makes conflicts when you use the default shipping address as billing adress?
```python
            submission = self.build_submission(
                order_kwargs={
                    'billing_address': address_form.save(commit=False),
                },
                payment_kwargs={
                    'bankcard_form': bankcard_form,
                    'stripe_token': bankcard_form.data.getlist('stripeToken')[0],
                    'billing_address_form': address_form
                }
            )
```

For me it causes
``` 
handle_order_placement() got multiple values for keyword argument 'billing_address' oscar/apps/checkout/views.py in submit, line 632
```

digicase:
I found the solution your problem @boyander 

When building the submission don't put the billing address in the order_kwargs. The submission should look like something like this:

    submission = self.build_submission(
        billing_address=billing_address_form.save(commit=False),
        payment_kwargs={
            ...
        }
    )

PR #1558, opened 2014-11-14
Allow changing of the Order.date_placed field

### LINKED ISSUE #1529: The Order.date_placed field shouldn't be auto_now_add=True
This changes when an order is updated later on which isn't generally the desired behaviour. Admittedly, this is rare but this field should be set explicitly.

monikasulik:
Removes the auto_now_add=True. Sets now as the default, but allows changing the date_placed field explicitly if an order is updated after it has been placed.

Fixes #1529


mbertheau:
@mvantellingen Thanks for the heads-up and sorry for the late reaction!

PR #1559, opened 2014-11-14
Remove mentions of vagrant from the docs

monikasulik:
Vagrant support was removed from release 1.0. This commit removes mentions of it from the docs.


maikhoepfel:
Cherry-picked across in https://github.com/django-oscar/django-oscar/commit/a17a79f7470b4a1888b08da2bd83a08b2c71ecf2. I removed the directory thing from ``setup.py`` and amended your commit before pushing.

Thanks!

PR #1565, opened 2014-11-19
Allow zero weight baskets to still get a weight-based charge

### LINKED ISSUE #1564: Zero weight baskets should still be chargable
Re https://groups.google.com/forum/#!msg/django-oscar/_VbrAmSikY8/a3rV9sBFI1AJ

We should still allow a weight band to catch zero price baskets.

codeinthehole:
There wasn't really a need to special-case zero weight baskets. This
does mean that stores that rely on this behaviour will need to create a
zero weight band.

If it's ok, let me know and I'll update the docs.

Fixes #1564

### DIFF
--- a/docs/source/releases/v1.1.rst
+++ b/docs/source/releases/v1.1.rst
@@ -21,6 +21,7 @@ Table of contents:
 Compatibility
 -------------
 
+
 .. _new_in_1.1:
 
 What's new in Oscar 1.1?
@@ -31,6 +32,8 @@ What's new in Oscar 1.1?
 Minor changes
 ~~~~~~~~~~~~~
 
+
+
 .. _incompatible_changes_in_1.1:
 
 Backwards incompatible changes in 1.1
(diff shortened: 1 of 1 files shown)


tests/integration/shipping/model_method_tests.py
```diff
@@ -116,6 +116,22 @@ def setUp(self):
...
+        self.assertEqual(D('4.00'), charge)
+
+    def test_zero_weight_baskets_can_have_no_charge(self):
+        self.standard.bands.create(upper_limit=0, charge=D('0.00'))
+        self.standard.bands.create(upper_limit=1, charge=D('4.00'))
+        charge = self.standard.get_charge(0)
+        self.assertEqual(D('0.00'), charge)
+
+    def test_get_band_for_zero_weight(self):
+        self.standard.bands.create(upper_limit=1, charge=D('4.00'))
```
willharris: Hmm, isn't this a duplication of test `test_zero_wegith_baskets_can_have_a_charge`... or maybe I'm going blind!

Issue #1560, opened 2014-11-14
Should promotion classes be hardcoded


binarymatt:
I was wondering if instead of doing a hard import PROMOTION_CLASSES like here:
https://github.com/django-oscar/django-oscar/blob/a04c7dfeb258ddf0687847098d82e351614b1d09/oscar/apps/dashboard/promotions/app.py#L5

Could that code instead be loaded with get_class? If that answer is yes, I would be more that willing to submit a pr.

PR #1567, opened 2014-11-20
dynamically load PROMOTION_CLASSES

binarydud:
This addresses the question I asked in #1560. It's a nice addition because it requires less forking if I want to add a custom promotion type.


sasha0:
Closing since PR is outdated and has conflicts and as already noted, it will make sense to extract whole application and do refactoring.

Issue #796, opened 2013-08-21
Removing dependencies from setup.py changes


codeinthehole:
From an internal email by @skyjur:

```
South:
        do we support south, but do we really need south to run oscar?

django-extra-views:
        I think we only use a single class from this library (the ModelFormsetView). Is it really worth to add a dependency because of one class? We can just copy-paste this class in our own abstract-views.

django-haystack:
        do we really require django-haystack? There might be support for it but I'm pretty sure it should be possible to run without it being installed.

python-memcached
        it looks like it is only used in demo file settings.py so it should be added as a requirements.txt of demo site
```

Some good points here - worth looking into

sasha0:
Closing, since Haystack is closely integrated into the Oscar and it don't see possibility to remove it easily. We'd revise this later, since @solarissmoke had idea of custom search backend. `django-extra-views` to be addressed in terms of the task #2225.

PR #1568, opened 2014-11-21
Remove django-compressor

mbertheau:
And replace it with a gulp file that automatically rebuilds the css when the less files change and possibly more bells and whistles.

I feel that django-compressor adds too much runtime complexity for frontend developer convenience that she is now used to getting from gulp and friends. Also, gulp and friends do a much better job at being convenient for the modern frontend developer:

- sourcemap support
- freedom of choice to use or not use django-compressor in the frontend. Right now if you want to use django-compressor in the frontend, you have to use it in the dashboard as well.
- no confusion and complexity arising from the fact that django-compressor parses templates outside of their normal context, bypassing inheritance (Blocks `jquery-ui-(js|css)` by yours truly)

I'm implementing this as we speak.

Concerns #796.

### DIFF
--- a/docs/source/howto/how_to_handle_statics.rst
+++ b/docs/source/howto/how_to_handle_statics.rst
@@ -27,43 +27,22 @@ Frontend vs. Dashboard
 
 The frontend and dashboard are intentionally kept very separate. They
 incidentally both use Bootstrap, but may be updated individually.
+
 The frontend is based on Bootstrap's LESS files and ties it together with
 Oscar-specific styling in ``styles.less``.
 
-On the other hand, ``dashboard.less`` just contains a few customisations that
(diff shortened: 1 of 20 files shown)


maikhoepfel:
I forgot we had this PR! Just to note, @jezdez just [announced](https://twitter.com/jezdez/status/568373925601783808) he won't be involved in ``django-compressor`` any more. So it's time we have another look at this. 

@codeinthehole What do you mean by "cache invalidation"? Adding MD5 stamps to filenames? Django got us [covered](https://docs.djangoproject.com/en/1.7/ref/contrib/staticfiles/#manifeststaticfilesstorage).

mvantellingen:
@mbertheau any update on this. I think we should move to gulp

mvantellingen:
This was done in #1890 
Closing this one.

PR #1572, opened 2014-11-21
Piper/make overriding search handler easier

### LINKED ISSUE #1566: Overriding the search handler used for product listings should be easier.
Currently, if you want to override the handler returned by `get_search_handler` on the various product listing views in the `catalogue` app, you can do one of the following things.
1. Subclass the view.
2. override `catalogue.search_handlers.get_product_search_handler_class`
3. override `catalogue.search_handlers.SimpleProductSearchHandler` or `catalogue.search_handlers.ProductSearchHandler`.
- **Option 1**  seems heavy, and it isn't _DRY_ if you want to override the handler globally.
- **Option 2** is relatively clean but seems awkward in the case that you just want to override what it returns.
- **Option 3** has to pay attention to what the return value of `get_product_search_handler_class` will be based on the return value of `is_solr_supported` in order to override the right class.

I think a new option would be helpful for those who have this use case.
- Add a setting `OSCAR_PRODUCT_SEARCH_HANDLER` which defaults to `None`.  When this value is set, the `catalogue.search_handlers.get_product_search_handler_class` function import and return the object at this path.
- Add a new search handler to be used when the `Elasticsearch` backend is in place for haystack, that uses the following code to narrow based on categories. (technically a separate feature so I'm happy to PR this in separately if you prefer)

``` python
class ElasticsearchProductSearchHandler(SearchHandler):
    form_class = BrowseCategoryForm
    model_whitelist = [Product]
    paginate_by = settings.OSCAR_PRODUCTS_PER_PAGE

    def __init__(self, request_data, full_path, categories=None):
        self.categories = categories
        super(ElasticsearchProductSearchHandler, self).__init__(request_data, full_path)

    def get_search_queryset(self):
        sqs = super(ElasticsearchProductSearchHandler, self).get_search_queryset()
        if self.categories:
            for category in self.categories:
                sqs = sqs.filter_or(category=category.full_name)
        return sqs
```

Is this a modification you would be open to me submitted a pull request for?

pipermerriam:
### What is the problem / feature ?

- No faceting support for the Elasticsearch on product listing pages (`CatalogueView` and `ProductCategoryView`)
- No easy way to specify the desired search handler class without heavy overrides.

### How did it get fixed / implemented ?

- Implemented a new search handler `ElasticsearchSearchHandler` which supports faceting for the Elasticsearch backend.
- Added a new setting `OSCAR_PRODUCT_SEARCH_HANDLER` which can be used to specify exactly what search handle you wish to use for listing pages.

Fixes #1566 

*Here is a cute animal picture for your troubles...*

thanksgivingday-dog-and-cat 1 _2_crop

### DIFF
--- a/docs/source/ref/settings.rst
+++ b/docs/source/ref/settings.rst
@@ -101,6 +101,14 @@ query-type facets.  The default is::
         ]),
     }
 
+``OSCAR_PRODUCT_SEARCH_HANDLER``
+-----------------------
+
+The search handler to be used in the product list views.
+
+Default::
+
+    None
 
 ``OSCAR_PROMOTION_POSITIONS``
 -----------------------------
--- a/src/oscar/apps/catalogue/search_handlers.py
+++ b/src/oscar/apps/catalogue/search_handlers.py
@@ -6,6 +6,7 @@ from oscar.core.loading import get_class, get_model
 BrowseCategoryForm = get_class('search.forms', 'BrowseCategoryForm')
 SearchHandler = get_class('search.search_handlers', 'SearchHandler')
 is_solr_supported = get_class('search.features', 'is_solr_supported')
+is_elasticsearch_supported = get_class('search.features', 'is_elasticsearch_supported')
 Product = get_model('catalogue', 'Product')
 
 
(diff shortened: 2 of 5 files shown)


pipermerriam:
I've rebased this PR, wonder if we can revive the discussion on this and figure out what needs to be done to get it mergable.

maikhoepfel:
Lovely work, @pipermerriam. And sorry for being so slow to respond; still pretty swamped at the moment.

Your explanation re: the setting makes good sense. Thank you for taking the time to do that; I fully agree.

We're still lacking a testing infrastructure to test against the different backends. Any work towards that is welcome; I don't currently see search being our focus. If I get some time to dedicate towards search, I think my first action would be throwing out haystack. 

Any chance you could contribute the Woosh backend? At least testing against that should be easy.

I merged your PR as-is, and did some minor changes in https://github.com/django-oscar/django-oscar/commit/c37241fec560f447391f3955e74dd529a1c0b8b9.

PR #1584, opened 2014-12-06
Remove duplicated italian locale

itbabu:
Now there are 2 italian translations:
- `it` (updated - 100%)
- `it_IT` (outdated - 10%, last update was in 2013)

The main issue is that almost 90% of italians (like me) have the system/browser language set to `it_IT`, so they will always see the outdated `it_IT` translation and not the updated `it` translation

Given that:
- We don't need multiple italian translations: Italian language in Italy (`it_IT`) and in Switzerland (`it_CH`) have no differences.
- the authors of `it_IT` appear already in the `it` .po file

I suggest to remove the `it_IT` translation from the repository and in transifex.


itbabu:
Whoops 
Italian duplicated `it_IT` translation is back to live in the [repository](https://github.com/django-oscar/django-oscar/tree/master/src/oscar/locale/it_IT/LC_MESSAGES) and in [Transifex](https://www.transifex.com/projects/p/django-oscar/language/it_IT/)

codeinthehole:
Files removed again in https://github.com/django-oscar/django-oscar/commit/723c9bd44c918b37d1578f311101cd41d4b455e7. I think they were accidentally reinstated by a folder move in December. Hopefully this time they won't come back.

PR #1603, opened 2014-12-23
Raise a ValueError if a string is passed to get_core_apps

codeinthehole:
It's such a common error, we should try and give a friendly error message when it occurs.


maikhoepfel:
Agreed & merged.

PR #1604, opened 2014-12-23
Disable migrations when running tests (for Django 1.7)

codeinthehole:
They make test runs really slow. This fix uses a cheeky gist to avoid running migrations for any apps:
https://gist.github.com/NotSqrt/5f3c76cd15e40ef62d09

### DIFF
--- a/tests/config.py
+++ b/tests/config.py
@@ -2,8 +2,17 @@
 import os
 import django
 
-from django.conf import settings, global_settings
+from django.conf import settings
 import oscar
+
+
+class DisableMigrations(object):
+
+    def __contains__(self, item):
+        return True
+
+    def __getitem__(self, item):
+        return "notmigrations"
 
 
 def configure():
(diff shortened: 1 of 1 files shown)


mvantellingen:
Pytest-django 2.8 now also implements this via:

`py.test  --nomigrations --reuse-db tests`

maikhoepfel:
I spent some time trying to get ``REUSE_DB`` working. First of all, Django default's to a in-memory SQLite database if the regular database is a SQLite database. So one has to [explicitly set a test DB name](https://docs.djangoproject.com/en/1.7/ref/settings/#test), otherwise `django-nose` will recreate it every time.
But even with that, I couldn't get it working. Most tests failed with operational errors complaining about missing tables, even though the database file is still present.

Did anybody else get it working?

mvantellingen:
Will be solved when #1724 is merged

Issue #1597, opened 2014-12-17
Order of form fields in ProfileForm – existing_user_fields


bgryszko:
When defining form fields with an `existing_user_fields` function, it mixes an order of fields. Maybe working only on `lists` or [OrderedSet](http://code.activestate.com/recipes/528878/) instead of `set` would be a solution? I can do PR for that.

Lists fields in form random order:
```python
class ProfileForm(forms.UserForm):
    class Meta:
        model = User
        fields = existing_user_fields(['my_new_field', 'first_name', 'last_name', 'email'])
```

Lists fields in proper order:
```python
class ProfileForm(forms.UserForm):
    class Meta:
        model = User
        fields = ['my_new_field', 'first_name', 'last_name', 'email']
```

PR #1605, opened 2014-12-24
Fix existing_user_fields ordering

bgryszko:
Working on sets causes a random order of returned fields while using
existing_user_fields as reported in django-oscar/django-oscar#1597.
Changed to work only on lists, which remain correct
order.


maikhoepfel:
Hi Bart, thank you very much. Excellent spot! I cherry-picked your commit across.

bgryszko:
I'm happy to contribute ;-) Thanks for merging.

Issue #1609, opened 2014-12-30
Customizing apps.address.forms.UserAddressForm and apps.checkout.forms.ShippingAddressForm


john-parton:
The US demo site customizes `apps.checkout.forms.ShippingAddressForm` to include a select widget for US State, as well as modifying a label and help_text.

But those changes are only for the checkout form, and not the `UserAddressForm`. Once the user has saved the address, if they got to change it at` /checkout/user-address/edit/<id>/`, then those customizations are gone.

The simplest solution is to just do all of the customizations again for `UserAddressForm`, but it would be slightly nicer if we could customize `apps.address.forms.AbstractAddressForm` directly.

sasha0:
@john-parton what's pending here?

john-parton:
I think it can be closed.

I also posted a mixin for this: https://github.com/django-oscar/django-oscar/pull/1610#issuecomment-76762025

sasha0:
@codeinthehole @maikhoepfel @monikasulik please close the issue. Thanks.

PR #1610, opened 2014-12-30
Better abstract forms

john-parton:
Move `AbstractAddressForm` to its own file and have `UserAddressForm` and `ShippingAddressForm` subclass `AbstractAddressForm` dynamically.

This will let a user to override `AbstractAddressForm`, so that both forms can be customized simultaneously.

Fixes https://github.com/django-oscar/django-oscar/issues/1609

### DIFF
--- a/oscar/apps/address/abstract_forms.py
+++ b/oscar/apps/address/abstract_forms.py
@@ -1,0 +1,14 @@
+from django.conf import settings
+from django import forms
+
+class AbstractAddressForm(forms.ModelForm):
+
+    def __init__(self, *args, **kwargs):
+        """
+        Set fields in OSCAR_REQUIRED_ADDRESS_FIELDS as required.
+        """
+        super(AbstractAddressForm, self).__init__(*args, **kwargs)
+        field_names = (set(self.fields) &
+                       set(settings.OSCAR_REQUIRED_ADDRESS_FIELDS))
+        for field_name in field_names:
+            self.fields[field_name].required = True
--- a/oscar/apps/address/forms.py
+++ b/oscar/apps/address/forms.py
@@ -1,23 +1,10 @@
-from django.conf import settings
-from django import forms
 
-from oscar.core.loading import get_model
+from oscar.core.loading import get_class, get_model
 from oscar.views.generic import PhoneNumberMixin
 
 UserAddress = get_model('address', 'useraddress')
 
-
-class AbstractAddressForm(forms.ModelForm):
-
-    def __init__(self, *args, **kwargs):
-        """
-        Set fields in OSCAR_REQUIRED_ADDRESS_FIELDS as required.
-        """
-        super(AbstractAddressForm, self).__init__(*args, **kwargs)
-        field_names = (set(self.fields) &
-                       set(settings.OSCAR_REQUIRED_ADDRESS_FIELDS))
-        for field_name in field_names:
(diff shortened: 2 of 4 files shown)


maikhoepfel:
John, I understand you're proposing this as it allows reducing duplication, but could you try to illustrate why this is important? What modifications do you wish to do to AbstractAddressForm?

I'm asking because I'm slightly reluctant to merge this because I can't think of any other place in Oscar where we allow both overriding the abstract class and the actual implementation. I could imagine cases where it makes it really hard to think about what is going on; and with Oscar letting you override classes, it's often hard enough.

What do you think?

john-parton:
I wanted to make some modifications to both `UserAddressForm` and `ShippingAddressForm` (and really anything that subclassed `AbstractAddressForm`) with a bunch of changes:

* change `state` to a `us_forms.USStateField` and limit the choices to contiguous states only
* Remove `country`
* Remove `line3`
* Add `help_text` to `state` about which states taxes are collected in
* Re-label `poscode` as "Zip Code"
* Some other minor changes

I had a bunch of repeated code between the two classes and wanted to simplify them.

I completely agree with not wanting to add unnecessary complexity, but at the time, it seemed like a reasonable solution. Now that I've had some time to think about it, I could probably remove the  duplication by defining a mixin, without having to make changes directly to Oscar.

I can see if I can get the mixin to work, and then maybe we could put some example code on the "Recipes" page of the docs. https://django-oscar.readthedocs.org/en/releases-1.0/howto/index.html

Thoughts?

maikhoepfel:
Hey John, sounds like you should indeed get it working without duplication in a mixin. I agree that's not ideal, but given my concerns from above, I think a decent approach overall.

And I'll happily welcome a new recipe :)

john-parton:
What's the best way to contribute a new recipe? Would I just modify the `docs/source/howto` and submit a PR? Or some other way?

#### How to override ShippingAddressForm and UserAddressForm without repeating yourself ####

```python
# apps.address.mixins
from django import forms

from localflavor.us import us_states as states
from localflavor.us import forms as us_forms

from oscar.core.loading import get_model

Country = get_model('address', 'Country')

class USAddressMixin(object):
    """ Mixin that customizes an AddressForm to make it more US-specific.
    """
    
        super(USAddressMixin, self).__init__(*args, **kwargs)
        self.fields['state'] = us_forms.USStateField(label="State", widget=forms.Select(choices=states.CONTIGUOUS_STATES),
                                                     help_text = "Only orders going to Kansas are liable for Sales tax")
        self.fields['line1'].label = "Address"
        self.fields['postcode'].label = "Zip code"

        if 'country' in self.fields:
            self.fields['country'].queryset = Country.objects.filter(pk='US')
            self.fields['country'].widget = forms.HiddenInput()
            self.fields['country'].initial = Country.objects.get(pk='US')

        for key in ('line2', 'line3'):
            if key in self.fields:
                del self.fields[key]
```

```python
# apps.address.forms
from oscar.apps.address import forms as address_forms

from apps.address.mixins import USAddressMixin

class UserAddressForm(USAddressMixin, address_forms.UserAddressForm):
    pass
```

```python
# apps.checkout.forms
from oscar.apps.checkout import forms as checkout_forms

from apps.address.mixins import USAddressMixin

class ShippingAddressForm(USAddressMixin, checkout_forms.ShippingAddressForm):
    pass

PR #1612, opened 2015-01-02
Updated Pillow and sorl-thumbnail dependencies.

wlonk:
There was a recent security update for Pillow (to version 2.7) that this will allow.

*Here's a cute animal picture for your troubles…*

Snow leopards!


codeinthehole:
Purrfect.

PR #1616, opened 2015-01-06
Documentation Clarification

certifiedloud:
I got stuck on the minor detail of how to add to kwargs. After reading the source code, I think the proposed changes to this doc will clarify for future users.

### DIFF
--- a/docs/source/howto/how_to_integrate_payment.rst
+++ b/docs/source/howto/how_to_integrate_payment.rst
@@ -50,8 +50,11 @@ communicating with payment gateways and creating the appropriate payment models.
 Payment logic is normally implemented by using a customised version of
 ``PaymentDetailsView``, where the ``handle_payment`` method is overridden.  This
 method will be given the order number and order total plus any custom keyword
-arguments initially passed to ``submit`` (as ``payment_kwargs``).  If payment is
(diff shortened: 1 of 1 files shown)


docs/source/howto/how_to_integrate_payment.rst
```diff
@@ -50,8 +50,11 @@ communicating with payment gateways and creating the appropriate payment models.
 Payment logic is normally implemented by using a customised version of
 ``PaymentDetailsView``, where the ``handle_payment`` method is overridden.  This
 method will be given the order number and order total plus any custom keyword
-arguments initially passed to ``submit`` (as ``payment_kwargs``).  If payment is
-successful, then nothing needs to be returned.  However, Oscar defines a few
+arguments initially passed to ``submit`` (as ``payment_kwargs``). If you need to make specifc
+kwargs available, then you'll have to override other methods from the same view as well. See the
+django-oscar-paypal project for details.
```
codeinthehole: I think this sentence needs to be more specific. Which methods do you mean? Is it this Paypal view that you're referring to? https://github.com/django-oscar/django-oscar-paypal/blob/master/paypal/express/views.py#L283-L292

Issue #1225, opened 2014-02-26
The offer app is not flexible enough


mvantellingen:
Hey,

For a project i'm working on we couldn't use the current offer models since they weren't flexible enough. The solution we use now is to create a separate app (offer_ext) and plug that in the offer Applicator.get_offers(). This works pretty well, however I think a better solution is to make the built-in offer app more flexible

Some of the problems I've encountered:
- We required discounts when a user bought a product from range 1 and a product from range 2. The number of products of each range is different per offer (sometimes buy 2 product X and get 50% discount on product Y).
- No option to add extra fields to ConditionalOffer. I think the solution is to create an AbstractOffer like the other apps.
- Offer name is unique, our client had multiple offers with the same name (although they are not always active at the same moment).
- Condition and benefit proxy class should not be unique. Also we needed more fields to implement the offers we wanted. 

I'll see if I can come up with a PR which adds some flexibility. I think the first step should be implementing the AbstractOffer/AbstractCondition/AbstractBenefit models

shahaamir:
I have taken twil's work further and added AbstractRange and AbstractRangeProduct models. I have merged twil@28dc759 into latest. 
We had a requirement where we needed to add a foreign key to Range model.

https://github.com/shahaamir/django-oscar/commits/overridable_offer_models

Is it OK to create a pull request for the same ?

codeinthehole:
On Mon, Jan 5, 2015 at 1:30 PM, aamir khan <notifications@github.com> wrote:

> I have taken twil's work further and added AbstractRange and
> AbstractRangeProduct models. I have merged twil@28dc759
> <https://github.com/twil/django-oscar/commit/28dc759> into latest.
> We had a requirement where we needed to add a foreign key to Range model.
>
> https://github.com/shahaamir/django-oscar/commits/overridable_offer_models
>
> Is it OK to create a pull request for the same ?
>
​Of course. Please go ahead.​

sasha0:
All issues, described in the task, has been already addresses.

For further offer application improvement we'd open a new task.

PR #1617, opened 2015-01-07
Overridable offer models

shahaamir:
Add abstract models ConditionalOffer, Range, RangeProduct
https://github.com/django-oscar/django-oscar/issues/1225


mvantellingen:
I've continued this in #1721 

Thanks!

PR #1618, opened 2015-01-08
Wrong data in orders.csv

ka7eh:
When orders are downloaded as CSV in the dashboard, the generated rows contain the keys from <code>metadata</code> rather than orders' values. I updated the line that should pass the values to <code>UnicodeCSVWriter</code>.


codeinthehole:
Thanks - I squashed your commits and rebased onto master.

codeinthehole:
Thanks - I squashed your commits and rebased onto master.

PR #1620, opened 2015-01-11
Fix typo for unsuspend alters_data

bharathwaaj:
The alters_data attribute on unsuspend is mistakenly written as suspend. This patch fixes this typo.


codeinthehole:
Thank you sir

Issue #1625, opened 2015-01-14
Sandbox: Advanced Search in dashboard/orders with date range results in attribute error


lijoantony:
1. Click on Advanced Search link next to Search button in /dashboard/orders
2. Enter date range 9/9/2014 - 12/12/2014
3. Click search
4. Django error page with below information is returned.

AttributeError at /en-gb/dashboard/orders/
'datetime.date' object has no attribute 'tzinfo'
Request Method:	GET
Request URL:	http://127.0.0.1:5000/en-gb/dashboard/orders/?csrfmiddlewaretoken=13ipHmklGCwrfxXksC5k6MjBjxeqJBuy&order_number=&name=&product_title=&upc=&partner_sku=&status=&date_from=9%2F9%2F2014&date_to=12%2F12%2F2014&voucher=&payment_method=&response_format=html
Django Version:	1.7.1
Exception Type:	AttributeError
Exception Value:	
'datetime.date' object has no attribute 'tzinfo'
Exception Location:	/home/lijo/.virtualenvs/eshop1/local/lib/python2.7/site-packages/django/utils/timezone.py in is_naive, line 349

TRACEBACK
===========
Environment:

Request Method: GET
Request URL: http://127.0.0.1:5000/en-gb/dashboard/orders/?csrfmiddlewaretoken=13ipHmklGCwrfxXksC5k6MjBjxeqJBuy&order_number=&name=&product_title=&upc=&partner_sku=&status=&date_from=9%2F9%2F2014&date_to=12%2F12%2F2014&voucher=&payment_method=&response_format=html

Django Version: 1.7.1
Python Version: 2.7.6
Installed Applications:
['django.contrib.auth',
 'django.contrib.contenttypes',
 'django.contrib.sessions',
 'django.contrib.sites',
 'django.contrib.messages',
 'django.contrib.admin',
 'django.contrib.flatpages',
 'django.contrib.staticfiles',
 'django.contrib.sitemaps',
 'django_extensions',
 'debug_toolbar',
 'template_timings_panel',
 'compressor',
 'apps.gateway',
 'widget_tweaks',
 'oscar',
 'oscar.apps.analytics',
 'oscar.apps.checkout',
 'oscar.apps.address',
 'oscar.apps.shipping',
 'oscar.apps.catalogue',
 'oscar.apps.catalogue.reviews',
 'oscar.apps.partner',
 'oscar.apps.basket',
 'oscar.apps.payment',
 'oscar.apps.offer',
[...]

PR #1626, opened 2015-01-14
Issue1625: Do not format datetime.date object using format_datetime

lijoantony:
utils.format_datetime expects an instance of datetime.datetime as input.
OrderListView.get_desc_context method calls format_datetime by passing datetime.date object which results in an AttributeError.

These calls to format_datetime have been removed and the date objects are directly serialized to string just like how it is done in dashboard/reports/reports.py


lijoantony:
Fix for #1625

codeinthehole:
Thanks for this. 

FWIW, you don't need to prefix commit messages with `IssueXXX`. Better to just include a `Fixes #xxx` in the commit message body (not the top line).

lijoantony:
Thanks for the advice. I will do that in future.

PR #1627, opened 2015-01-14
Change the doc link to django-treebeard

dduong42:


mbertheau:
@dduong42 Thanks a lot :)

PR #1631, opened 2015-01-16
allow cookies to be set as secure cookie.

mikedingjan:
Make the secure property of cookies configurable.


mvantellingen:
rebased and merged, dank u :-)

PR #1632, opened 2015-01-16
google analytics moved to head

jlmadurga:
Google analytics tracking script moved from body to head as it is recommended by google:

https://support.google.com/analytics/answer/1008080?hl=en#GA

I don't know if proposed before but just to comment I don't find useful to track dashboard pages. Dashboard might use other base.html


mvantellingen:
Thanks! Merged it via a rebase

PR #1634, opened 2015-01-19
Clicking back or refreshing page in Firefox doesn't reset buttons' state

ka7eh:
There's a bug in Firefox (https://bugzilla.mozilla.org/show_bug.cgi?id=654072), which prevents buttons from getting reset.
For example, if I press 'Ship to this address' on shipping-address view of the checkout page and go to the next page and then press the back button, the 'Ship to this address' button stays disabled and shows its data-loading-text, i.e. 'Saving...'. Refreshing the page doesn't help either.
A workaround is resetting all buttons on 'pageshow', similar to what's mentioned here: https://github.com/twbs/bootstrap/issues/1117
I updated ui.js and dashboard.js to do so in forms.init().


mbertheau:
Thank you for the heads up and the PR. The workaround in the PR has the downside that it removes resets the disabled attribute for all buttons, including buttons that should be disabled. Before merging we'll have to see how big of an actual problem that is for us.

ka7eh:
I closed this pull request as it wasn't the best solution.
Is it possible to convert it to an issue from here?

mbertheau:
Not that I know of. Please open a new issue.

ka7eh:
Just copy/pasted it as an issue.

PR #1642, opened 2015-01-23
Fix path in sandbox missing image symlink

wedaly:
Steps to reproduce:
1) Follow the steps in the documentation to set up a sandbox (create the virtualenv, create the sandbox, and start the server).
2) Log in to `/dashboard` as the superuser.
3) Click "Manage" under "Catalogue and Stock"

Currently in master, this produces the following error: "Improperly configured at "/en-gb/dashboard/catalogue/"
image

This PR updates the path in the symlink so that the image displays correctly.

Please let me know if there is anything I can improve to get this merged.  Thanks!
-- Will


wedaly:
This appears to be fixed in master now, so I'm going to close this PR.

PR #1630, opened 2015-01-15
Issue/675/per variant image support

john-parton:
## Summary ##
* Should close https://github.com/django-oscar/django-oscar/issues/1611
* Working towards https://github.com/django-oscar/django-oscar/issues/675

### Changes ###
* Enable association of variants with specific images
    * New `ManyToMany` relationship
    * Dashboard integration
    * Basket integration (show correct thumbnail)
    * `ProductDetailView` integration
* `AddToBasketForm` has been modified to better reflect how a user interacts with it
    * When a user views a variant, the other 'siblings' are displayed in the form
    * Form uses `AdvancedRadioSelect` instead of `AdvancedSelect`, and wraps the choices in links to the actual variant pages
    * `child_id` hack has been removed, replaced with more explicit `ModelChoiceField`
    * `AdvancedSelect` widgets has been enhanced so that disabled values do not have to be pre-computed and passed to the constructor, but rather passed along with choices (details in source)
    * `AdvancedChoice` and `DisabledChoice` helper classes to define disabled options
    * More options for `AdvancedSelect`: `AdvancedRadioSelect`, `AdvancedSelectMultiple`, `AdvancedCheckboxSelectMultiple`
* `ProductDetailView` now returns a full page for a variant
    * The plan is to add in a progressive enhancement, so we can do an AJAX request to that URL endpoint, and get some HTML fragments or some JSON representing the variant

### Weak points / To-do ###
* The UI for associating images with variants is currently a little clunky. 
* Associating variants with images is currently pretty tedious, maybe find some way to streamline it
* Currently doesn't include a progressive enhancement for making variant selection simpler
* Documentation & Tests
* The UI still doesn't allow the user to drill down by "dimensions" (e.g. Color, Size, Flavor, etc.), and fundamentally still presents the variants in a flat fashion

#### PS ####
I'm planning on making regular contributions to Oscar. Please let me know if there's anything I can do to make this process better!

John P.

Issue #1611, opened 2014-12-30
Image not available for variants in basket


john-parton:
If a variant's parent has images the basket still shows a "Image Not Available" placeholder for all variants.

The expected behavior would be for the variant to default to the parents image.

Here's a workaround in my app:

``` python
class Product(AbstractProduct):

    def primary_image(self):
        # Hack to make primary image look at the parent if it's a child

        if self.structure == self.CHILD:
            product = self.parent
        else:
            product = self

        return super(Product, product).primary_image()
```

I can submit a pull request with a proper patch if you want.

john-parton:
https://github.com/john-parton/django-oscar/tree/issue/675/per_variant_image_support
has explicit handling for variant images, and fixes this properly, instead of just a simple workaround.

Issue #675, opened 2013-05-16
Better variant product support


codeinthehole:
Variant products have been second class citizens in Oscar for a while and it's time to rectify this.  We need to ensure that products which have variations along one or more dimensions are supported properly with the right data being passed into the template to allow custom widgets to be used.

Example: A T-shirt is available in range of colours and sizes.  We need to be able to support widgets like this one (from gap.com):
![image](https://f.cloud.github.com/assets/80975/511882/80927d84-be11-11e2-9296-161e0333763e.png)

Sometimes all variants will be the same price but sometimes they will be different.  We need to think about how we handle both scenarios.

Related tickets: https://github.com/tangentlabs/django-oscar/issues/481, https://github.com/tangentlabs/django-oscar/issues/379
Related mailing list thread: https://groups.google.com/forum/#!topic/django-oscar/nKVelLiVQmI, https://groups.google.com/forum/?fromgroups=#!topic/django-oscar/PXs6jgdgdWs

john-parton:
I've started some basic work on this.

![screenshot from 2015-01-08 16 09 53](https://cloud.githubusercontent.com/assets/2071543/5671858/f70bea44-9750-11e4-91a0-ec4eb4e96326.png)

Here's some of the interesting bits:

```python
# templatetags.py

def _get_variant_descriptors(product, timeout=DEFAULT_TIMEOUT):
    
    descriptors = {}
    
    for attribute_value in ProductAttributeValue.objects.filter(product=product.children.all()).select_related('product', 'attribute'):
        
        attribute_name, value = attribute_value.attribute.name, attribute_value.value
        
        descriptors.setdefault( attribute_name, {}).setdefault(value, set()).add(attribute_value.product_id)
    
    # In general, we want things sorted when we output them
    return sorted( (attribute, sorted(value_map.items())) for attribute, value_map in descriptors.items() )

@register.inclusion_tag('catalogue/variant_descriptors.html')
def render_variant_descriptors(product):
    return {
        'product': product,
        'descriptors': _get_variant_descriptors(product)
    }
```
Output of `_get_variant_descriptors` looks like this:
```python
[
    (u'Filler', [
        (u'Down Feather', set([26064, 26062])), 
        (u'Poly Fiber', set([26065, 26063]))
    ]), 
    (u'Size', [
        (u'18" x 18"', set([26062, 26063])), 
        (u'22" x 22"', set([26064, 26065]))
    ])
]
```

[...]

PR #1643, opened 2015-01-23
Add per-variant image support

john-parton:
This is the same as  https://github.com/django-oscar/django-oscar/pull/1630 but against `master`. Please close https://github.com/django-oscar/django-oscar/pull/1630
## Summary
- Should close https://github.com/django-oscar/django-oscar/issues/1611
- Working towards https://github.com/django-oscar/django-oscar/issues/675
### Changes
- Enable association of variants with specific images
  - New `ManyToMany` relationship
  - Dashboard integration
  - Basket integration (show correct thumbnail)
  - `ProductDetailView` integration
- `AddToBasketForm` has been modified to better reflect how a user interacts with it
  - When a user views a variant, the other 'siblings' are displayed in the form
  - Form uses `AdvancedRadioSelect` instead of `AdvancedSelect`, and wraps the choices in links to the actual variant pages
  - `child_id` hack has been removed, replaced with more explicit `ModelChoiceField`
  - `AdvancedSelect` widgets has been enhanced so that disabled values do not have to be pre-computed and passed to the constructor, but rather passed along with choices (details in source)
  - `AdvancedChoice` and `DisabledChoice` helper classes to define disabled options
  - More options for `AdvancedSelect`: `AdvancedRadioSelect`, `AdvancedSelectMultiple`, `AdvancedCheckboxSelectMultiple`
- `ProductDetailView` now returns a full page for a variant
  - The plan is to add in a progressive enhancement, so we can do an AJAX request to that URL endpoint, and get some HTML fragments or some JSON representing the variant
### Weak points / To-do
- The UI for associating images with variants is currently a little clunky. 
- Associating variants with images is currently pretty tedious, maybe find some way to streamline it
- Currently doesn't include a progressive enhancement for making variant selection simpler
- Documentation & Tests
- The UI still doesn't allow the user to drill down by "dimensions" (e.g. Color, Size, Flavor, etc.), and fundamentally still presents the variants in a flat fashion

John P.

### DIFF
--- a/src/oscar/apps/basket/fields.py
+++ b/src/oscar/apps/basket/fields.py
@@ -9,9 +9,9 @@ class VariantChoice(AdvancedChoice):
     # Bootstrap 2 & 3 classes
     DISABLED_CLASSES = ['muted', 'text-muted']
 
-    def __init__(self, label, disabled=False, href=None):
-        self.href = href
-        super(VariantChoice, self).__init__(label, disabled=disabled)
+    def __init__(self, label, **kwargs):
+        self.href = kwargs.pop('href', None)
+        super(VariantChoice, self).__init__(label, **kwargs)
 
     def __unicode__(self):
         if self.href:
--- a/src/oscar/apps/basket/fields.py
+++ b/src/oscar/apps/basket/fields.py
@@ -1,0 +1,55 @@
+from django import forms
+from django.utils.encoding import force_text
+from django.utils.html import format_html
+
+from oscar.forms.widgets import AdvancedChoice, AdvancedRadioSelect
+
+class VariantChoice(AdvancedChoice):
+
+    # Bootstrap 2 & 3 classes
+    DISABLED_CLASSES = ['muted', 'text-muted']
+
+    def __init__(self, label, disabled=False, href=None):
+        self.href = href
+        super(VariantChoice, self).__init__(label, disabled=disabled)
+
+    def __unicode__(self):
+        if self.href:
+            if self.disabled:
+                disabled = format_html(' class="{0}"', ' '.join(self.DISABLED_CLASSES))
+            else:
+                disabled = ''
+            return force_text( format_html('<a href="{0}"{1}>{2}</a>', self.href, disabled, self.label) )
+        else:
+            return self.label
+
+class VariantChoiceField(forms.ModelChoiceField):
+    """
+    Field that renders variant choices of the AddToBasket form.
+    """
+
+    def __init__(self, *args, **kwargs):
+        if 'widget' not in kwargs:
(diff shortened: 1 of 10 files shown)


mvantellingen:
Hi John,

Sorry for not coming back to this PR sooner. This PR seems to include two things: 
 - better variant support
 - fallback to images on the master product if the child product doesn't have any.

I think this should be kept in two separate PR's, this makes it easier for us to review and merge it.

Thanks.

john-parton:
I'll try to see if I can factor it out. Could you point out which lines you think should be separated?

Thanks

john-parton:
Let me see if I can elaborate.

With Oscar in its current form, it's impossible to determine which images that are associated with a parent actually belong to which variants. For instance, you could have a T-shirt that comes in red, green and blue. In the cart, the best you can do is just grab the primary image, but that might not correspond to the correct color. A user might easily be confused if a red shirt appears in their cart when they added a green variant.

This PR adds a ManyToManyField that allows you to specify exactly which images go with which variant. This makes it trivial to select the correct image in the cart.

It's not really "two fixes," but rather "killing two birds with one stone."

mvantellingen:
I think a clean solution would be the following:
 - Have a separate product detail pages for each child. Selecting a variant (child) with the dropdown will load the page of the child
 - Alter the dashboard to allow uploading images on child products
 - If a child product has no images then fall back to the images on the parent.

This way the different prices for each child can also be rendered properly. 

I've used this approach on for example: https://shop.audi.nl/lifestyle/kleding/shirts/BV3131301105-t-shirt-s-line

john-parton:
Uploading a new image for each variant seems like a poor fix. In the T-shirt example, you could have small, medium and large sizes. You wouldn't necessarily want to upload an image for each size variation.

That's why I implemented it just as a ManyToManyField.

sasha0:
First of all, thanks for all your hard work @john-parton.

I personally agree with proposed approach by @mvantellingen and since we aim to introduce new dashboard layout, it's a good time to revise product image upload too.

solarissmoke:
I feel like we need a new ticket to define what the ideal feature set for catalogue images would be, and how this should be exposed in the dashboard (the current image upload stuff is pretty clumsy).

sasha0:
Could you please create it? What's your opinion about this PR?

SanskarSans:
can i get the branch for this to test in my app, please?

sasha0:
I reviewed this again and we discussed with the team and agreed that approach proposed by Michael long time ago - https://github.com/django-oscar/django-oscar/pull/1643#issuecomment-97575119 would be the best solution.

Thus, this PR is no longer relevant, so I'm closing it.

PR #1644, opened 2015-01-24
Use treebeard 3.0 which silences Django 1.7 warnings

mvantellingen:


mbertheau:
Woohoo! :)

codeinthehole:
Does this have any backwards compatibility implications? Can we continue to support Django 1.6.x for Oscar 1.1?

mvantellingen:
It should still support Django 1.6, as far as I can tell it's mostly a release to make it compatible with Django 1.7+

See https://github.com/tabo/django-treebeard/blob/master/CHANGES

clintonb:
Any idea when this change will be released? It didn't make 1.0.2.

clintonb:
Never mind. I just saw the milestone.

PR #1645, opened 2015-01-24
Update checkout tests with new factories

itbabu:
I  added fuctional tests to some checkout views: IndexView, ShippingAddressView, UserAddressUpdateView and DeleteUserAddressView.

There are some views left but in the meantime a test highlights that phonenumber is not working as expected with Python 3 and Django 1.6

```python
    def test_submitting_valid_form_modifies_user_address(self):
        page = self.get(reverse('checkout:user-address-update', kwargs={'pk': self.user_address.pk}),
                        user=self.user)
        form = page.forms['update_user_address']
        form['first_name'] = 'Tom'
        response = form.submit()
        self.assertRedirectUrlName(response, 'checkout:shipping-address')
        self.assertEqual('Tom', UserAddress.objects.get().first_name)
```
This test will fail:
```python
(Pdb) response
<200 OK text/html body=b'\n\n<!DO...l>\n'/14519>
(Pdb) response.form['phone_number'].value
'Country Code: 49 National Number: 3513296645 Country Code Source: 1 Preferred Domestic Carrier Code: '
```
It seems that the form is filled with the stringified  phone number object and so the submit fails.

```python
(Pdb) response.html
'(...)<span class="error-block"><i class="icon-exclamation-sign"></i> This is not a valid local or international phone format.</span>(...)'
```
I still have to figure out why this happens

### DIFF
--- a/src/oscar/apps/checkout/views.py
+++ b/src/oscar/apps/checkout/views.py
@@ -177,12 +177,6 @@ class ShippingAddressView(CheckoutSessionMixin, generic.FormView):
                 # User has selected a previous address to ship to
                 self.checkout_session.ship_to_user_address(address)
                 return redirect(self.get_success_url())
-            elif action == 'delete':
-                # Delete the selected address
-                address.delete()
-                messages.info(self.request, _("Address deleted from your"
-                                              " address book"))
-                return redirect(self.get_success_url())
             else:
                 return http.HttpResponseBadRequest()
         else:
--- a/src/oscar/templates/oscar/checkout/shipping_address.html
+++ b/src/oscar/templates/oscar/checkout/shipping_address.html
@@ -33,7 +33,7 @@
                                             {% endfor %}
                                         {% endblock %}
                                     </address>
-                                    <form action="{% url 'checkout:shipping-address' %}" method="post">
+                                    <form action="{% url 'checkout:shipping-address' %}" method="post" id="select_shipping_address_{{ address.id }}">
                                         {% csrf_token %}
                                         <input type="hidden" name="action" value="ship_to" />
                                         <input type="hidden" name="address_id" value="{{ address.id }}" />
--- a/src/oscar/templates/oscar/checkout/user_address_delete.html
+++ b/src/oscar/templates/oscar/checkout/user_address_delete.html
@@ -17,7 +17,7 @@
 {% block checkout_title %}{% trans "Delete address?" %}{% endblock %}
 
 {% block shipping_address %}
-    <form action="." method="post">
+    <form action="." method="post" id="delete_address_{{ object.id }}">
         {% csrf_token %}
         <div class="well">
             <address>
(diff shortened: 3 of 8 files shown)


maikhoepfel:
Wow, lovely work. I'm having a look at the phone number issue now.

maikhoepfel:
I lost wifi on the train and then got changed to a train without power outlets... I'll get to look at it on Friday.

itbabu:
I'll add few more tests in the next days.

Is there a reason to keep both [`assertRedirectUrlName`](https://github.com/django-oscar/django-oscar/blob/master/src/oscar/test/testcases.py#L87-L90) and [`assertRedirectsTo`](https://github.com/django-oscar/django-oscar/blob/master/src/oscar/test/testcases.py#L76-L80) assertions?

maikhoepfel:
Marco, you're right, those two helpers are duplicates. I removed one here: https://github.com/django-oscar/django-oscar/commit/09f66c35eaad53adcfc98bbb965d16e5e68a9393

I also had a look at the unused delete code in ``ShippingAddressView`` and have to agree with you. As it's not really related to this PR, I took the liberty of cherry-picking it onto master here: https://github.com/django-oscar/django-oscar/commit/549decb01632714d124a412aa53dafd9e72cdeff I only modified the commit message.

Thanks for this. I'll investigate the phone number thing after lunch.

maikhoepfel:
@itbabu, why are you suggesting adding unique IDs to the <form>s? I don't think there's harm in it; I'd just like to understand the reasoning.

itbabu:
@maikhoepfel It simplifes and make more understandable a functional test when we need to [select](http://webtest.readthedocs.org/en/latest/forms.html) a form inside a page. 
Sometimes a page has lot of forms  (see these tests: `test_can_delete_a_user_address_from_shipping_address_page`
`test_can_select_an_existing_shipping_address`)

There are already some tests in Oscar using this.
For example see https://github.com/django-oscar/django-oscar/blob/485a5d09963a213fd802a264c8f9351b83418371/tests/functional/customer/profile_tests.py#L32-L41

maikhoepfel:
@itbabu, thanks for all the work and taking the time to explain. 

I just pushed a squashed version of your work in https://github.com/django-oscar/django-oscar/commit/a0ce8f22dc0975502e123527dd3fe0c97be9d0d3. Apart from some PEP8 changes and using ``assertRedirectsTo`` instead of the removed method, it's unchanged. I also finally managed to track down the bug you found: https://github.com/django-oscar/django-oscar/commit/9537531062f5eea5563c6cc8c06fab3564228ccd

I did quite a bit of cleanup along the way: https://github.com/django-oscar/django-oscar/commit/7b31081ec5c8a49f546eccc654b29ca5df9998f7 and https://github.com/django-oscar/django-oscar/commit/ee4307056c44c10d40b6dbac8766b16a534748e8.

With that, I'm closing this PR. You're most welcome to add more tests; please create a new PR for that.

Issue #1530, opened 2014-10-23
Translatability issues in orders dashboard


maikhoepfel:
The issue is (as so often) where we concatenate strings, which doesn't work for some languages. The offending block is here: https://github.com/tangentlabs/django-oscar/blob/master/oscar/apps/dashboard/orders/views.py#L144-L211

It was reported in Transifex here: https://www.transifex.com/projects/p/django-oscar/translate/#eu/oscar/19718168 

The fix is to have full translation strings for each choice.

sasha0:
The issue does seem to exist any more - dashboard order list filters translated separately and concatenated afterwards, which works fine.

PR #1646, opened 2015-01-25
i18n order search filters

wedaly:
This is an attempt to address #1530.  In the current implementation, search filters are concatenated together and displayed in the header of the search results table.  This is problematic for translators, since other languages have different grammar rules for combining fragments like these.

Since there are 10 different possible filters, it seemed unreasonable to create translation strings for each possible combinations of filters.  Instead, I made a slight change to the UI in order to display each applied filter separately:

image

By default, the filter display is toggled off so users aren't distracted from the search results.

I have also added integration tests for this feature using the Django test client.  To avoid defining a lot of repetitive test cases, I added `ddt` (Data Driven Tests) as a dependency and used it in the new test.  I'm a fan of `ddt` because it makes tests more succinct; but I can remove it if you'd prefer not to introduce another dependency.

I'm happy to make any changes necessary to get this merged, so please let me know!

Thanks!
-- Will

### DIFF
--- a/src/oscar/apps/dashboard/orders/views.py
+++ b/src/oscar/apps/dashboard/orders/views.py
@@ -22,6 +22,7 @@ from oscar.apps.order import exceptions as order_exceptions
 
 Partner = get_model('partner', 'Partner')
 Transaction = get_model('payment', 'Transaction')
+SourceType = get_model('payment', 'SourceType')
 Order = get_model('order', 'Order')
 OrderNote = get_model('order', 'OrderNote')
 ShippingAddress = get_model('order', 'ShippingAddress')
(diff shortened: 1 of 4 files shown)


wedaly:
@codeinthehole would you be able to review this?

codeinthehole:
Yes, I'll take a look. Just trying to clear an evening to go through the 1.1 issus.

maikhoepfel:
Those are good changes, thanks Will. 

I understand why you're using ``ddt``, but as this is the only case in our test suite that I can think of where things really multiply like this, I'd rather not add it as a new dependency. We already have a few legacy dependencies in our test suite that we're trying to clear, so we're a bit reluctant to add a new one. Could you remove it?

The JS i18n URL is something all Oscar deployment would have to add. Instead of doing that, could you just create two HTML elements and have one hidden, and then toggle them appropriately in the Javascript code? That would avoid the extra URL and extra HTTP requests.

Feel free to bug me about this; it seems @codeinthehole is quite busy at the moment.

wedaly:
@maikhoepfel Thanks for the review, and sorry this fell off my radar the last few weeks.  I've updated the PR to remove the `ddt` and `jsi18n` dependencies.

mvantellingen:
Hi @wedaly, i've just merged this PR with some small modifications:

1. Use bootstrap 3's labels to represent search filters. We might want to look at django-filter for something more sophisticated in the future.
2. Rewrite the test using WebTestCase and move it to the functional tests for the order dashboard.

Thanks again and sorry for the late response!

Issue #1085, opened 2013-12-16
Use a custom StockInfo class for variants


codeinthehole:
As noted in #481, it might be better to have a custom `StockInfo` class for variants that includes the min and max price of an item.

Some research required...

sasha0:
It indeed will allow to indicate variant prices appropriately. WDYT @django-oscar/core ?

specialunderwear:
closing all issues that are feature requests that don't have PR's attached.

Issue #1648, opened 2015-01-27
Price ranges for parent products


john-parton:
The default strategy classes in `partner.strategy` as well as the default templates don't allow parents to express their price as a range.

It would be nice to be able to do something like this with minimal configuration:

![screenshot from 2015-01-27 13 45 57](https://cloud.githubusercontent.com/assets/2071543/5925691/0b81555c-a62b-11e4-83c2-c0580d4028ed.png)

Likewise, the price range would show on the product page.

I'm working on a fix.

Possibly related: https://github.com/django-oscar/django-oscar/issues/675

john-parton:
I think this might be a duplicate of https://github.com/django-oscar/django-oscar/issues/1085

sasha0:
Closing as a duplicate of #1085.

PR #1649, opened 2015-01-27
Issue/1648/price range

john-parton:
This is a fix for https://github.com/django-oscar/django-oscar/issues/1648

* Added a `__cmp__` method to `partner.prices.Base` so that prices can be sorted and compared in templates.
* `PurchaseInfo` is now a `namedtuple` with `min_price` and `max_price` instead of `price`
    * `price` is still a property that aliases `min_price` (to emulate old behavior for backwards compatability)
* `Selector.parent_pricing_policy` returns a list of prices or `Unavailable` instead of a single price
* `fetch_for_parent` actually looks at the prices returned and determines what the `min_price` and `max_price` should be for `PurchaseInfo`
* `templates/oscar/catalogue/partials/stock_record.html` now looks at `min_price` and `max_price` and renders a single price or a price range as appropriate.

### DIFF
--- a/oscar/apps/partner/prices.py
+++ b/oscar/apps/partner/prices.py
@@ -33,6 +33,11 @@ class Base(object):
     #: Price currency (3 char code)
     currency = None
 
+    # Used to enable sorting of prices
+    # Ignores currency, but is otherwise correct if all Base instances are the same price
+    def __cmp__(self, other):
+        return self.effective_price - other.effective_price
+
     def __repr__(self):
         return "%s(%r)" % (self.__class__.__name__, self.__dict__)
 
--- a/oscar/apps/partner/prices.py
+++ b/oscar/apps/partner/prices.py
@@ -34,7 +34,7 @@ class Base(object):
     currency = None
 
     # Used to enable sorting of prices
-    # Ignores currency, but is otherwise correct if all Base instances are the same price
+    # Ignores currency, but is correct if all Base instances are the same currency
     def __cmp__(self, other):
         return self.effective_price - other.effective_price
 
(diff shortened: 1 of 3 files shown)


john-parton:
I think this also fixes https://github.com/django-oscar/django-oscar/issues/1085

john-parton:
@maikhoepfel Could I get some feedback on this?

I think this is an important feature and I believe my solution is relatively straight forward.

Thanks.

ademuk:
:+1:This change is working well for me.

FYI @john-parton Not currently merging automatically. Some tests would be good too.

john-parton:
It should merge automatically again.

Any specific suggestions for tests?

john-parton:
I see that it's failing some tests now. I'll see if I can get this working again.

john-parton:
Looks like a factory was creating PurchaseInfo instances with the price kwarg. Updated the factory to explicitly set the min/max price.

john-parton:
Had to remove `__cmp__` and replace it with `@total_ordering`, `__lt__`, and `__eq__` because `__cmp__` is not supported in Python 3.

john-parton:
@ademuk Is this still working well for you? Any suggested changes?

@mvantellingen Can we get this merged in here soon? I think it's a pretty sensible change.

mvantellingen:
Hi @john-parto, I think this should be done after 1.2 is released and in combination with https://github.com/django-oscar/django-oscar/pull/1298

john-parton:
@mvantellingen Thanks for the feedback. I'll try to keep an eye on https://github.com/django-oscar/django-oscar/pull/1298 and modify my PR to be compatible once its merged in.

craigloftus:
Close due to inactivity. Happy to open it again if there is motivation to finalise it.

PR #1650, opened 2015-01-31
Fix after user registration, the page does not redirect to checkout page...

sunsong:
... as url requested bug.

Happens when an anonymous user buying and checking out, the user will be asked to login or register. After the registration, the page will be redirected to account profile page instead of finishing the checkout process.

### DIFF
--- a/src/oscar/apps/customer/views.py
+++ b/src/oscar/apps/customer/views.py
@@ -217,6 +217,10 @@ class AccountAuthView(RegisterUserMixin, generic.TemplateView):
         return _("Thanks for registering!")
 
     def get_registration_success_url(self, form):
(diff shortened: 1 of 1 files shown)


mvantellingen:
Thanks! :-)

MARUTHESH:
I just added OSCAR_ALLOW_ANON_CHECKOUT = True in my setting.py file, even after it redirects me to the login page when I click Proceed to Checkout button, I don't understand what is wrong with this, some one please help me out

solarissmoke:
@MARUTHESH please stop spamming issue threads with support requests. Github issues are not used for support - please use the mailing list for that. Thanks.

PR #1657, opened 2015-02-05
This fix #1647. Checking for voucher availability is working correctly

### LINKED ISSUE #1647: Offer condition in Applicator.get_basket_offers always returns True.
`Applicator.get_basket_offers()` checks for applicable basket offers using something like `if voucher.is_available_to_user(user)` (see https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/offer/utils.py#L99). However, the return value of `is_available_to_user()` is a 2-tuple, so this will always evaluate to True.

jonasborges:
Checking for voucher availability is working correctly on Applicator.get_basket_offers(). 
Tests updated.

### DIFF
--- a/src/oscar/apps/offer/utils.py
+++ b/src/oscar/apps/offer/utils.py
@@ -96,7 +96,7 @@ class Applicator(object):
             return offers
 
         for voucher in basket.vouchers.all():
-            if voucher.is_active() and voucher.is_available_to_user(user):
+            if voucher.is_active() and voucher.is_available_to_user(user=user, no_message_on_result=True):
                 basket_offers = voucher.offers.all()
                 for offer in basket_offers:
                     offer.set_voucher(voucher)
--- a/src/oscar/apps/offer/utils.py
+++ b/src/oscar/apps/offer/utils.py
@@ -96,7 +96,8 @@ class Applicator(object):
             return offers
 
         for voucher in basket.vouchers.all():
-            if voucher.is_active() and voucher.is_available_to_user(user=user, no_message_on_result=True):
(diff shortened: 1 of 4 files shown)


src/oscar/apps/offer/utils.py
```diff
@@ -96,7 +96,7 @@ def get_basket_offers(self, basket, user):
             return offers
 
         for voucher in basket.vouchers.all():
-            if voucher.is_active() and voucher.is_available_to_user(user):
```
codeinthehole: I'm not a fan of boolean variables like `no_message_on_result` - I would rather we used an index in the below conditional or read the two return values into variable - perhaps this:
```python
if voucher.is_active() and voucher.is_available_to_user(user=user)[0]:
```

codeinthehole:
Thanks for the PR - let me know what you think about my suggestion

jonasborges:
Hello, David!

I agree, maybe <code>no_message_on_result</code> wasn't a good idea.
However, I prefer to treat the result before the test condition. Like this:

```python 
is_voucher_available_to_user, _ = voucher.is_available_to_user(user=user)
if voucher.is_active() and is_voucher_available_to_user:
```

I think it's more clear to read. Don't you?

codeinthehole:
@jonasborges That's fine by me - make those changes and I'll merge it.

PS - it's slightly better to use `__` rather than `_` for throw-away variables. `_` is normally an alias for the gettext translation function so it's better to use something different.

jonasborges:
@codeinthehole It's done.

I'm new on that, so thanks for the tip! :D

maikhoepfel:
Thank you. I cherry-picked your changes across in 1d650a1

PR #1660, opened 2015-02-07
Fix for  #1652 (ensure_postcode_is_valid_for_country and required fields)

### LINKED ISSUE #1652: Postcode field raises ValidationError even if set as not required
I've set up
<pre>
OSCAR_REQUIRED_ADDRESS_FIELDS = ('first_name', 'last_name', 'line4', 'line1', 'country')
</pre>
but form does not validates empty postcode field. Seems like _ensure_postcode_is_valid_for_country_ [1] method doesnt know anything about this feature.

[1] https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/address/abstract_models.py#L262

okfish:
Do not validate postcode field only if it is not required and left empty. Validate anyway if user had filled in. Fixes #1652.

### DIFF
--- a/src/oscar/apps/address/abstract_models.py
+++ b/src/oscar/apps/address/abstract_models.py
@@ -1,6 +1,7 @@
 import re
 import zlib
 
+from django.conf import settings
 from django.db import models
 from django.utils.encoding import python_2_unicode_compatible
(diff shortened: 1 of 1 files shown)


maikhoepfel:
Thanks!

PR #1661, opened 2015-02-07
Fix `TEMPLATE_DIRS` test config

itbabu:
The config was picking the wrong directory so the custom test templates were not used during tests.

Notes

- Fix missing i18n load tags
- Remove mixed tab/space indentation
- Add two settings tests


codeinthehole:
Looks good to me - thanks.

PR #1664, opened 2015-02-09
Improve test coverage in checkout views

itbabu:
This is the second part of #1645.

This PR adds new functional tests for ShippingAddressView, PaymentDetailsView and ThankYouView.
I had to use a custom template to be able to test some functionalities of PaymentDetailsView.


mvantellingen:
Thanks!! I've merged it with some additional pep8 fixes

PR #1671, opened 2015-02-17
Improve Bootstrap 3 forms in the fontend site.

itbabu:
These are small improvements in the Bootstrap3 forms for the frontend website

Here the changes:
* fix `make css` paths
* fix error styles on form fields (focus and label should be red)
* align form buttons with fields
* remove `help-block` class inside alert box to avoid alert
  misalignment and text not readable
* avoid class `form-actions`: this class is removed from boostrap 3 and
  creates button misalignments
* use class `input-group` in the basket quantity form. Class
  `form-inline` applies only to forms within viewports that are
  at least 768px wide.
* move `well` divs outside forms
* remove mixed tab/space indentation in address_form.html
* proper style quantity field in whislist detail
* style label `required` using only css
* use `max-width` instead of `width` for fields
* update README
* compile .less files into .css files
 
A form for testing most of these changes could be the shipping Address form in the checkout process.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -98,9 +98,9 @@ compiledmessages:
 
 css:
 	# Compile CSS files from LESS
-	lessc --source-map --source-map-less-inline src/oscar/static/oscar/less/styles.less oscar/static/oscar/css/styles.css
-	lessc --source-map --source-map-less-inline src/oscar/static/oscar/less/responsive.less oscar/static/oscar/css/responsive.css
-	lessc --source-map --source-map-less-inline src/oscar/static/oscar/less/dashboard.less oscar/static/oscar/css/dashboard.css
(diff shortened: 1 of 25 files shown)


itbabu:
Whoops I just noticed that my less compiler changed all the roundings in the css files :-(
Is there a specific version of lessc to use?

mvantellingen:
Just struggled by it myself. I think we need to use lessc 1.7.5.

You can install this via `npm install less@1.7.5`. Will look through your changes asap

mbertheau:
You need less < 2.0.0:

```
npm install less@'<2.0.0'
```

It's mentioned in the docs here: https://github.com/django-oscar/django-oscar/blob/1eed7c8b43df3ab0a91a679bdb282e383f50f34b/docs/source/internals/contributing/development-environment.rst#writing-lesscss

mvantellingen:
Hi, I've rebased it on master and merged it! Thanks

PR #1672, opened 2015-02-17
Fix inconsistent Order.shipping_status

mvantellingen:
Refactor the shipping_status method to return a consistent last 
complete status. Since dictionaries have no guaranteed order, and also
since the shipping status queryset isn’t ordered the method didn’t
always return the actual last complete status.

Note that this should also be backported for a 1.0.2 release.

### DIFF
--- a/src/oscar/apps/order/abstract_models.py
+++ b/src/oscar/apps/order/abstract_models.py
@@ -1,4 +1,4 @@
-from itertools import chain
+from collections import OrderedDict
 from decimal import Decimal as D
 import hashlib
 
@@ -223,24 +223,28 @@ class AbstractOrder(models.Model):
 
     @property
     def shipping_status(self):
-        events = self.shipping_events.all()
+        """Return the last complete shipping event for this order."""
+
+        # As safeguard against identical timestamps, also sort by the primary
+        # key. It's not recommended to rely on this behaviour, but in practice
+        # reasonably safe if PKs are not manually set.
+        events = self.shipping_events.order_by('-date_created', '-pk').all()
         if not len(events):
             return ''
 
         # Collect all events by event-type
-        map = {}
+        event_map = OrderedDict()
         for event in events:
             event_name = event.event_type.name
-            if event_name not in map:
-                map[event_name] = []
-            map[event_name] = list(chain(map[event_name],
-                                         event.line_quantities.all()))
+            if event_name not in event_map:
+                event_map[event_name] = []
+            event_map[event_name].extend(list(event.line_quantities.all()))
 
         # Determine last complete event
         status = _("In progress")
-        for event_name, event_line_quantities in map.items():
+        for event_name, event_line_quantities in event_map.items():
             if self._is_event_complete(event_line_quantities):
-                status = event_name
+                return event_name
         return status
 
     @property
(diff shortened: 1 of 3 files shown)


maikhoepfel:
Wowsa. Will look at it tomorrow.

src/oscar/apps/order/abstract_models.py
```diff
@@ -223,24 +223,23 @@ def shipping_tax(self):
 
     @property
     def shipping_status(self):
-        events = self.shipping_events.all()
+        events = self.shipping_events.order_by('-date_created').all()
```
maikhoepfel: We should order by PK here as well, in case two timestamps are identical. Do we have a guarantee that PKs are created in order?

src/oscar/apps/order/abstract_models.py
```diff
@@ -223,24 +223,23 @@ def shipping_tax(self):
...
+                event_map[event_name] = []
+            event_map[event_name].extend(list(event.line_quantities.all()))
 
         # Determine last complete event
         status = _("In progress")
-        for event_name, event_line_quantities in map.items():
+        for event_name, event_line_quantities in event_map.items():
             if self._is_event_complete(event_line_quantities):
-                status = event_name
+                return event_name
```
maikhoepfel: This now returns the first "completed" status, not the last. Probably unintended?

maikhoepfel:
I left a few comments. Apart from that, LGTM. Feel free to merge when my questions are addressed and backport. I'm happy to issue a 1.0.2 release.

Reminder to self: add note to release notes.

src/oscar/apps/order/abstract_models.py
mbertheau: Although in practice they often are, the common requirement for PKs is only uniqueness. An order on them may not even exist.

maikhoepfel: The [Django docs](https://docs.djangoproject.com/en/dev/topics/db/models/#automatic-primary-key-fields) aren't very explicit about it, but still leave me with the impression that the primary key will only ever go up. "Auto increment" suggests the same, doesn't it?

mbertheau: Well, auto increment is a term from MySQL and refers to the practical default implementation of primary keys. A PostgreSQL sequence-based primary key exhibits the same characteristic. I don't know about oracle.

However a primary key may also be not autogenerated, i.e. a category slug. In that case it's not guaranteed to be created in order.

However what we really need to define here is which event we want in case that there are two with the same date_created. We could also define a unique constraint on date_created.

src/oscar/apps/order/abstract_models.py
mvantellingen: Since the queryset orders on the `date_created` descending the last complete status is returned.

I'll add some extra tests for this

mvantellingen:
This adds some new factories with Factory-boy to create orders. In a separate PR I'd like to spit the newfactories to several modules.

src/oscar/apps/order/abstract_models.py
```diff
@@ -223,24 +223,24 @@ def shipping_tax(self):
 
     @property
     def shipping_status(self):
-        events = self.shipping_events.all()
+        """Return the last complete shipping event for this order."""
+        events = self.shipping_events.order_by('-date_created').all()
```
maikhoepfel: Could you add sorting by primary key as well, and leave a comment to explain why?
e.g. `# As safeguard against identical time stamps, we also sort by primary key. It's not recommended to rely on this behaviour, but in practice reasonably safe if PKs are not manually set.`
I confirmed this behaviour with @aaugustin.

codeinthehole:
Thanks for this - looks good.

mvantellingen:
@maikhoepfel I've addressed your comments regarding the additional order field and added a comment.

Thanks for reviewing

Issue #1119, opened 2014-01-09
Rethink handling of category slugs/full_name fields 


maiksprenger:
full_name and slug, respectively and can quickly hit the limit of 255
characters. Increasing the limit beyond that is not really feasible, at
least with MySQL.
#929 reported this and attempted a fix by increasing the limit beyond 255 chars, but that breaks with MySQL. The limit is somewhere around 760 bytes, but as with Unicode we never actually know how many bytes are consumed, we can't considerably increase beyond 255 chars.

I've also attempted moving to TextFields, but similar issues crop up; MySQL can't have an index on TextFields even with a max_length set: http://stackoverflow.com/questions/11129309/how-do-i-specify-an-index-for-a-textfield-in-django-with-a-mysql-backend

As we're thinking about how to translate category URLs, this should be given a proper think for a good solution.

specialunderwear:
this was solved in #1674

Issue #561, opened 2013-03-11
Allow a category slug to be edited


codeinthehole:
Follow up to https://github.com/tangentlabs/django-oscar/pull/322

Category slugs are currently not editable, but generated from names.  We should allow them to be edited.  

We will probably need a new field to track which categories have customised names.

PR #1674, opened 2015-02-20
Overhaul category slug handling

### LINKED ISSUE #1670: Category handling overhaul bucket
At my current job, we're running into some issues with category handling. I get some time to improve the state of things in Oscar.

Goals:
- Easily allow editing slugs independently of names. slugs are important for SEO and might need to match legacy URLs, so the auto-generating of slugs should be an optional step. (#561)
- Address MySQL size limits (#1119). 

I haven't thought it through just yet, but I think removing the ``full_slug``/full name denormalization is a good first step, as it nicely simplifies the slug updating process. I hope that we can address any performance issues with a simple cache.

I'll attempt to watch out for:
- not increasing query counts considerably
- not breaking existing URLs

Bonus goals:
- Ensure that it's easy to translate categories.
- Ensure that it's easy to display them alphabetically.
- Ensure that it's easy to not display categories without products.

Further reading:
- https://github.com/django-oscar/django-oscar/pull/322 worked towards editing the slugs independently. But we're seeing issues where the parent's save method still overrides the slugs.

maikhoepfel:
This is the PR to implement the changes suggested in #1670. I didn't run into any problems. The only fiddly bit is that we need to add a data migration to remove the ancestor slugs from the ``slug`` field. But doing that seemed easier than any other approach I could come up with.

When this is merged, we can close #1670, #561 and #1119. I'll leave it open for feedback for a week and will then merge if no blockers are mentioned till then.

I added a South schema migration because we haven't yet dropped Django 1.6 support and I don't want to fail our tests, but really this PR is written with removed South support in mind. I have a work-in-progress branch to remove Django 1.6 support lying around; I hope to be able to finish it soon.


docs/source/releases/v1.1.rst
```diff
@@ -60,6 +62,40 @@ Minor changes
 Backwards incompatible changes in 1.1
 -------------------------------------
 
+Categories refactor
+~~~~~~~~~~~~~~~~~~~
+The Category model contain two denormalisations to improve performance: it
```
mbertheau: s/contain/contained/?

docs/source/releases/v1.1.rst
```diff
@@ -73,6 +109,17 @@ Misc
...
+.. warning::
+   This release contains a data migration for category slugs. It will get
+   run automatically if you haven't forked the catalogue app. If you have
+   forked it, it is critical you run a copy of that migration when upgrading.
+
+* Catalogue:
+  - This app contains model changes. They should be picked up automatically.
+  - This app contains a data migration that you must run:
+    ``0003_data_migration_slugs``. It shouldn't matter whether it is executed
+    before or after the schema migration.
```
mbertheau: What do you mean by 'It shouldn't matter whether it is executed before or after the schema migration.", and which schema migrations does that sentence refer to?

src/oscar/apps/catalogue/views.py
```diff
@@ -170,14 +170,35 @@ def get(self, request, *args, **kwargs):
...
+            concatenated_slugs = self.kwargs['category_slug']
+            slugs = concatenated_slugs.split(Category._slug_separator)
+            try:
+                last_slug = slugs[-1]
+            except IndexError:
+                raise Http404
+            else:
+                for category in Category.objects.filter(slug=last_slug):
+                    if category.full_slug == concatenated_slugs:
+                        return category
```
mbertheau: It sounds like we might get bug reports for this later, if I understood correctly:

Given two categories: `topic1 -> thing` and `topic2 -> thing`;
Before, `/catalogue/topic/thing` would give the topic -> thing category;
Now it gives whichever of the two categories happens to be sorted first.

docs/source/releases/v1.1.rst
maikhoepfel: The migration that changes models. I'll try to clarify.

src/oscar/apps/catalogue/views.py
maikhoepfel: Ah, this docstring is outdated. I don't enforce uniqueness any more. Will fix.

But I'm not sure I understand your comment. Are `topic1` and `topic2` slugs? Then `catalogue/category/topic/thing` will never lead anywhere. I compare against ``full_slug``, so there really should only be one.

I'm tempted to deprecate this old behaviour. I don't think it's used anywhere (Oscar never hands out a URL without a primary key), it's causing problems, and I doubt there's SEO benefits. Amazon includes random data all across their URLs. What do you think?

docs/source/releases/v1.1.rst
maikhoepfel: Will fix.

src/oscar/apps/catalogue/views.py
mbertheau: Ah, indeed, I missed the fact that you compare against `full_slug`. Should be alright then :)

I'd be alright with removing the slug-only variant, but I'd check history to see whether there's more clue as to why that was introduced in the first place.

maikhoepfel: It's how we used to do things in Oscar 0.6 or so, and was kept around since then.

docs/source/releases/v1.1.rst
```diff
@@ -73,6 +109,20 @@ Misc
...
+
+.. warning::
+   This release contains a data migration for category slugs.  If you have
+   forked it, it is critical you run a copy of that migration when upgrading.
+
+Catalogue
+  * This app contains model changes.
+  * This app contains a data migration: ``0003_data_migration_slugs``.
+    It shouldn't matter whether you apply this migration before or after
+    the model changes.
```
mbertheau: Do you mean the 0002 and 0004 migrations introduced in this PR? If you do, I find it confusing that you mention migration application order. Usually migrations are applied according to their dependencies. Why is it important to mention that the order doesn't matter?

maikhoepfel: Because if you have forked the app, you need to manually declare the dependencies for the data migration. I hoped to have made that clear in my updates to ``topics/updating``. How can I phrase it better?

maikhoepfel:
I added a commit to deprecate accessing a category page without a primary key.

PR #1675, opened 2015-02-22
Fix the <br> tags in commtype_order_placed_body.html template

lijoantony:
Two of the <br> tags had a typo, which affected the email content.


mbertheau:
Thanks :)

PR #1676, opened 2015-02-22
Eliminate Garmsby and FP Sports from the list of Oscar projects.

xpostudio4:
Both sites are not available anymore: 

You can check their all url: 
* garmsby.co.uk (it's offline)
* www.fpsport.it  (was replaced with a static landing page)


codeinthehole:
Thanks - been meaning to do that for ages.

xpostudio4:
Where can I find another list of websites that use Oscar?, I would like to update it.

PR #1677, opened 2015-02-23
Unify to (basket, user=None, request=None) argument order in Applicator

mbertheau:
This order offers the simplest API for common needs - as evidenced by the
fact that the apply_offers crook could be removed from two test files,
saving on imports as well.


codeinthehole:
LGTM. Needs a note in the release notes since it's not backwards compatible.

mbertheau:
Added a bit to the release notes.

Please merge if satisfied :)

Thanks

maikhoepfel:
Cool!

PR #1678, opened 2015-02-24
Clerk

mbertheau:
This branch builds on #1677 and #1423. This PR is not meant for merging yet – at least documentation is missing, but also more complicated payment methods haven't been implemented yet and will most probably require API changes.

From the commit message of the first commit:

Introduce a friendly clerk to place an order

The clerk provides a pure python API to create an order, it doesn't need a request or a view. This is useful for creating orders in tests, factories, celery tasks or manually on the python command line.

The CheckoutFlow has been adapted to use the Clerk to place an order. To that end the payment has been extracted as a separate step to `Clerk.take_payment`.

Payment method implementations now don't have to know as many internals to plug into oscar as before. For payment methods without further user interaction it is now sufficient to return a list of `payment.api.PaymentSource` instances and a list of `payment.api.PaymentEvent` instances from `OrderPlacementMixin.handle_payment`. These classes are lightweight `namedtuples` that can be used without access to the Django ORM.

Other payment methods should raise `payment.exceptions.RedirectRequired` in `OrderPlacementMixin.handle_payment`.  The result view should mix in `checkout.flow.CheckoutFlow` and call `handle_payment_failed(message)` or `handle_payment_successful(basket_id, payment_sources, payment_events)`.  `handle_payment_successful` will care about restoring the basket and placing the order.

Here's the code we use to create an order in the tests:

``` python
Clerk = get_class('checkout.clerk', 'Clerk')
Basket = get_model('basket', 'Basket')
Selector = get_class('partner.strategy', 'Selector')

basket, __ = Basket.objects.get_or_create(owner=user)
basket.strategy = Selector().strategy(request=None, user=user)
basket.flush()
basket.add_product(product)
Applicator().apply(basket, user)

clerk = Clerk(user, basket)

clerk.receive_payment(payment_sources, payment_events)
clerk.place_order()
```

The shipping address and other data can be set on the clerk like it's done in `OrderPlacementMixin.get_clerk`.

I'm not completely happy with the API yet. Redirect payment method result views should have the possibility to just raise PaymentError or UnableToTakePayment just like non-redirect methods can do in `OrderPlacementMixin.handle_payment`. There's no inherent need for these APIs to differ. A dream would be if you could just put

``` python
OSCAR_PAYMENT_METHODS = ['django_oscar_paypal', 'django_oscar_stripe']
```

in the settings, without the need to override obscure mixins.

Example for the redirect:
https://github.com/machtfit/adyen/blob/eab53be6a35d246467580a587f5aa5436f9a63c2/oscar_adyen/mixins.py

Example for the result:
https://github.com/machtfit/adyen/blob/eab53be6a35d246467580a587f5aa5436f9a63c2/oscar_adyen/views.py


sasha0:
Closing since PR is outdated, has conflicts, lacking documentation and test coverage. Although as POC looks quite interesting. 👍

PR #1682, opened 2015-02-25
Update all ModelForm’s to specify Meta.fields

mvantellingen:
Specifying `excludes` is deprecated in Django 1.7 and not supported in
Django 1.8.


maikhoepfel:
Thanks for doing this!

Issue #1684, opened 2015-02-26
Product dashboard UPC search returns empty set for variants


john-parton:
If you search by a variant's UPC, it returns no results. Expected behavior is to return the parent row.

I have a fix for this and will submit a PR promptly.

maikhoepfel:
I agree this would be nice to have.

PR #1685, opened 2015-02-26
Fix false-negative for UPCs when UPC matches a variant.

john-parton:
Fixes https://github.com/django-oscar/django-oscar/issues/1684

I removed the `exists()` logic. I can update the pull request with that logic in place if you think it's really necessary.

### DIFF
--- a/oscar/apps/dashboard/catalogue/views.py
+++ b/oscar/apps/dashboard/catalogue/views.py
@@ -115,9 +115,14 @@ class ProductListView(SingleTableMixin, generic.TemplateView):
         data = self.form.cleaned_data
 
         if data.get('upc'):
-            # Filter the queryset by upc
-            matches_upc = Product.objects.filter(upc__icontains=data['upc'])
-            queryset = queryset.filter( Q(id=matches_upc.values('id')) | Q(id=matches_upc.values('parent_id')) )
+            # If there's an exact UPC match, it returns just the matched
+            # product. Otherwise does a broader icontains search.
+            qs_match = queryset.filter(Q(upc=data['upc']) | Q(parent__upc=data['upc']))
+
+            if qs_match.exists():
+                queryset = qs_match
+            else:
+                queryset = queryset.filter(Q(upc__icontains=data['upc']) | Q(parent__upc__icontains=data['upc']))
 
         if data.get('title'):
             queryset = queryset.filter(title__icontains=data['title'])
--- a/oscar/apps/dashboard/catalogue/views.py
+++ b/oscar/apps/dashboard/catalogue/views.py
@@ -115,13 +115,9 @@ class ProductListView(SingleTableMixin, generic.TemplateView):
         data = self.form.cleaned_data
 
         if data.get('upc'):
-            # If there's an exact UPC match, it returns just the matched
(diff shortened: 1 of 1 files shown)


oscar/apps/dashboard/catalogue/views.py
```diff
@@ -115,13 +115,9 @@ def apply_search(self, queryset):
         data = self.form.cleaned_data
 
         if data.get('upc'):
-            # If there's an exact UPC match, it returns just the matched
-            # product. Otherwise does a broader icontains search.
-            qs_match = queryset.filter(upc=data['upc'])
-            if qs_match.exists():
-                queryset = qs_match
-            else:
-                queryset = queryset.filter(upc__icontains=data['upc'])
```
maikhoepfel: Instead of your changes, what do you think of just doing the following?

``queryset = queryset.filter(Q(upc__icontains=data['upc']) | parent__upc__icontains=data['upc']))``

That way, we don't redefine the queryset and keep the ``exists`` logic. I also find it slightly more readable.

john-parton:
Your proposed change won't work, because `queryset` is already filtered with the `browsable` manager, which doesn't include variants, and `Q(parent__upc__icontains=data['upc'])` will only match variants. (Load up https://github.com/john-parton/django-oscar/commit/a73df0c895780a1c1c551b1f1bb0c7f30979266f and see for yourself.)

I've pushed a change that includes the exact UPC matching with `exists` logic.

I understand that the code isn't obvious. Often, I will write the SQL to generate a query by hand, and then go back and write the Python to get Django's ORM to dance. This sometimes results in code that doesn't appear to be idiomatic Python/Django, because Django encourages JOINs over subqueries.

Let me know if you want me to make any other changes.

Thanks for the feedback.

john-parton:
Do you need me to do anything else to accept this PR?

jmound:
@john-parton I was searching for solutions to this issue as well and was glad to find your fix.  The only negative, which I'm sure you've seen too is how the matched result isn't displayed in the results view but its parent. I was thinking of building a more complex result view that would include the actual matched result(s), but your approach gets 95% of the job done with a few lines of code.

+1 on the merge (and a thank you). For now, I have to fork the dashboard app just for this fix. Perhaps there is some reluctance to merge this in without being able to turn off the behavior, and it should be configured with a setting?

john-parton:
@jmound I thought about modifying it so that children could display in search, but I wanted the behavior to be consistent: either children are always displayed in search or children are never displayed in search. Having children only be visible if you do a UPC search, but otherwise not being visible is too context-sensitive for my tastes (and somewhat violates the principle of least surprise). I wasn't sure what the ramifications of always including children would be, so I left it as is. Plus, I figured keeping my PR minimal would ensure that it gets accepted.

If there's consensus for doing it differently, I would be more than happy to submit another PR.

jmound:
@john-parton Totally agree on the design call.

maikhoepfel:
Heya. I squashed this PR into https://github.com/django-oscar/django-oscar/commit/f0ba48d8ef396e3c8a99592b6ad23b1141994c05 and just pushed it onto master. Thank you for your work, and sorry for being so slow to respond!

PR #1688, opened 2015-03-04
Use a regular settings file for tests

maikhoepfel:
I'm not sure why @codeinthehole, back in the days,  opted to manually
configure settings instead of using a regular settings file for the test
suite. Most likely the reasons for doing so have disappeared.

So I turned our tests/config.py into a normal settings file, and use that
accordingly. That means we do things the standard Django way, and it's easier
to drop into a session with the test settings.


maikhoepfel:
@mvantellingen I tried adding the new settings file to the `[pytest]` section in `setup.cfg`, but somehow that didn't work. But I figure it can't be that hard.

mvantellingen:
@maikhoepfel we can make it pytest compatible again with the following tests/conftest.py code:

```
from django.conf import settings

from . import settings as test_settings

def pytest_configure():
    settings.configure(**test_settings.__dict__)
```

maikhoepfel:
Currently fails for Django 1.6 (expectedly), but once @mbertheau got to have a look at the Django 1.6 PR, this should pass.

maikhoepfel:
Rebased and implemented a variant of @mvantellingen's fix to keep pytest working. And now merged.

PR #1690, opened 2015-03-05
Updated some templates in customer/emails, for the sake of consistency!

ka7eh:
Updated some templates in customer/emails so they extend 'customer/emails/base.*' instead of 'oscar/customer/emails/base.*'


mbertheau:
Thanks, good catch! :)

Issue #1136, opened 2014-01-16
Improve signals


codeinthehole:
I'm writing a CRM extension for Oscar and want to consolidate and improve the signals available.  Specifically:
- [ ] Introduce new signals for key events (eg starting checkout, failing payment)
- [ ] Ensure searches raise a signal correctly
- [ ] Ensure signals pass the request instance where applicable
- [ ] Use normal imports, not `get_class`.  I don't think it's sensible to dycamically swap signals.

PR #1691, opened 2015-03-05
Update Basket Signals

itbabu:
This PR follows [this google group topic](https://groups.google.com/forum/#!topic/django-oscar/X2QpKi_9bhs)

Here the changes:
* Add ``basket_removal`` signal
* Add quantity and purchase info to signals arguments
* Handle ``basket_removal`` and ``basket_addition`` on basket variation, remove line, save for later and reorder events
* Add basket and reorder tests
* Add names to basket update buttons to be able to reference them in tests
* Update documentation

A view notes:
* I didn't touch the analytics app, now the ``basket_addition`` signal would be raised more frequently and it would have a slightly different meaning. Is this ok?
* I didn't change the current basket behaviour as new tests highlight (apart for signals handling):
    * basket quantity variations are ignored if a user click save for later
    * basket quantity variations are saved if a user clicks remove (line)
* I got the ``purchase_info`` from basket line when this was possible. Is this the best way to obtain it?
* This PR is related  to #1136


mvantellingen:
Cool! Is it possible to move the code out of the view and in a separate utility function? Or perhaps move it to the form?

itbabu:
Hi Michael,

Here some thoughts about the ``BasketView``:

I could move  [L247-L265](https://github.com/itbabu/django-oscar/blob/414ad1ccfaa33b4be2c84cc34ac5e0e9a18043ab/src/oscar/apps/basket/views.py#L247-L265) lines outside ``formset_valid()` in a proper method.
The  other signal related lines are simple send calls.

The tricky part is that signals are raised  only if specific conditions are met:

* A `save_for_later` event disregards any other change in the basket, is available to signed in users only, raises a removal signal for its line.

* The `remove` event saves any other change in the basket, is available to anyone, raises a removal signal for its line, raises addition/removal signals.

* The `update` event saves any other change in the basket, is available to anyone, raises  addition/removal signals.

Using form and formset for raising signals means:
* pass ``request`` and maybe others parameters to forms and formset
* Modify the view behaviour since when `save_for_later` is `True` `formset.save()` is not called.
* loop multiple times the formset ( I'm not 100% sure about this atm)

itbabu:
After further considerations I'm going I'm going to move the ``basket_removal`` and ``basket_addition`` related code inside ``BasketLineForm`` and ``BaseBasketLineFormSet``.

itbabu:
The basket signal related logic is outside ``BasketView`` now.

The logic is splitted between ``BasketLineForm`` and  ``BaseBasketLineFormSet``.

I could use ``BaseBasketLineFormSet`` only. 

The pro is that I could reduce all the logic into one utility function.
The con is that I should do an additional loop over the changed forms to get the signals arguments and send the signals.

mbertheau:
When I think about how this can be used from django-oscar-api, I think it'd be more cleanly separated and loosely coupled if the signal sending code didn't need a view, a request or a form.

itbabu:
@mbertheau Late response...
I did some refactoring this week. 
Now the basket variation signal sending code doesn't need a form or a view. I made the request optional.

auvipy:
@itbabu could u plz fix the merge conflict?

itbabu:
@auvipy Done.
@mvantellingen Could you review this PR when you have some time? It got lost in the PR bucket for some months :)

sasha0:
@django-oscar/core I'm not very enthusiastic regarding signals extension and agree with the point - https://lincolnloop.com/blog/django-anti-patterns-signals/, so would pretty much like to know your opinion about this.

PR #1694, opened 2015-03-08
Use oscar.test.contextmanagers.mock_signal_receiver in review tests

itbabu:
This PR simply removes few lines of duplicated code.


maikhoepfel:
Ah, good spot. Thanks.

PR #1696, opened 2015-03-11
Added pagination to the voucher dashboard page

maerteijn:
Because the page is really, really slow with a lot of vouchers


codeinthehole:
Thanks for that.

Issue #1698, opened 2015-03-19
Range can never include child products


john-parton:
There is no good way to apply a discount to a child product.

Here's a scenario: You run a bookstore and want to discount hard cover books, but not paperbacks. Your catalog is set up so that paperback and hardcovers are children. You want to discount hardcovers but not paperbacks. One approach would be to create a new `Range` and add all the hard covers to the new `Range`, but ranges are always associated with parents or standalone products.

`Range.contains_product` has the following lines:

```python
# Child products are never part of the range, but the parent may be.
if product.is_child:
    product = product.parent
```

monikasulik:
The pull request was merged so this issue can be closed

mvantellingen:
Yeah sorry forgot to close it.

PR #1699, opened 2015-03-19
Allow for children to be explicitly included in ranges. (Allows for rang...

john-parton:
Fix for https://github.com/django-oscar/django-oscar/issues/1698

There should be just one backwards incompatible changes: this changes the parameter that is passed to `load_proxy(self.proxy_class)().contains_product(product)` so any custom `Condition` that overrides `contains_product` will have to make sure it can handle accepting a child product. This should probably be documented.

### DIFF
--- a/src/oscar/apps/offer/abstract_models.py
+++ b/src/oscar/apps/offer/abstract_models.py
@@ -824,9 +824,6 @@ class AbstractRange(models.Model):
         Default display_order for a new product in the range is 0; this puts
         the product at the top of the list.
         """
-        if product.is_child:
-            raise ValueError(
-                "Ranges can only contain parent and stand-alone products.")
 
         initial_order = display_order or 0
         RangeProduct = get_model('offer', 'RangeProduct')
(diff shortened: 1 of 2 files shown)


codeinthehole:
Looks good - could you add a test to verify the expected behaviour?

john-parton:
I'm not very good at testing, so sorry in advance if this isn't quite right.

* Added test cases for ranges that include children
* Removed checking to see if a product is a child

Let me know if you need anything else. Thanks.

mvantellingen:
Hi @john-parton, could you rebase this on master? Then i'll get this merged. 

Thanks! :-)

john-parton:
I had some trouble rebasing it, but I think I got it. (I'm still accustomed to the bzr and svn ways of doing things, unfortunately.)

Let me know if you need anything else.

Thanks,
John

mvantellingen:
Hi John, Thanks for the rebase! I've squashed the 4 commits to a single on and commited it on master.

PR #1700, opened 2015-03-21
Allow robots to index the ecommerce pages again.

itbabu:
All e-commerce pages would be removed from search results if someone uses oscar default base.html. 

I noticed this while upgrading a project to 1.0.x.
Could this be backported to the 1.0 branch?

**Changes**:
* Remove ``none``: it's equivalent to ``noindex, nofollow``
* Use ``nocache`` (Bing) and ``noarchive`` (Google, Yahoo): we don't want search engines to cache copy of web pages and products prices in search results.

**References**:
https://developers.google.com/webmasters/control-crawl-index/docs/robots_meta_tag
https://developer.mozilla.org/it/docs/Web/HTML/Element/meta

### DIFF
--- a/src/oscar/templates/oscar/base.html
+++ b/src/oscar/templates/oscar/base.html
@@ -13,7 +13,7 @@
         <meta name="description" content="{% block description %}{% endblock %}" />
         <meta name="keywords" content="{% block keywords %}{% endblock %}" />
         <meta name="viewport" content="{% block viewport %}width=device-width{% endblock %}" />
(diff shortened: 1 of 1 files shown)


codeinthehole:
Ouch, well spotted. I think that was accidentally introduced in 64e21b1eb88e888ad59073c44997b6fc5101da25 - it was meant for the dashboard only.

mvantellingen:
Yep my bad :-( But I remember that this was already fixed by @maikhoepfel, can't find the issue however

Issue #1697, opened 2015-03-17
Dashboard: double escaping in category description


jsvini:
In the dashboard when you add a Category description with latin characters they will be saved scaped.
example: the word ```"Descrição"``` will be saved in the database like this ```"Descri&ccedil;&atilde;o"```

The issue is that in the dashboard when the template is rendered, the description will be escaped again, wich will produce this: ```Descri&amp;ccedil;&amp;atilde;o```, so instead of showing ```"Descrição"``` it will show ```"Descri&ccedil;&atilde;o"``` to the user.

Add the <a href="https://docs.djangoproject.com/en/1.7/ref/templates/builtins/#safe" target="_blank">```safe```</a> filter to ```record.description``` in the <a href="https://github.com/django-oscar/django-oscar/blob/0ce8443d181c3da10cf249f103779ed63691f6be/src/oscar/apps/dashboard/catalogue/tables.py#L53">CategoryTable</a> class fixed my problem.

discort:
Hi David and Vinícius! Field description fills with tinymce. I found the sames problem in stackoverflow 
http://stackoverflow.com/questions/12337188/django-tinymce-and-unicode

You need to add  entity_encoding : "raw (http://www.tinymce.com/wiki.php/Configuration:entity_encoding) to tinymce init
https://github.com/django-oscar/django-oscar/commit/5412bb119e7e05b69b8ee98da7ab6adc11d5a742

mvantellingen:
Should be fixed with #1702

PR #1702, opened 2015-03-26
added raw encoding to tinymce

discort:
https://github.com/django-oscar/django-oscar/issues/1697#issuecomment-86602991


mvantellingen:
Thanks!

PR #1705, opened 2015-04-01
Fix race condition when tracking num_basket_additions on a voucher

pipermerriam:
### What is the problem / feature ?

The current implementation that maintains the counts for `Voucher.num_basket_addition` is subject to race conditions.  If two users apply the same voucher at approximately the same time, one of the additions can easily be lost.  The same applies to voucher removal.

### How did it get fixed / implemented ?

Instead of using `voucher.save()` change to use an `.update()` statement using an `F()` statement.

https://docs.djangoproject.com/en/1.7/ref/models/queries/#avoiding-race-conditions-using-f

This delegates the increment/decrement of this count to the database which allows for this operation to be safe from multiple concurrent updates.

### How can someone test / see it ?

I don't know how to write a test case that exposes this bug without some nasty mocking, `time.sleep()` and threading.  Ick.

*Here is a cute animal picture for your troubles...*

cat-wearing-a-pointy-hat

### DIFF
--- a/src/oscar/apps/voucher/receivers.py
+++ b/src/oscar/apps/voucher/receivers.py
@@ -1,14 +1,19 @@
+from django.db.models import F
 from oscar.apps.basket import signals
 
 
 def track_voucher_addition(basket, voucher, **kwargs):
     voucher.num_basket_additions += 1
-    voucher.save()
+    voucher.__class__._default_manager.filter(pk=voucher.pk).update(
+        num_basket_additions=F('num_basket_additions') + 1,
+    )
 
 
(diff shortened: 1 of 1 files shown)


pipermerriam:
I wasn't exactly sure whether I should use `get_class('voucher', 'voucher')` or go with `voucher.__class__`.  Feel free to push me another direction if you guys want.

mvantellingen:
Thanks! using `__class__` seems fine to me. I've added some unittests and merged it via a separate branch due to some error on my side :-)

PR #1707, opened 2015-04-02
Use fetch_for_line where appropriate

adamend:
I'm working on a project where pricing depends on Line attributes. Therefore we need to calculate prices in our Strategy's `fetch_for_line`. While working through this I found that `fetch_for_line` is not called in some places where it seems like it should be. So far I've found two places that have access to a Line, but are calling `fetch_for_product(line.product)` instead of `fetch_for_line(line)`.

- `oscar.apps.basket.AbstractLine.purchase_info` - The symptom here was that prices in the basket were not displaying properly, since they were ultimately using `fetch_for_product`

- `oscar.apps.checkout.session.CheckoutSessionMixin.check_basket_is_valid` - The symptom here was that setting the availability policy in `fetch_for_line` was having no effect (specifically, making it Unavailable did not prevent checkout).

### DIFF
--- a/src/oscar/apps/basket/abstract_models.py
+++ b/src/oscar/apps/basket/abstract_models.py
@@ -730,8 +730,8 @@ class AbstractLine(models.Model):
         """
         if not hasattr(self, '_info'):
             # Cache the PurchaseInfo instance.
-            self._info = self.basket.strategy.fetch_for_product(
-                self.product, self.stockrecord)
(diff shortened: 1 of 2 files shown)


adamend:
Note: The test failure seems to be a Travis CI - Github connection error. I can't seem to re-run the tests, I assume I don't have the proper permissions.

mvantellingen:
Merged it! Thanks :+1:

Issue #1710, opened 2015-04-08
Adding django 1.8 support


auvipy:

krzysztofp:
+1

karol-gruszczyk:
+1

blager:
+1

tpro:
+1

bartlomiejborzucki:
+1

mvantellingen:
Yup, if no one beats me to it i'll look at it this weekend

auvipy:
while changes between 1.6 and 1.7 were not trivial, changes between 1.7 and 1.8 should not be that difficult to support togather.

mvantellingen:
Well one of the issues is that not all our dependencies support 1.8 yet. And we also need to decide if we integrate this in Oscar 1.1.0 (which will include Bootstrap 3) or release it as a 1.0.4 (thus make it an easier upgrade path). I'll leave this decision to @maikhoepfel and @codeinthehole

maikhoepfel:
Hi all, I'm happy to announce I just merged #1711. Django 1.8 is now supported on the master branch. A client of mine is using Oscar and wants to move to Django 1.8 pretty soon, so I'm hopeful I'll get time to package Oscar 1.1 soon.

maikhoepfel:
We do have some work left to do: https://github.com/django-oscar/django-oscar/issues/1728

PR #1711, opened 2015-04-08
Django 1.8 support

mvantellingen:
Note that this is work in progress!

See #1710


maikhoepfel:
Closing because it is superseded by #1712. Thank you for looking into this!

I think it's a good idea if we add pytest-django support in a different commit. That's why I didn't use any of the commits from here; hope you don't mind.

PR #1714, opened 2015-04-14
Fix #1689

### LINKED ISSUE #1689: Category data migration throws an error
I just migrated my local Oscar installation and the recent [data migration](https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/catalogue/migrations/0003_data_migration_slugs.py) on categories threw an AttributeError on ``Category._slug_separator``. I think I don't fully understand what model I have at that point in time.

No stacktrace because I was on a train and writing down from memory.

vovanbo:
Use Oscar's core method `get_model` to get Category model in data migrations.

**Explanation:**
Documented way to data migration in Django 1.7+ is using `apps.get_model` in forward/backward migrations. But this is not the same method as `get_model` from `django.apps.apps`. In migrations operations used ModelState instances which actually is:

> Represents a Django Model. We don't use the actual Model class
> as it's not designed to have its options changed - instead, we
> mutate this one and then render it into a Model as required.

It's from [here](https://github.com/django/django/blob/1.7.7/django/db/migrations/state.py#L140-L142).

So, `apps.get_model` return *fake* model, which actually is deconstructed model class instance (ModelBase) with no properties, no class methods, only with model fields.

Side effect of using Oscar's `get_model` is requirement to Category model's properties `full_slug` and `_slug_separator` will not be changed/renamed in future.

### DIFF
--- a/src/oscar/apps/catalogue/migrations/0003_data_migration_slugs.py
+++ b/src/oscar/apps/catalogue/migrations/0003_data_migration_slugs.py
@@ -2,17 +2,18 @@
 from __future__ import unicode_literals
 
 from django.db import models, migrations
+from oscar.core.loading import get_model
 
 
 def remove_ancestor_slugs(apps, schema_editor):
-    Category = apps.get_model('catalogue', 'Category')
+    Category = get_model('catalogue', 'Category')
     for category in Category.objects.all():
         category.slug = category.slug.split(Category._slug_separator)[-1]
         category.save()
 
 
 def add_ancestor_slugs(apps, schema_editor):
-    Category = apps.get_model('catalogue', 'Category')
+    Category = get_model('catalogue', 'Category')
     for category in Category.objects.all():
         category.slug = category.full_slug
         category.save()


maikhoepfel:
Thanks a bunch for digging into this. I'm not sure if actually loading the model is safe at this point though. IIRC the Django docs example for data migrations does what the original code does. Maybe we should only use Oscar's `get_model` to be able to fetch the `_slug_separator`? What do you think?

mbertheau:
We probably have to put the slug separator in a setting and use that instead of the model.

vovanbo:
First, we should decide how to be with backward migration, because it using `full_name` property. This property is strongly depends on treebeard's MP nodes, and therefore using RawSQL query (as one of probably solutions) for backward migration is complicated and may be wrong way.

And yes, separators can be moved to settings. But seems to it's so low-level settings that I don't think it will be used anywhere else. And again, general part of issue is backward migration.

mvantellingen:
Seems to be already fixed with 60b13

Issue #1717, opened 2015-04-16
oscar_fork_app issue when forking to a current dirrectory using dot(.)


manevant:
Hello.
When im trying to fork some app to current dirrectory with blank `folder_path` argument i get a `CommandError` exception from `../oscar/management/commands/oscar_fork_app.py` with error:
> "You must specify an app label and a folder to create  the new app in"

While im try to fork with dot (like a `./manage.py oscar_fork_app promotions .` ) in `__init__.py` file and `config.py` file i got a invalid path.
```python
# __init__.py
default_app_config = '..promotions.config.PromotionsConfig'

# config.py
from oscar.apps.promotions import config
class PromotionsConfig(config.PromotionsConfig):
    name = '..promotions'
```

Now I am preparing to fix this error. 
**Please specify, how can do it in right way?**

--edit--

I did commit that solves the problem, what do I do next?
p.s.it's my first time contributing, forgive my possible retarding :)

maikhoepfel:
Hello @manevant, sorry for being slow to get back to you. Thank you for the PR as well.

Can you explain why you'd like to fork the app in the top level folder? I think it's pretty common to have an ``apps`` folder under which you keep your Oscar apps. We had that structure in mind when building the ``fork_app`` command. I think it should technically be possible to not do that and keep them all at root level, but I'm not convinced it's a good idea.

Could you try to explain your motivation a bit more? 

Related: I do think we have an issue where we should be looking at absolute paths, and expand relative paths into such. But I'm not sure right now.

manevant:
Hello @maikhoepfel !

Only one reason why i want to copy forked apps into a root directory - our project guideline(btw i dont agree them :) )
So right now i had to create app/ directory and fork into them.
That's absolutely not critical for me but i think it's may be a good idea to give (who want) this feature. 

What do you think about that?

PS-BTW: Does Oscar page with frequently asked questions in the documentation?

joelhaasnoot:
+ 1 for allowing apps in the root directory - this is the default project layout for Django. Unless you consider oscar apps different from other apps - then you might want to namespace them. My project is mainly overriding various bits and pieces of oscar, so a flat project layout is fine.

codeinthehole:
Fixed by #1718

PR #1718, opened 2015-04-16
fix for django_oscar issue_#1717

manevant:
Checking a folder_path, if its a dot - replace to a blank string.
possible fix for #1717 issue.


maikhoepfel:
Labelled needs more info, please see #1717

PR #1720, opened 2015-04-26
Add another auth backend that uses automatically username or email

Adrihein:


maikhoepfel:
Hi @Adrihein, thank you for this. Can you explain the reasoning behind this PR? What's the goal you're trying to achieve? If you want to allow logging in by email and username, I think you might be able to set both Oscar's ``EmailBackend`` and Django's regular auth backend. 

What do you think?

Adrihein:
Hi,
With EmailOrUsernameModelBackend you can provide a double authentication ways. You don't have to choose between email or username for signup, as currently.
The user will be able to choose if he wants to log with his email or username.

maikhoepfel:
Thanks for getting back to me, and sorry for not being clear. As far as I understand https://docs.djangoproject.com/en/1.8/topics/auth/customizing/#specifying-authentication-backends, it is already possible to specify multiple authentication backends for Django. So you should be able to just specify Django's standard backend, and Oscar's. And that should let people be able to login with username or password, as they wish.

Hence I'm closing this PR. Please let me know if you I misunderstood something.

PR #1724, opened 2015-04-30
Pytest 

mvantellingen:
So this replaces nose with pytest. It also removes the bdd tests which weren't executed. We might want these back in the future via pytest-bdd


mbertheau:
Yay! :)

maikhoepfel:
Lovely work, thanks. I left a few comments that could be addressed, but would be okay with merging it as is. @mvantellingen, feel free to take it from here; I'm sure @codeinthehole is okay with it.

PR #1727, opened 2015-05-02
Hook for changing tax rate during runtime

michalraska:
Added hook to FixedRateTax class for getting tax rate with possibility to change it at runtime without need to restart server or select by product class.

To the hook method ```get_rate``` are passed parameters product and stockrecord. Product is passed because in some countries (e.g. Czech Republic) is possible to have 2 vat rates (basic and reduced) and in my opinion is best place to this information ```ProductClass``` model which is reachable from product instance. StockRecord instance is passed for info about currency. Actually I am not able to imagine case when stock record is needed but if some exists it is ready.

With this I added hook method for exponent too. There is passed ```StockRecord``` instance because exponent is in most cases dependent on currency.

### DIFF
--- a/src/oscar/apps/partner/strategy.py
+++ b/src/oscar/apps/partner/strategy.py
@@ -260,7 +260,9 @@ class FixedRateTax(object):
     def pricing_policy(self, product, stockrecord):
         if not stockrecord:
             return prices.Unavailable()
-        tax = (stockrecord.price_excl_tax * self.rate).quantize(self.exponent)
+        rate = self.get_rate(product, stockrecord)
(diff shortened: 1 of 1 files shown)


maikhoepfel:
This makes good sense, thank you! We'll gladly take it from here.

Note to self: when merging, add short docstring to both methods, explaining they serve as a hook for overriding. Also make sure we don't directly call ``.rate`` and ``.exponent`` anywhere in the code base.
For bonus points, add a test for a varying tax rate.

maikhoepfel:
@michalraska, thank you very much for your work. I cherry-picked it onto master.

Issue #1713, opened 2015-04-13
Product attribute's code generated with dashes


vovanbo:
Let's see in the following code:
```python
>>> class Tmp(object):
...     pass
...     
>>> tmp = Tmp()
>>> setattr(tmp, 'temp-code', 123)
>>> tmp.temp-code
Traceback (most recent call last):
  File "<input>", line 1, in <module>
AttributeError: 'Tmp' object has no attribute 'temp'
>>> tmp.__dict__
{'temp-code': 123}
>>> getattr(tmp, 'temp-code')
123
```
Attribute was set, object have this attribute, but this attribute can not be accessed directly as `tmp.temp-code`. That's right, this is Python.
Product's attributes with spaces in their names have generated code with dashes. So, these attributes can not be accessed with `product.attr.some-attribute`, because of dash in attribute's codename is interpreted as minus sign in program's code. Documentation for `ProductAttributesContainer` offer this way to get attribute's values, but one way to get value is `getattr(product.attr, 'some-attribute')`.
This behavior can be fixed with using AutoSlugField with underscore as separator [here](https://github.com/django-oscar/django-oscar/blob/master/src%2Foscar%2Fapps%2Fcatalogue%2Fabstract_models.py#L780-L785):
```python
class AbstractProductAttribute(models.Model):
    ...
    code = models.SlugField(
        _('Code'), max_length=128,
        validators=[RegexValidator(
            regex=r'^[a-zA-Z\-_][0-9a-zA-Z\-_]*$',
            message=_("Code can only contain the letters a-z, A-Z, digits, "
                      "minus and underscores, and can't start with a digit"))])
    ...
```
Any thoughts?

john-parton:
This isn't a bad idea. I have several places in my code where I have to work around this issue using `getattr` and this approach might save me some headache.

maikhoepfel:
Hi,

good point! But I wouldn't use AutoSlugField because then the attribute
names become unpredictable and often code will rely on them. I'd just add a
validator to the field that ensures we only allow valid characters.

Vladimir or John, do you think you could prep a PR for it? Otherwise,
please create an issue and we'll look at it when time permits.

Cheers,

Maik

maikhoepfel:
Whoops, just saw this is already an issue. Ignore that paragraph. A PR is still welcome :)

john-parton:
Submitted a PR.

[...]

PR #1729, opened 2015-05-05
Forbid hyphens from AbstractProdutAttribute.code

john-parton:
Fixes https://github.com/django-oscar/django-oscar/issues/1713

Validates that `AbstractProdutAttribute.code` is a valid Python identifier.

The only caveat is that if you supply a name with spaces, Django's `prepopulated_fields` setting will generate a code with hyphens, which will then have to be manually fixed.

### DIFF
--- a/src/oscar/apps/catalogue/abstract_models.py
+++ b/src/oscar/apps/catalogue/abstract_models.py
@@ -806,9 +806,9 @@ class AbstractProductAttribute(models.Model):
     code = models.SlugField(
         _('Code'), max_length=128,
         validators=[RegexValidator(
-            regex=r'^[a-zA-Z\-_][0-9a-zA-Z\-_]*$',
+            regex=r'^[a-zA-Z_][0-9a-zA-Z_]*$',
             message=_("Code can only contain the letters a-z, A-Z, digits, "
-                      "minus and underscores, and can't start with a digit"))])
+                      "and underscores, and can't start with a digit"))])
 
     # Attribute types
     TEXT = "text"


mvantellingen:
Thanks. Can you add a unittest for this? If you have no time for this then let us know. We can add it then :-)

john-parton:
I'm a little tight on time right now. What kind of test would you like to see? Just make sure that something like `foo_bar` validates and `foo-bar` raises a `ValidationError`?

john-parton:
We might want to make a note in the release notes as well.

It's a small thing, but if there's already a code in the database that has a hyphen, the user will be prompted to fix it. But if there's some software in the wild that relies on `ProductAttribute.code` not changing (and I'm sure there is), they could be broken by changing the code.

Like I said, small, but probably worth noting.

john-parton:
Don't really have time to submit a proper change:

```python
import keyword

_validate_python_identifier_basic = RegexValidator(
            regex=r'^[a-zA-Z_][0-9a-zA-Z_]*$',
            message=_("Code can only contain the letters a-z, A-Z, digits, "
                      "and underscores, and can't start with a digit"))

def validate_python_identifier(value):
    if keyword.iskeyword(value):
        # patch this up with internationalization and an error code
        raise ValidationError('Cannot be any of the following: %s' % ', '.join(keyword.kwlist)
    _validate_python_identifier_basic(value)

class AbstractProductAttribute(models.Model):
    ...
    code = models.SlugField(
        _('Code'), max_length=128,
        validators=[validate_python_identifier])
```

mvantellingen:
I've just looked how this is done in the namedtuple() code:

```python
        if type(name) != str:                                                      
            raise TypeError('Type names and field names must be strings')          
        if not all(c.isalnum() or c=='_' for c in name):                           
            raise ValueError('Type names and field names can only contain '        
                             'alphanumeric characters and underscores: %r' % name)
        if _iskeyword(name):                                                       
            raise ValueError('Type names and field names cannot be a '             
                             'keyword: %r' % name)                                 
        if name[0].isdigit():                                                      
            raise ValueError('Type names and field names cannot start with '       
                             'a number: %r' % name)  
```

PR #1732, opened 2015-05-10
Run `pip wheel -r` over all requirement files in travis

mvantellingen:
Since both the demo and the migrations test use psycopg2 we might as
well build these as wheels. Should probably speed up travis runs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,12 +35,13 @@ env:
         - DJANGO=Django==1.8.1
 
 before_install:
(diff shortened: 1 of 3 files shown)


mvantellingen:
Travis thinks that the cache is changed between the Django 1.7 / Django 1.8 tests. If we are able to fix that the build might be a bit faster

mvantellingen:
This is not required anymore since pip 7

PR #1733, opened 2015-05-11
Changed name of method from 'queryset' to 'get_queryset' in ProductAdmin which silences Django 1.8 warning

ka7eh:


mvantellingen:
Thanks!!

Issue #1739, opened 2015-05-21
No support for multi-valued ProductAttributeValue 


john-parton:
There's this PR which attempts to resolve this issue: https://github.com/django-oscar/django-oscar/pull/879

I'm not sure it's the "best" way of fixing this, but I think we should have a way of making this work.

Right now, as a workaround, I'm just using a text field and entering in values comma-separated, which works OK, but obviously is somewhat brittle.

I might submit a PR here with a different solution.

Thanks,

John

anentropic:
I have same issue.

Currently there is support for multi-valued option in `apps.dashboard.catalogue.forms`... I started trying to implement it...

I can remove the `unique_together` from `AbstractProductAttributeValue` with an extended class (obviously the model then ought to have further validation, which looks like will be handled by #879)

But I can't add the `MULTI_OPTION` choice to `TYPE_CHOICES` on the `AbstractProductAttribute` model by extending it, because I would have to redefine the existing field.

As a minimal change I would like to see the `MULTI_OPTION` added to `TYPE_CHOICES` asap even if the proper validation etc is not in place, so we are not blocked from hacking up something for ourselves in the meantime.

john-parton:
Thanks for your input.

I think you're 100% right that we should at least add a `MULTI_OPTION` choice to `TYPE_CHOICES` because an application developer simply cannot modify it currently.

However, if it is possible to include a partially- or fully-featured implementation, we should do so.

john-parton:
I updated the changes as discussed in https://github.com/django-oscar/django-oscar/pull/1745#issuecomment-106035074

This seems like more reasonable behavior.

Also amended the commit so that I'm the author (was showing @codeinthehole as the author after I did the rebase) and squashed it into a single commit.

There are still no tests, but I would really appreciate some input before continuing.

solarissmoke:
Fixed in #2309.

PR #1740, opened 2015-05-22
Add a multi_option value type. This will create a new table for the m…

john-parton:
Fixes https://github.com/django-oscar/django-oscar/issues/1739

I took inspiration from https://github.com/django-oscar/django-oscar/pull/879 but tried to implement it in such a way that it would be backwards compatible.

It's not a particularly elegant solution. There's now both a `ForeignKey` and `ManyToManyField` to `catalogue.AttributeOption`, but `ProductAttributeValue` isn't a particularly elegant model to begin with.

I initially thought I would add a `ManyToManyField` and then make the option field actually use it, but doing so would make backwards compatibility really difficult.
## To do

I am willing to do these, but I would appreciate some input before proceeding.
- Update the initial migration.
- Documentation. 
- Tests

Thanks,

John

### DIFF
--- a/src/oscar/apps/catalogue/abstract_models.py
+++ b/src/oscar/apps/catalogue/abstract_models.py
@@ -984,7 +984,7 @@ class AbstractProductAttributeValue(models.Model):
     value_richtext = models.TextField(_('Richtext'), blank=True, null=True)
     value_date = models.DateField(_('Date'), blank=True, null=True)
     value_multi_option = models.ManyToManyField(
-        'catalogue.AttributeOption', blank=True, null=True,
+        'catalogue.AttributeOption', blank=True,
         related_name='multi_valued_attribute_values',
         verbose_name=_("Value multi option"))
     value_option = models.ForeignKey(
--- a/src/oscar/apps/catalogue/abstract_models.py
+++ b/src/oscar/apps/catalogue/abstract_models.py
@@ -797,6 +797,7 @@ class AbstractProductAttribute(models.Model):
     RICHTEXT = "richtext"
     DATE = "date"
     OPTION = "option"
+    MULTI_OPTION = "multi_option"
     ENTITY = "entity"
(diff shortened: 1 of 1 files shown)


john-parton:
As per discussion https://github.com/django-oscar/django-oscar/pull/1745#issuecomment-106025693, it might be better to make `value` return a `QuerySet` for instead of a `Manager` for multi-valued fields. You could still get the manager if you really need it by accessing `ProductAttributeValue.value_multi_option`.

john-parton:
One thing you can do with an option is `attribute.value = 'Option'` and it will automatically look up the option and make sure it's valid: will raise a `DoesNotExist` exception if you try to set it to an invalid option.

With this, you can't do something like `attribute.value = ['Option 1', 'Option 2', 'Option 3']`.

If this functionality is necessary, I could work it in.

anentropic:
After looking at both, I prefer this to #879 

and I do think it's slightly better returning a `QuerySet` :)

re setting the value... I was surprised that I didn't need to override anything in `_set_value` for eg the Product update form to work, but I guess to be able to do `attribute.value = ['Option 1', 'Option 2', 'Option 3']` programmatically then we'd need handling of the m2m field

You haven't removed the `unique_together` from `AbstractProductAttributeValue` in this PR, can it work without doing that?

I assumed we need to, in which case we should ad back in a custom `validate_unique` conditionally for single-value options

john-parton:
Yes, `unique_together` should remain. The m2m relationship is on an entirely different table and subject to its own constraints.

john-parton:
Looks like this adds some lines that are too long (according to flake8). I'll see if I can wrap them.

john-parton:
Is there anything else I need to do with this? 

@maikhoepfel Any thoughts?

solarissmoke:
Finished this off and merged in #2309.

PR #1745, opened 2015-05-27
enable multi_option choice for attributes

anentropic:
See my comment here:
https://github.com/django-oscar/django-oscar/issues/1739#issuecomment-105931475

This is a minimal change allowing us to hack in our own support for multi options while PRs #879 and #1740  are considered

### DIFF
--- a/src/oscar/apps/catalogue/abstract_models.py
+++ b/src/oscar/apps/catalogue/abstract_models.py
@@ -792,6 +792,7 @@ class AbstractProductAttribute(models.Model):
     RICHTEXT = "richtext"
     DATE = "date"
     OPTION = "option"
+    MULTI_OPTION = "multi_option"
     ENTITY = "entity"
     FILE = "file"
     IMAGE = "image"
@@ -803,6 +804,7 @@ class AbstractProductAttribute(models.Model):
         (RICHTEXT, _("Rich Text")),
         (DATE, _("Date")),
         (OPTION, _("Option")),
+        (MULTI_OPTION, _("Multi-option")),
         (ENTITY, _("Entity")),
         (FILE, _("File")),
         (IMAGE, _("Image")),


john-parton:
This is not a bad idea. Figuring out how the multi option choice should actually work could be left to the application developer.

However, I think it would be better if it were implemented fully out-of-the-box.

This PR is okay (seems to work, but might have some compatibility issues): https://github.com/django-oscar/django-oscar/pull/879
This is my PR, which I think is better (I'm biased, I know.): https://github.com/django-oscar/django-oscar/pull/1740

anentropic:
I agree there should be a full solution out of the box

I just intended this as a quick fix that could be merged without harm in the meantime

anentropic:
the minimum model tweaks (once commit above merged) to get it working were less than I expected 

```python
class ProductAttribute(AbstractProductAttribute):
    def _validate_multi_option(self, value):
        for sub_value in value:
            self._validate_option(sub_value)


class ProductAttributeValue(AbstractProductAttributeValue):
    class Meta(AbstractProductAttributeValue.Meta):
        unique_together = ()

    value_multi_option = models.ManyToManyField(
        'catalogue.AttributeOption',
        blank=True, null=True,
        verbose_name=_("Value multi-option"),
        related_name='productattributemultivalue_set'
    )

    def _get_value(self):
        value = super(ProductAttributeValue, self)._get_value()
        if isinstance(value, models.Manager):
            value = value.all()
        return value

    def _set_value(self, new_value):
        super(ProductAttributeValue, self)._set_value(new_value)

    value = property(_get_value, _set_value)

    @property
    def _multi_option_as_text(self):
        return ', '.join(str(val) for val in self.value)
```

obviously single-option attributes are now missing out on their uniqueness validation

but above works (at least to save and view a product with multi-options) and can be implemented in user code

john-parton:
This looks good, but it's getting awfully close to just being my PR. ;-)

I think the only real difference is that you make `value` return a `QuerySet`, whereas I just leave it as a `Manager` and then duck type it later on.

Honestly, I have no strong preference one way or the other. Thinking about it, your way might be slightly better.

anentropic:
ah you're right, I don't even need to remove the `unique_together` :)

sasha0:
Closing in favor of #1740.

PR #1746, opened 2015-05-27
Bugfix: in the dashboard the "Is Active" search no more broken

pdonadeo:
In a query, used for searching active vouchers in the dashboard, the name
of the filtered fields was wrong.


src/oscar/apps/dashboard/vouchers/views.py
```diff
@@ -2,10 +2,11 @@
 
 from django.contrib import messages
 from django.core.urlresolvers import reverse
-from oscar.core.loading import get_model
 from django.http import HttpResponseRedirect
+from django.utils import timezone
 from django.utils.translation import ugettext_lazy as _
 from django.views import generic
+from oscar.core.loading import get_model
```
mvantellingen: Could you group the oscar imports?

PEP8:
```
Imports should be grouped in the following order:

standard library imports
related third party imports
local application/library specific imports
You should put a blank line between each group of imports.
```

pdonadeo: @mvantellingen Of course! I'll group the includes next week, when I'll back from vacation.

pdonadeo:
@mvantellingen I reordered the import blocks to cpmply to PIP8.

Regarding the unit test... is it really necessary? The bug was simply a field name misspelled in a filter.

pdonadeo:
Bump! :smile:

mvantellingen:
@pdonadeo we should make sure we never break it again, so that's why we need a unittest ;-)

I'll try to look at it later, and merge this now. Thanks :-)

pdonadeo:
@mvantellingen thanks for merging the patch. I have to admit that the problem with unit test, in this case, is that I'm not able to write one, so... thanks again.

Issue #1748, opened 2015-05-27
Cannot save child product in admin: product class is required


john-parton:
If you edit a child product in the admin, and try to save, you get a validation error saying that the product_class is required. If you try to save it with a product_class, you get a validation error saying that children can't have a product class.

It essentially makes editing children through the admin impossible.

I'll submit a PR to fix this issue.

PR #1749, opened 2015-05-27
Allow product_class to be blank.

john-parton:
Fixes https://github.com/django-oscar/django-oscar/issues/1748

`Product._clean_standalone` and `Product._clean_parent` already check to make sure `product_class` is not `None`, so there's really no other changes to be made.


mvantellingen:
Thanks! :+1:

Issue #1654, opened 2015-02-03
Cannot remove item from basket if another line is available


john-parton:
Issue with `BasketView` not removing an item

### Steps to replicate

* Add an item to your basket
* Add a different item to your basket
* Set the stock on one of the items to `0`, making it unavailable
* Go to the cart page and attempt to remove the item that is still in stock

### Result

The page bounces back with a message `Your basket couldn't be updated` and the product you attempted to remove is still there.

### Expected behavior

At the very least, the generic message should be replaced with the more specific `'<product>' is no longer available to buy (no stock available). Please adjust your basket to continue`. Ideally, the item you attempted to remove should actually be removed.

### Related

If you try to change the quantity of the in stock item, it bounces back with `Your basket couldn't be updated` and the quantity remains unchanged.

### Notes

It looks like this behavior is because no actions are taken if `BasketLineFormSet` is invalid, even though there could be consistent behavior for a "partially valid" formset. I'm unsure of how `data-behaviors="remove"` works. Would it make more sense for this to just be a plain link? Design decision regarding plain link vs javascript?

codeinthehole:
Good report - I'll have a look into this for 1.1

john-parton:
Thanks. 

I did some digging and figured out that it has to do with the default `FormSet` behavior for Django. Basically, each form that is *not* marked to be deleted is validated. If any one of the forms is invalid, then the entire formset is invalid. If the formset is invalid, then no changes (not even the marked deletions) are made. You can actually test out that behavior through the Django admin.

monikasulik:
Discussed this with David and fixing this completely would have been too hacky. So instead we've opted to improve the message a bit, so the user hopefully knows what to do when similar issues occur.

maikhoepfel:
I just merged #1763, so am closing this issue.

PR #1763, opened 2015-06-04
Improve basket validation message

monikasulik:
Partially fixes (or at least improves the situation with) https://github.com/django-oscar/django-oscar/issues/1654


maikhoepfel:
Lovely, ta!

PR #1766, opened 2015-06-04
Fix validation message in basket

### LINKED ISSUE #1765: In cases where the stock is lower than the number of allocated products incorrect message appears
To reproduce:
1. Add 2 products to basket
2. On one of them set num of allocated products as 1 and stock as 0.
3. On the basket page try to do an operation on the one that is still in stock. The message that appears is "a maximum of -1 can be bought" instead of the "no stock available" message

monikasulik:
Fixes stockrecord availability validation message in the case when stock is lower than allocation. Previously the message that appeared contained negative numbers. For example: "a maximum of -1 can be bought". Now it will display: "no stock available"

Fixes #1765

### DIFF
--- a/src/oscar/apps/partner/availability.py
+++ b/src/oscar/apps/partner/availability.py
@@ -83,7 +83,7 @@ class StockRequired(Base):
         self.num_available = num_available
 
     def is_purchase_permitted(self, quantity):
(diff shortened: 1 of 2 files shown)


maikhoepfel:
Lovely, ta.

PR #1768, opened 2015-06-04
Forbid hyphens in CommunicationEventType code field

### LINKED ISSUE #1653: Dashboard email preview crashes on email templates with hyphen in code
To reproduce, add a `CommunicationEventType` instance through the admin page with a hyphen in its code field (e.g. `TEST-DASH`). Open the dashboard and navigate to Content > Email Templates; an internal server error occurs.

Probably a good fix would be to add validation to the `code` field to ensure that it matches `\w+` as the regular expression in https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/dashboard/communications/app.py#L17 suggests.

codeinthehole:
Fixes #1653

### DIFF
--- a/src/oscar/apps/customer/abstract_models.py
+++ b/src/oscar/apps/customer/abstract_models.py
@@ -5,6 +5,7 @@ import random
 from django.conf import settings
 from django.contrib.auth import models as auth_models
 from django.core.urlresolvers import reverse
+from django.core.validators import RegexValidator
 from django.db import models
 from django.template import Template, Context, TemplateDoesNotExist
 from django.template.loader import get_template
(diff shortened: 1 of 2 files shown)


maikhoepfel:
LGTM

src/oscar/apps/customer/abstract_models.py
```diff
@@ -141,6 +142,12 @@ class AbstractCommunicationEventType(models.Model):
     code = AutoSlugField(
         _('Code'), max_length=128, unique=True, populate_from='name',
         separator=six.u("_"), uppercase=True, editable=True,
+        validators=[
+            RegexValidator(
+                regex=r'^[a-zA-Z_][0-9a-zA-Z_]*$',
+                message=_(
+                    "Code can only contain the letters a-z, A-Z, digits, "
+                    "and underscores, and can't start with a digit."))],
```
mvantellingen: Shouldn't this also have the non_python_keyword validator?

PR #1779, opened 2015-06-13
Update address book on billing address save

### LINKED ISSUE #1532: Saving a billing address doesn't update address book
It probably makes sense to add a billing address to a user's addressbook as part of checkout. We already do this for shipping addresses, but not for billing addresses. Should be a simple change.

sasha0:
Fixes #1532

### DIFF
--- a/docs/source/releases/v1.5.rst
+++ b/docs/source/releases/v1.5.rst
@@ -51,7 +51,9 @@ These methods/modules have been removed:
 
 Minor changes
 ~~~~~~~~~~~~~
-
+- Add billing address to user's address book during checkout (`#1532`_). Number of usages for billing and shipping
+   addresses tracked separately: billing address in `UserAddress.num_orders_as_billing_address` field
+   and shipping address in `UserAddress.num_order_as_shipping_address` accordingly.
 
 
 Dependency changes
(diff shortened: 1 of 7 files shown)


mvantellingen:
Looks good to me, could you also add a note to the changelog for 1.2?

mvantellingen:
After looking more into this (need it myself actually), I think the address.num_orders isn't correct anymore. Since each order now increments two (or one and the same) address.num_order properties

mvantellingen:
continued: perhaps we need to store the number of times an address is used as a billing address and the number of times it is used as a shipping address?

sasha0:
@mvantellingen agreed. I'll make changes.

sasha0:
@mvantellingen ok done. Please take a look.

mvantellingen:
Looks good. Is it possible to a a property `num_orders` with a deprecation warning?

sasha0:
@mvantellingen yep, done.

digicase:
Do you think this pull request can be merged in soon? It looks like the needs more work requirement has been fulfilled by @sasha0 now.

sasha0:
@digicase right, necessary changes were made accordingly to provided feedback, but let's ask @mvantellingen to review this.

sasha0:
Moved release notes to the upcoming release 1.3.

sasha0:
@digicase it's finally merged.

PR #1780, opened 2015-06-13
Remove reviews loading to context and add custom model query instead

### LINKED ISSUE #395: Create template tag for reviews
See https://github.com/tangentlabs/django-oscar/commit/6029026b7585f14bd88bfe33d4b4b296f3f7d14b#commitcomment-2209224

It makes more sense to have reviews loaded via a templatetag rather than in the ``get_context_data`` method.

sasha0:
Fixes #395

### DIFF
--- a/src/oscar/apps/catalogue/views.py
+++ b/src/oscar/apps/catalogue/views.py
@@ -12,7 +12,6 @@ from oscar.apps.catalogue.signals import product_viewed
 from oscar.core.loading import get_class, get_model
 
 Product = get_model('catalogue', 'product')
-ProductReview = get_model('reviews', 'ProductReview')
 Category = get_model('catalogue', 'category')
 ProductAlert = get_model('customer', 'ProductAlert')
 ProductAlertForm = get_class('customer.forms', 'ProductAlertForm')
@@ -65,7 +64,6 @@ class ProductDetailView(DetailView):
 
     def get_context_data(self, **kwargs):
         ctx = super(ProductDetailView, self).get_context_data(**kwargs)
-        ctx['reviews'] = self.get_reviews()
         ctx['alert_form'] = self.get_alert_form()
         ctx['has_active_alert'] = self.get_alert_status()
         return ctx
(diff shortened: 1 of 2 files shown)


mvantellingen:
We could also create a custom ReviewsManager and use `product.reviews.visible.all` in the templates. That way it might also be easier to customise it.

sasha0:
@mvantellingen actually, custom manager already available - `product.reviews.approved.all`.
So that, template filter not needed anymore, just minor template edit.

krzysztofp:
It's still broken in the non-incognito mode, all cached deleted etc. When page is loaded all is fine for ... tens of milliseconds and then the orange div jumps up:

image

krzysztofp:
Just to clarify - I was testing on LIVE, I don't have the test environment locally. So if you think this will resolve the problem, you can merge to master and I can test on LIVE again :)

sasha0:
@krzysztofp could you please specify what exactly is broken? Thanks.

src/oscar/templatetags/reviews_tags.py
```diff
@@ -36,3 +38,7 @@ def may_vote(review, user):
 @register.filter
 def is_review_permitted(product, user):
     return product and product.is_review_permitted(user)
+
+@register.filter
+def get_reviews(product):
```
codeinthehole: A better name would be `approved_reviews`. You could perhaps pass in the number of reviews wanted too instead of piping to the `slice` filter.

sasha0: Or maybe we could use custom model manager query instead of template filter? `{{ product.reviews.approved.all|slice:":3" }}`

auvipy:
This branch has conflicts that must be resolved

sasha0:
@auvipy rebased branch and reworked to simplest solution by just calling custom model manager instead, as proposed above.

mvantellingen:
:+

mvantellingen:
Perhaps i'm doing something wrong, but it doesn't seem to work:


    In [1]: from oscar.apps.catalogue import models
    In [2]: product = models.Product.objects.first()
    In [3]: product.reviews.approved.all()
    ---------------------------------------------------------------------------
    AttributeError                            Traceback (most recent call last)
    <ipython-input-3-8ff3eef58baa> in <module>()
    ----> 1 product.reviews.approved.all()

    AttributeError: 'RelatedManager' object has no attribute 'approved'

sasha0:
@mvantellingen had to rework it in PR #1920. Please see.

PR #1783, opened 2015-06-16
Order and line status verbose names

### LINKED ISSUE #1232: Order status and line status can't be translated
The status pipelines relies on strings to work. I just marked them translatable, but got an import error for that file - I presume marked strings can't be used as dictionary keys. That makes it hard to translate the order status and line status.

sasha0:
Fixes #1232

### DIFF
--- a/src/oscar/apps/order/abstract_models.py
+++ b/src/oscar/apps/order/abstract_models.py
@@ -134,6 +134,13 @@ class AbstractOrder(models.Model):
     set_status.alters_data = True
 
     @property
+    def status_verbose_name(self):
+        status_verbose_names = getattr(settings, 'OSCAR_LINE_STATUS_VERBOSE_NAMES', {})
+        if self.status and status_verbose_names and self.status in status_verbose_names:
+            return status_verbose_names[self.status]
+        return None
+
+    @property
     def is_anonymous(self):
         # It's possible for an order to be placed by a customer who then
         # deletes their profile.  Hence, we need to check that a guest email is
@@ -541,6 +548,13 @@ class AbstractLine(models.Model):
         self.status = new_status
         self.save()
     set_status.alters_data = True
+
+    @property
+    def status_verbose_name(self):
+        status_verbose_names = getattr(settings, 'OSCAR_ORDER_STATUS_VERBOSE_NAMES', {})
+        if self.status and status_verbose_names and self.status in status_verbose_names:
+            return status_verbose_names[self.status]
+        return None
 
     @property
     def category(self):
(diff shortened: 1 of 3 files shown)


codeinthehole:
First of all - thanks for all your excellent hard work.

Forgive my OCD, but I'd prefer you to write commit messages as per this guide: http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html

Eg:

- Keep the first line short and start it with a capital letter
- Use the present tense rather than the past tense

You don't need to include the number of the issue you're fixing in every commit message - better to just put "Fixes #..." in the PR description. It would be useful to have more informative titles and descriptions in the PRs too.  

I'll look at the code now...

src/oscar/apps/order/abstract_models.py
```diff
@@ -134,6 +134,13 @@ def set_status(self, new_status):
     set_status.alters_data = True
 
     @property
+    def status_verbose_name(self):
+        status_verbose_names = getattr(settings, 'OSCAR_LINE_STATUS_VERBOSE_NAMES', {})
+        if self.status and status_verbose_names and self.status in status_verbose_names:
```
codeinthehole: You don't really need the `.. and status_verbose_names ..` do you? The condition wouldn't error if `status_verbose_names` was an empty dict.

src/oscar/apps/order/abstract_models.py
```diff
@@ -134,6 +134,13 @@ def set_status(self, new_status):
     set_status.alters_data = True
 
     @property
+    def status_verbose_name(self):
+        status_verbose_names = getattr(settings, 'OSCAR_LINE_STATUS_VERBOSE_NAMES', {})
```
codeinthehole: This is the wrong setting to use isn't it? Should be `OSCAR_ORDER_STATUS_VERBOSE_NAMES`.

src/oscar/apps/order/abstract_models.py
```diff
@@ -543,6 +550,13 @@ def set_status(self, new_status):
     set_status.alters_data = True
 
     @property
+    def status_verbose_name(self):
+        status_verbose_names = getattr(settings, 'OSCAR_ORDER_STATUS_VERBOSE_NAMES', {})
```
codeinthehole: Wrong setting I think

codeinthehole:
Couple of things to correct - but we should be able to squeeze this into 1.1

src/oscar/apps/order/abstract_models.py
sasha0: Intended to allow different settings for orders and lines. What would you suggest, one settings for both?

src/oscar/apps/order/abstract_models.py
sasha0: @codeinthehole fixed.

sasha0:
@mvantellingen per our latest discussion - do we want to rework status pipeline on top of `django-fsm`?

sasha0:
Per latest discussion with @mvantellingen, we'll experiment with `django-fsm` in order to rework statuses pipeline in the next versions. Thus, closing this PR since none of this changes will be used in the refactoring mentioned above.

PR #1787, opened 2015-06-23
basket: Do None check of unit_price_excl_tax before multiplication

lijoantony:
If price of a product, which is already in a basket becomes unavailable, Line.line_price_excl_tax property throws a TypeError while multiplying quantity with unit_price_excl_tax, since unit_price_excl_tax is None. Do a None check before multiplication to avoid this.

### DIFF
--- a/src/oscar/apps/basket/abstract_models.py
+++ b/src/oscar/apps/basket/abstract_models.py
@@ -759,7 +759,9 @@ class AbstractLine(models.Model):
 
     @property
     def line_price_excl_tax(self):
-        return self.quantity * self.unit_price_excl_tax
+        if self.unit_price_excl_tax:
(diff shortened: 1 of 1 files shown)


mvantellingen:
This is an issue which happended to me also a number of times. I remember discussing this with @maikhoepfel and that we should fix this at the proper location.

@maikhoepfel ? :-)

sasha0:
@lijoantony @mvantellingen we could try to re-create this case with test. What do you think?

sasha0:
Finally added test to re-create this case - #2283.

sasha0:
Instead of falling back to zero price, I think it's better to skip price calculation and not to show price in the basket template, since it's unavailable. Closing in favor of #2283. Thanks.

PR #1790, opened 2015-06-26
Mistranslation resolved

Nekmo:


mvantellingen:
Hi, translation are done via transifex, see https://www.transifex.com/codeinthehole/django-oscar/

Could you update it there? Thanks :-)

PR #1791, opened 2015-06-27
update django-widget-tweaks version

slav0nic:
django-widget-tweaks 1.3 doesn't work correct with django 1.8

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -62,7 +62,7 @@ setup(name='django-oscar',
           'django-tables2>=1.0.4,<1.1',
           # Used for manipulating form field attributes in templates (eg: add
           # a css class)
-          'django-widget-tweaks>=1.3,<1.4',
+          'django-widget-tweaks>=1.4,<1.5',
       ],
       dependency_links=[],
       # See http://pypi.python.org/pypi?%3Aaction=list_classifiers


kunkku:
Please merge #1793 before this if you really want to use version 1.4.

@slav0nic: As you authored the respective commit in django-widget-tweaks, could you please explain why it is essential to deepcopy the fields? I can run Oscar with version 1.3 just fine on Django 1.8. Using deepcopy makes things fail in ways that are unpredictable and hard to debug as demonstrated by the aforementioned pull request.

Moreover, django-widget-tweaks has less than 200 LOC. My fear is that we end up writing more lines of boilerplate code to avoid the problems caused by deepcopy.

slav0nic:
you can run test for 1.3 version with django 1.8 and check issue https://bitbucket.org/kmike/django-widget-tweaks/issue/17/django-18-and-append_attr

kunkku:
I did some testing and it seems shallow copying suffices (https://github.com/kmike/django-widget-tweaks/pull/18).

slav0nic:
1.4.1 version released without `copy`, sorry for this)

mvantellingen:
@slav0nic so we need a minimum of 1.4.1?

slav0nic:
yes

mvantellingen:
Ok, i've committed it. Thanks :-)

PR #1792, opened 2015-06-27
Drop jQuery UI dependencies

### LINKED ISSUE #1462: Drop jQuery UI and jquery-ui-datepicker
#1457 replaced the dashboard's datetimepicker with a much nicer Bootstrap-themed one. While reviewing that, I realised that the customer order list page in the frontend also uses the replaced ``jquery-ui-datepicker``. If I'm not mistaken, that is also the only reason we keep the ``jquery-mousewheel`` and jQuery UI dependencies. 

We should move the order list page to use the new datetime picker, and drop all unneeded dependencies.

sasha0:
Fixes #1462


src/oscar/templates/oscar/dashboard/layout.html
```diff
@@ -19,12 +19,10 @@
 {% block extrastyles %}
     {{ block.super }}
     {% compress css %}
-        {% block jquery-ui-css %}
-            <link rel="stylesheet" href="{% static "oscar/js/jquery-ui/jquery-ui.css" %}">
-        {% endblock %}
         <link rel="stylesheet" href="{% static "oscar/js/select2/select2.css" %}" />
         <link rel="stylesheet" href="{% static "oscar/css/select2-bootstrap.css" %}" />
         <link rel="stylesheet" href="{% static "oscar/js/bootstrap-datetimepicker/bootstrap-datetimepicker.css" %}" />
+        <link rel="stylesheet" type="text/css" href="{% static "oscar/css/datetimepicker.css" %}" />
```
mvantellingen: The three lines above have no type attribute. So we should either add them there too or remove it from the new line

sasha0: @mvantellingen done.

mvantellingen:
Cool, thanks. I'll add a bullet to the 1.2 release notes remarking that JQuery UI is no longer included.

PR #1796, opened 2015-06-30
PartnerAddressForm's name field needs max_length

ka7eh:
PartnerAddressForm's name field is a CharField with no max_length. It causes a DataError if users enter a name with more than 128 characters (max_length of Partner's name).


mvantellingen:
Good find, why did you choose exactly for 128?

mvantellingen:
I didn't read your explanation correctly, merged :-)

PR #1798, opened 2015-07-06
Restrict non-staff user's access to stock records of linked partners only

lijoantony:
Currently a non-staff user has access to all the stock records, if there is at least one stock record from their linked partner for the product. This is wrong in a market place scenario. Access should be restricted to the stock records of the linked partner only. With this fix, non-staff user can create/delete/update stock records for their linked partners only.


mvantellingen:
Good catch. We should probably at a test case for this in the future.

Issue #1803, opened 2015-07-10
currency filter fails if there's no active translation


john-parton:
```python
AttributeError: 'NoneType' object has no attribute 'find'
```

This is because Django 1.7 used to return LANGUAGE_CODE if there's no active translation. Modify line 25 in `templatetags.currency_filters` to restore old behavior:
```python
        'locale': to_locale(get_language() or settings.LANGUAGE_CODE),
```

maikhoepfel:
Sounds about right, John. Thanks for finding this; and feel free to submit a PR and ping me.

PR #1805, opened 2015-07-14
Fall back to Django 1.7 behavior for currency_filter. Fixes an Attrib…

john-parton:
Closes https://github.com/django-oscar/django-oscar/issues/1803

`currency_filter.currency` now falls back to `settings.LANGUAGE_CODE`.

Added a test.

### DIFF
--- a/tests/integration/templatetags/test_currency_filters.py
+++ b/tests/integration/templatetags/test_currency_filters.py
@@ -35,8 +35,8 @@ class TestCurrencyFilter(TestCase):
             'price': ''
         }))
 
-    @translation.override(None, deactivate=True)
     def test_handles_no_translation(self):
-        self.template.render(template.Context({
-            'price': D('10.23'),
-        }))
+        with translation.override(None, deactivate=True):
+            self.template.render(template.Context({
+                'price': D('10.23'),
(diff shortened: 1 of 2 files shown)


john-parton:
Looks like it failed on 1.7 because I was using override as a decorator, but it's only usable as a decorator in 1.8.

I'll push out a change that uses it as a context manager.

tests/integration/templatetags/test_currency_filters.py
```diff
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 from decimal import Decimal as D
 
+from django.utils import translation
 from django.test import TestCase
 from django import template
```
mvantellingen: Minor nitpick, could you alphabetically sort these three imports?

john-parton: That is pretty nit picky... ;-)

mvantellingen: Yeah I realise that, but we have to try to keep the source neat. Letting this slip sometimes will add-up in the long run :(

mvantellingen:
Thanks :-)

PR #1807, opened 2015-07-14
Addresses deprecation warnings in preparation for Django 1.9

aericson:
Addresses RemovedInDjango19Warnings in Django 1.8 and to work with upcoming Django 1.9

I have noticed that `setup.py` says this needs `django>=1.7.8`, therefore I did not bother to make this backward compatible with versions prior to 1.7. Let me know if this needs to change.

Thanks


mvantellingen:
Thanks

Issue #1808, opened 2015-07-15
Squashing basket migrations raises CircularDependencyError


john-parton:
Steps to reproduce:
1. Build the sandbox as described here: http://django-oscar.readthedocs.org/en/latest/internals/sandbox.html#run-the-sandbox-locally
2. Run `sites/sandbox/manage.py squashmigrations basket 0005`
3. Try to start the server back up `sites/sandbox/manage.py runserver`

As far as I can tell, it's because `basket.Basket` has a `ManyToManyField` to `voucher.Voucher`; `voucher.VoucherApplication` has a `ForeignKey` to `voucher.Voucher` and `order.Order`; and `order.Order` has a `ForeignKey` to `basket.Basket`.

I'm not sure how big of an issue this is, but it seems like there is a lot of complicated interdependence between those 3 apps at the model level.

I'll submit a PR with a proposed fix.

``` python
Traceback (most recent call last):
  File "sites/sandbox/manage.py", line 10, in <module>
    execute_from_command_line(sys.argv)
...
django.db.migrations.graph.CircularDependencyError: order.0001_initial, basket.0001_squashed_0005_auto_20150604_1450, voucher.0001_initial
```

solarissmoke:
Closing this based on the discussion in #1809.

PR #1809, opened 2015-07-15
Move Basket.vouchers to Voucher.baskets with migration.

john-parton:
Should close https://github.com/django-oscar/django-oscar/issues/1808

This moves the `vouchers` `ManyToManyField` definition to the `Voucher` model as a `baskets` `ManyToManyField` with a `vouchers` `related_name` for backwards compatibility.

I've included a migration that will remove the old field and add the new field without removing the data. 

I'm not 100% sure how migrations interact with app forking. Probably something that needs to be addressed before this is accepted.

### DIFF
--- a/src/oscar/apps/basket/abstract_models.py
+++ b/src/oscar/apps/basket/abstract_models.py
@@ -42,12 +42,6 @@ class AbstractBasket(models.Model):
     status = models.CharField(
         _("Status"), max_length=128, default=OPEN, choices=STATUS_CHOICES)
 
-    # A basket can have many vouchers attached to it.  However, it is common
-    # for sites to only allow one voucher per basket - this will need to be
-    # enforced in the project's codebase.
-    vouchers = models.ManyToManyField(
-        'voucher.Voucher', verbose_name=_("Vouchers"), blank=True)
-
     date_created = models.DateTimeField(_("Date created"), auto_now_add=True)
     date_merged = models.DateTimeField(_("Date merged"), null=True, blank=True)
     date_submitted = models.DateTimeField(_("Date submitted"), null=True,
--- a/src/oscar/apps/basket/migrations/0006_move_basket_vouchers.py
+++ b/src/oscar/apps/basket/migrations/0006_move_basket_vouchers.py
@@ -1,0 +1,32 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+
+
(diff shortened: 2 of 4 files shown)


john-parton:
Looks like if you have forked both `voucher` and `basket`, then `makemigrations` will drop the table and recreate it with a new name, so all of the Vouchers on Baskets are lost.

Not entirely sure what happens if you've only forked one app. I suspect nastiness.

Looks like this issue is far more complicated than I initially suspected.

mvantellingen:
Hi John,  

I do agree with the fact that vouchers should have an m2m relation with a basket instead of the current option. Can't we just rename the association table in a migration?

john-parton:
@mvantellingen If you look at my PR, it does exactly that: renames the table.

```python
    database_operations = [
        migrations.AlterField(
            model_name='basket',
            name='vouchers',
            field=models.ManyToManyField(to='voucher.Voucher', db_table=b'voucher_voucher_baskets', verbose_name='Vouchers', blank=True),
        ),
    ]
```

As I alluded to, it's slightly more complicated than that, though. We have to let Django know that the "removed" model from one app is the same as the "added" model on the new app, hence all of the business with `migrations.SeparateDatabaseAndState`.

There's some more information on SO: http://stackoverflow.com/questions/25648393/how-to-move-a-model-between-two-django-apps-django-1-7

I've used this approach before, and while it does work quite well, I have since found an important caveat: Django drops and recreates content types for the models. Anything that has a relation to the content type can potentially be broken. The two usual suspects are permissions and generic foreign keys. However, I have an approach that works around this issue relatively neatly.

Sorry if that's a bit of an info dump. Let me know if you have any questions.

john-parton:
It looks like only the Python 3.3 tests are failing. I'm not sure this is specific to my code or not.

We might need to add `from __future__ import unicode_literals` to the top of our modules:

https://docs.djangoproject.com/en/1.8/topics/migrations/#supporting-python-2-and-3

solarissmoke:
I am not sure that this is a good idea. The direction of an m2m is purely semantic - so is it worth risking breaking lots of things by trying to reverse it? I can see this causing all sorts of problems for projects that have forked either of these apps.

As regards the issue in #1808, I think the answer there is simply that you can't squash all the migrations and have to retain a couple of steps to handle the circular dependencies. There is a separate issue there about some of the apps being very tightly coupled.

sasha0:
I agree. We'd think rather about decoupling relations, rather than introduce potentially fragile changes.

Issue #1815, opened 2015-07-25
NoReverseMatch at /dashboard/content-blocks/update/rawhtml/2/


tamchy:
Hello, I have encountered a problem.
I have searched, but found nothing on my problem.

After i created content block and assigned it to URL, for example:
/ 
or 
/catalogue/ 

this content-block cannot be seen on assigned page, furthermore I cannot update my content-block anymore and i have the following error:

Reverse for 'promotion-list-by-url' with arguments '()' and keyword arguments '{u'path': u''}' not found. 1 pattern(s) tried: [u'dashboard/content-blocks/page/(?P<path>/([\\w-]+(/[\\w-]+)*/)?)$']

But, if i assign to blank url or to /catalogue (without last slash), it writes, that url cannot be finded.

Please help.

binarydud:
the issue is that in this commit 55fb961011736d75f7a1c4225e8948e096d7c9da the page_url field was not included. Locally, I overrode this form and added the page_url to the meta.fields. If I get time I will create a pull request.

msadig:
I got the same issue. Even though the issue fixed and merged to the master branch,but it didn't solve the problem which I encountered.

![error](http://i.imgur.com/YYLikW9.png)

kkx:
This patch worked for me, have you ever tried to removed all links created before the fix? The main problem here was that the attribute page_url could not be saved when a new link is added to a promotion page due to the meta.fields configuration. Therefore you have to clean up these links first.

msadig:
@kkx thanks for the reply. 
I'm fairly new to Oscar, that's why I didn't understand well what you mean by saying I gotta "clean up links" :(. But I managed to handle the problem with Django Admin (not the Oscar's dashboard). On the  Promotions › Page Promotions edited and wrote the URL, and it seems to work ok now.

Now problem is the i18n: I append a language prefix to each url (https://docs.djangoproject.com/en/dev/topics/i18n/translation/#language-prefix-in-url-patterns).

Therefore when I try to add a promotion to `/`, it doesn't work, but with `/en/` it works. It seems that I have to manually specify the URL page for each language, but when I try to add content block to other languages, for example `/ru/`, Oscar tell me that the page doesn't exist.

erjaymodi:
go to apps/dashboard/promotions/forms.py

and edit:

class PagePromotionForm(forms.ModelForm):
...
class Meta:
        model = PagePromotion
        fields = ['position']

to instead be:

class Meta:
[...]

PR #1816, opened 2015-07-29
Fix missing page_url promotions form field

binarydud:
fixes https://github.com/django-oscar/django-oscar/issues/1815

The issue is that in this commit: https://github.com/django-oscar/django-oscar/commit/55fb961011736d75f7a1c4225e8948e096d7c9da fields for forms were moved from excludes to the meta.fields. It looks like the page_url field was missed.


bharling:
+1 from us, this fix worked for us.

mvantellingen:
Nice, thanks :-)

PR #1817, opened 2015-07-30
Fix bug when basket.id attribute is not set when unfreezing order ret…

dorianfm:
…urn from anonymous expree checkout via django-oscar-paypal

### DIFF
--- a/src/oscar/apps/offer/applicator.py
+++ b/src/oscar/apps/offer/applicator.py
@@ -91,7 +91,7 @@ class Applicator(object):
         code
         """
         offers = []
-        if not basket.id or not user:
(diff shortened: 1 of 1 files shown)


nfletton:
@dorianfm Looks like the actual problem is due to a change in the signature of the Applicator classes apply() method in Oscar 1.1 and requires a change to django-oscar-paypal rather than here. See django-oscar/django-oscar-paypal#104

mvantellingen:
@dorianfm is this indeed still a problem with the latest django-oscar-paypal release?

sasha0:
Should be already fixed in django-oscar-paypal.

PR #1825, opened 2015-08-19
Use hidden inputs for results forms, instead of hiding visible inputs.

solarissmoke:
Hi,

A small fix to the way hidden parameters are rendered in the search/category results forms. It uses hidden inputs instead of wrapping normal inputs in a hidden div.

There is also currently an ID conflict between the main search form `q` field and the hidden one used in the results forms which interferes with JS manipulations. Rendering the latter manually ensures that only the main search form input has the ID.

### DIFF
--- a/src/oscar/templates/oscar/search/results.html
+++ b/src/oscar/templates/oscar/search/results.html
@@ -42,13 +42,12 @@
 
 {% block content %}
     <form method="get" action="." class="form-horizontal">
-        {# Render other search params in a hidden block #}
-        <div style="display:none">
-            {% for value in selected_facets %}
-                <input name="selected_facets" value="{{ value }}" />
-            {% endfor %}
-            {{ search_form.q }}
-        </div>
+        {# Render other search params as hidden inputs #}
+        {% for value in selected_facets %}
+            <input type="hidden" name="selected_facets" value="{{ value }}" />
+        {% endfor %}
+        <input type="hidden" name="q" value="{{ search_form.q.value }}" />
+
         {% if paginator.count %}
             {% if paginator.num_pages > 1 %}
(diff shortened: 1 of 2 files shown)


src/oscar/templates/oscar/search/results.html
```diff
@@ -42,13 +42,12 @@
...
-            {% for value in selected_facets %}
-                <input name="selected_facets" value="{{ value }}" />
-            {% endfor %}
-            {{ search_form.q }}
-        </div>
+        {# Render other search params as hidden inputs #}
+        {% for value in selected_facets %}
+            <input type="hidden" name="selected_facets" value="{{ value }}" />
+        {% endfor %}
+        <input type="hidden" name="q" value="{{ search_form.q.value }}" />
```
mvantellingen: We could also use widget-tweaks here (since it is now required anyway):

   {% render_field search_form.q type='hidden' %}

solarissmoke: This would still insert an `id` attribute that conflicts with the navbar search form?

mvantellingen: And {{ search_form.q.as_hidden }}  ? :)

solarissmoke: Again, that inserts an ID attribute which is the whole issue this PR is trying to address. The options are:

- Render one of those forms manually.
- turn off auto_id on the search form class.

I am happy to do the latter, but it risks breaking a lot more things for people relying on that ID being there.

sasha0: As per our latest discussion in Slack, we decided on the manual form render, as the optimal solution.

PR #1829, opened 2015-08-24
hide the 'option group' select unless the attribute type selected is 'Option'

v1kku:


maikhoepfel:
Sounds sensible, thanks!

PR #1832, opened 2015-08-25
Fix validate_payment_event for payment events with no lines specified

craigloftus:
Currently  `order.processing.EventHandler.handle_payment_event` defaults `lines` and `line_quantities` to `None`, however, calling it with these defaults will result in an exception being raised in `validate_payment_event` because it expects iterables.

This commit also adds tests to cover `validate_payment_event`.


maikhoepfel:
Excellent analysis and fix. Thanks Craig!

PR #1833, opened 2015-08-26
Clarify location of Haystack Connections

kapari:
The current version didn't specify where the Haystack Connections went; first time through, I assumed we were still in URLs.


mvantellingen:
Not really a hard opinion but it might be better to keep the title intact and add a small note in the paragraph that this can be done in the Django settings. This keeps the current title style intact. What do you think?

sasha0:
Yeah, I agree with @mvantellingen. Hence, closing this PR in favor of #2301 per suggested edit above.

PR #1834, opened 2015-08-27
use range.all_products instead of range.included_products where necessary

v1kku:
This should have been part of #1549


mvantellingen:
Cool, thanks ;-)

PR #1841, opened 2015-09-08
Add --initial-only option in oscar_populate_countries.

jimabramson:
This change adds support for a new command-line option to `oscar_populate_countries` which will suppress an error message / nonzero exit state when the countries have already been populated.   This is useful for idempotent automation tasks which use the same code for first-time installation and successive updates.  Without using the option, the behavior of the command is unchanged.


jimabramson:
@rlucioni @feanil FYI.

maikhoepfel:
That makes total sense. Thanks!

PR #1842, opened 2015-09-10
ProfileUpdateView should allow exclusion of email field

ka7eh:
ProfileUpdateView looks for email field in cleaned_data, while it can be excluded from ProfileForm, as I did, which throws an error.
Anyhow, it's better to make sure email field exists in cleaned_data.


maikhoepfel:
I totally agree. I just cherry-picked your commit to master. Thanks @ka7eh!

PR #1844, opened 2015-09-14
Fix email template extending the wrong base template

solarissmoke:
The plain text template was extending the HTML base instead of the text one.


mvantellingen:
Thanks :-)

PR #1846, opened 2015-09-14
Unavailable product bug fixes

MatthewWilkes:
This fixes a couple of bugs relating to an edge case in unavailable products. If a product uses prices.Unavailable as its price strategy (i.e., the pricing information is not known), open baskets will fail to calculate their price and result in an internal server error.

In addition, custom availability strategies may cause a product to be unavailable even though it has a stock record. This patch removes the duplication of the `StockRequired`'s check for availability from line items when generating warning messages.

### DIFF
--- a/src/oscar/apps/basket/abstract_models.py
+++ b/src/oscar/apps/basket/abstract_models.py
@@ -352,6 +352,12 @@ class AbstractBasket(models.Model):
             except ObjectDoesNotExist:
                 # Handle situation where the product may have been deleted
                 pass
+            except TypeError:
+                # Handle Unavailable products with no known price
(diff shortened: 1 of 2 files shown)


MatthewWilkes:
This PR had gained conflicts due to others being merged. I've rebased it onto master.

maikhoepfel:
I love PRs that come with tests. This all seems sane, thanks @MatthewWilkes!

MatthewWilkes:
@maikhoepfel I wouldn't dream of leaving something like this untested. "Yeah, this bit just checks if we can sell this product, I probably didn't mess it up, let's just ship it and see."

Thanks for the merge, I appreciate it! It'll let us get back onto official releases.

Issue #1850, opened 2015-09-21
Use TestCase.assert* methods instead of Python assert statement


john-parton:
There are around ~20 different places in the test suite where `assert` is used instead of the appropriate `assert*` method.

I'll submit a PR in a few.

PR #1851, opened 2015-09-21
Replace assert statements with appropriate assert* method.

john-parton:
Should close https://github.com/django-oscar/django-oscar/issues/1850

Replace 'assert' statement with 'assert*' methods.

Convert tests.unit.forms.widget_tests to use TestCase class.

Convert bankcard tests to use TestCase.

Convert tests.unit.logging_tests to use TestCase.

### DIFF
--- a/tests/functional/dashboard/catalogue_tests.py
+++ b/tests/functional/dashboard/catalogue_tests.py
@@ -77,7 +77,7 @@ class TestAStaffUser(WebTestCase):
         )
         form = page.forms[0]
         form['productcategory_set-0-category'] = category.id
-        assert form['title'].value != new_title
+        self.assertNotEqual(form['title'].value, new_title)
         form['title'] = new_title
         form.submit()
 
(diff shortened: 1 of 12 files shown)


mvantellingen:
I actually prefer simple assert statements these days since pytest handles that nicely for us. Although for Oscar this might be better. Is there any specific reason to move to this?

john-parton:
Mostly just for consistency. I'm trying to get familiar with the tests and having two different ways of doing things is confusing [to me, at least].

Part of the PR is converting a few tests over to be class-based as well, again, for consistency.

john-parton:
There's also this discussion on stack overflow: http://stackoverflow.com/a/6361186

MatthewWilkes:
Personally, I prefer the assert methods too, for the formatting. For example, asserting lists are equal gets you the diff of the elements rather than a simple failure (assuming you're using a modern copy of unittest, that is).

Also, the syntax is hokey. This is how you specify a custom failure string

    assert foo==True, "Foo is false in barmethod"

and

    assert(foo==True, "Foo is false in barmethod")

asserts that a tuple containing a boolean and a string is nonzero.

mvantellingen:
Ok. note that pytest patches the default assert statement to do that and more. But consistency is a nice plus

mvantellingen:
Thanks for all the hard work btw @john-parton !  :+1:

PR #1852, opened 2015-09-21
Remove all instances of the number 1 being hard-coded in translation strings.

MatthewWilkes:
In English, the singular is always 1, but this is not universally true. In French, for example, the singular is 0 and 1. As 0 is usually special-cased this wasn't biting us, but some more unusual combinations are incorrect. For example, in Russian 101 is singular. Without this patch the Russian for '101 search results' is displayed as '1 search result', for example.

This patch makes the number being pluralised available and used in every translation string where 1 was previously hard-coded. Individual translators can still provide translations that don't use the variables, if they so wish.

Without this patch, translators could still use the variables in the singular strings, so if this is rejected I'd suggest at a minimum updating the Russian translations (and potentially others, depending on their plural rules) to use the variables, to stop incorrect strings showing in the UI. However, I think that the fact these incorrect strings were added by translators in the first place demonstrates that we should use variables wherever possible in English, to avoid this problem in future.

Thoughts?


mvantellingen:
Thanks for the great explanation. Learned something from it :+1:

Issue #1848, opened 2015-09-18
"Remove" button does not remove products from Range


john-parton:
The remove checkboxes and "Remove selected products" button work, but the red remove button doesn't.

This is on `/dashboard/ranges/5/products/`

Looks like it's implemented in JS?

`<a data-behaviours="remove" href="#" class="btn btn-danger">Remove</a>`

john-parton:
Probably related, you get a JS error on that page: `TypeError: $(...).sortable is not a function`

PR #1854, opened 2015-09-22
Use jQuery().prop instead of jQuery().attr for toggling checked for d…

john-parton:
This fixes https://github.com/django-oscar/django-oscar/issues/1848

Basically, the dashboard client js was using the old school `jQuery().attr` method when it should be using `jQuery().prop`


mvantellingen:
Thanks :+1:

Issue #1849, opened 2015-09-21
Inconsistent model instance creation in tests


john-parton:
Looking through the test suite, there are fully 3 different ways of creating `Product` instances. Which way is considered the "most correct" way?
- In `tests.functional.dashboard.catalogue_tests`, `Product` is imported using `get_model` and manipulated directly. Why doesn't it use a factory? If there's a reason for it, there should be a comment in the code.
- In `tests.functional.catalogue.catalogue_tests`, some tests use `oscar.test.factories.create_product`, and other tests use `oscar.test.factories.ProductFactory`.
- In `tests.functional.dashboard.product_tests`, some tests use `ProductFactory` and other tests use `Product` directly, but this time imported using `from oscar.apps.catalogue.models import Product`

Perhaps also confusingly, `oscar.test.factories.create_stockrecord` will implicitly create a product if you don't pass one in.

I think we should pick one way and consistently use it throughout the test suite. _I have already set aside the time to fix this issue, I just need feedback on which way I need to go._

PR #1857, opened 2015-09-25
Replace factories.create_product with factories.ProductFactory

john-parton:
Here's some initial work to close https://github.com/django-oscar/django-oscar/issues/1849

The plan is to replace all instances of `oscar.test.factories.create_product` with `oscar.test.factories.StandaloneProductFactory` or `oscar.test.factories.ParentProductFactory` as appropriate.

Right now, almost every test that involves `Product` ends up instantiating standalone products. Once the tests are cleaned up, I plan on adding some more tests cases to cover parent and child products.


mvantellingen:
Hi John,

Thanks for all the work. I'll look into it better at a later time, but one initial point of feedback: I'd like to keep the `ProductFactory`. You can make `StandAloneProductFactory` a subclass of it. Also ProductFactory(structure='standalone') should be the same.

An other point is that you can set the stockrecord price in one go: `ProductFactory(stockrecords__price_excl_tax=x)`

john-parton:
I left `ProductFactory` in as an alias for `StandaloneProductFactory`. My rationale is that the 3 structures of products are different enough that tests should be explicitly separated for clarity, and that `Product` is actually rather ambiguous. I would like to see the tests be clearer about what kind of product they're testing, be it a standalone product or parent-child pair.

Cool tip with the double underscore. Kind of illustrates what I mean, though. `ProductFactory(structure='parent', stockrecords__price_excl_tax=x)` doesn't make any sense (parents can't have stockrecords).

I'll defer to you on this, of course, but I thought I should elaborate a little.

john-parton:
I also noticed that some tests, for instance `tests/unit/partner/import_tests.py` import a bunch of models directly, instead of using the `get_model` function.

```python
from oscar.apps.partner.importers import CatalogueImporter
from oscar.apps.partner.exceptions import ImportingError
from oscar.apps.catalogue.models import ProductClass, Product
from oscar.apps.partner.models import Partner
```

Should these use `get_model`?

john-parton:
I found a bug after rewriting a test: https://github.com/django-oscar/django-oscar/issues/1860
I submitted a PR to fix the bug here: https://github.com/django-oscar/django-oscar/pull/1861

It's a tiny fix, but because it's not directly related to testing, I thought I should submit an issue/PR for it.

@mvantellingen Should I wait until the change is pulled into `master` or should I pull that into this branch as well?

Thanks.

john-parton:
@mvantellingen I merged in master. Someone wrote some new tests using the old functional factories, so I had to update them to use the new classes. Really pretty easy, honestly.

john-parton:
@mvantellingen Merged in master again. `TestScales.test_decimals` was broken. 

Basically, the weight attribute is a `float` (and there is no decimal type), so when the scale weighs 3 items that weight `0.3`, it ends up with `3 * Decimal(0.3)` which not the same as `Decimal('0.9')` due to float approximation. One solution is to quantize the decimal result with the `Decimal.quantize` method to round the weight to the nearest tenths (as I have done here). Another solution would be to use `self.assertAlmostEqual` instead, but I thought that was less clear.

john-parton:
@mvantellingen Is there anything else I need to do with this one?

john-parton:
@mvantellingen Is there any desire to get this merged?

Test writers have continued to use the `create_product` function instead of the class-based factories since Feb., which makes keeping this PR up-to-date very difficult. At this point, it might be easier to redo all of my changes against a fresh copy of master.

Thanks.

solarissmoke:
Closing as there are too many conflicts. Need to start again or just make this change piecemeal as/when the relevant tests are updated.

PR #1862, opened 2015-09-29
Remove unexisting JS method call from dashboard order list template

### LINKED ISSUE #1813: oscar.dashboard.orders.initFilterToggle is not a function
Hi,

when accessing the orders list in the dashboard the browser console is showing this error:
Uncaught TypeError: oscar.dashboard.orders.initFilterToggle is not a function 

We have experienced the error in our project, but also the oscar sandbox is affected. From the code seems that the orders_list.html template initialize this:

    {% block onbodyload %}
        {{ block.super }}
        oscar.dashboard.orders.initTable();
        oscar.dashboard.orders.initFilterToggle();
        oscar.dashboard.search.init();
        {% if form.errors %}
        $('#SearchModal').modal('show');
        {% endif %}
    {% endblock onbodyload %}

But there is actually no function initFilterToggle() in dashboard.js

sasha0:
Fixes #1813

### DIFF
--- a/src/oscar/templates/oscar/dashboard/orders/order_list.html
+++ b/src/oscar/templates/oscar/dashboard/orders/order_list.html
@@ -196,7 +196,6 @@
     {% block onbodyload %}
         {{ block.super }}
         oscar.dashboard.orders.initTable();
-        oscar.dashboard.orders.initFilterToggle();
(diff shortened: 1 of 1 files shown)


maikhoepfel:
Thanks for looking into this, Sasha!

PR #1864, opened 2015-10-02
Fix voucher expiration message

### LINKED ISSUE #1863: Vouchers are rejected as 'expired' if added before their validity
If a voucher is added to the basket _before_ its `start_datetime`, the error message says `The voucher has expired`, which is not the case.

gabalese:
Fixes #1863


maikhoepfel:
Yeah, good change. Thanks!

PR #1866, opened 2015-10-03
Fix indirect update of session data, fixes #1865

### LINKED ISSUE #1865: ShippingAddressView.get_initial <Country: XXX> is not JSON serializable
If you are to return back to the shipping address after you've reached further into the checkout process and you selected a country from the list, doing a GET request to the address /checkout/shipping-address/ would result in the following error:

`TypeError at /shop/checkout/shipping-address/ <Country: XXXXX> is not JSON serializable`

The issue is in here https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/checkout/views.py#L140

Modifying the initial dictionary in this function modifies the same reference that is going to be stored in the session.

Bogh:
This changes the issue only for ShippingAddressView.get_initial case.

If this fix isn't sufficient I could make another update and make the change in the following method
`https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/checkout/utils.py#L24`
thus fixing the issue for all situations.

I could also write a test for this.

### DIFF
--- a/src/oscar/apps/checkout/views.py
+++ b/src/oscar/apps/checkout/views.py
@@ -140,6 +140,7 @@ class ShippingAddressView(CheckoutSessionMixin, generic.FormView):
     def get_initial(self):
         initial = self.checkout_session.new_shipping_address_fields()
         if initial:
+            initial = initial.copy()
             # Convert the primary key stored in the session into a Country
             # instance
             try:


maikhoepfel:
Good fix, thanks. So I'll merge. But I'd really like a fix in the `utils.py` and tests are always welcome :) Thanks Bogdan!

PR #1867, opened 2015-10-13
Allow for children to have different images from their siblings

john-parton:
This is my attempt to reduce the size of https://github.com/django-oscar/django-oscar/pull/1643

This should close https://github.com/django-oscar/django-oscar/issues/1611

It really only does a very small number of things:
- Adds a `child_images` field
- Exposes it in the dashboard
- When a user adds a child to the basket, it uses the `child_images` field to determine which image show.

Things this _does not_ do:
- Change any template
- Show the variant image on the product page (before the user adds it to their cart)
- Make associating images with children particularly easy. (The widget in the dashboard is usable for products with small numbers of images and variants, but quickly becomes cumbersome when you have dozens of them.)
- Test my changes

@mvantellingen or another maintainer, can you please let me know if you think this approach is viable. and I'll add some tests.


mvantellingen:
Hi John, will try to look at it ASAP

john-parton:
@mvantellingen Merged in master. There was a conflict with migrations. I wasn't sure what the best way to resolve that would be, but I made an attempt. Let me know if you need anything changed.

SalahAdDin:
:+1:

sasha0:
Closing within the PR #1643 for the same reason - https://github.com/django-oscar/django-oscar/pull/1643#issuecomment-364450052

Issue #1868, opened 2015-10-13
Basket Line Reference with already encoded utf-8 options


jlmadurga:
In an ecommerce model with Options a Basket Line Reference is created by its options representation encoded as utf-8. It raises a problem when option value is already encoded in utf-8. 

```python
    def _create_line_reference(self, product, stockrecord, options):
        """
        Returns a reference string for a line based on the item
        and its options.
        """
        base = '%s_%s' % (product.id, stockrecord.id)
        if not options:
            return base
        return "%s_%s" % (base, zlib.crc32(repr(options).encode('utf8')))
```

I imagine this encoding is done because only basket.AttributeLine value supported is charField. In extended models with more rich Attribute values it is a problem.

jlmadurga:
Sorry forgot python3 bytes enconding

sasha0:
I suppose the issue can be closed, since related PR was merged.

sasha0:
Closed since related PR already merged.

PR #1869, opened 2015-10-13
Fix create_line_reference for options already enconded to utf-8

jlmadurga:
Issue #1868 

Just remove the utf-8 encoding done in create_line_reference when adding a Product to Basket with Options already encoded to utf-8.

Unit test provided.


jlmadurga:
Sorry! Forgot python3 bytes encoding

mvantellingen:
Thanks!

Issue #1615, opened 2015-01-05
BasketMiddleware.merge_baskets and checkout process


john-parton:
Under some circumstances, baskets are merged together, where it would actually make sense for a more recent basket to replace older baskets.

Here's an example for when it is (almost certainly) inappropriate to merge two baskets together:
- A logged in user adds an item to their basket
- The user logs out (or uses a different computer)
- User adds a different item to their basket
- Basket now shows only the 2nd item
- User proceeds to checkout (Obviously if a user presses the checkout button at this point, they intend to checkout with the basket in its current state.)
- User is prompted to log in
- User logs in and is redirected to first step of checkout process
- BasketMiddleware has merged the two baskets together, so now there is a basket with both items in it.
- User fills out shipping information, billing information, etc.
- Only then is the user presented with this new basket that has (surprise) both items in it

Instead of silently merging the two carts, maybe the user should be notified that there are two carts, and be prompted to pick one or somehow at least draw attention to the fact that they have been merged. If the first cart is quite old, it might be safe to discard the first cart silently.

Some discussion elsewhere:
http://stackoverflow.com/questions/11942764/is-it-appropriate-to-merge-guest-user-cart-items-with-items-in-the-cart-when-a-u
http://ux.stackexchange.com/questions/52785/should-shopping-carts-from-different-sessions-be-merged-or-not

mvantellingen:
I've had the same issue on a project I was working on. We chose to always use cookie-based baskets, which also solved the same problem.

See also #1317

john-parton:
@mvantellingen Can you provide me with some more information regarding the cookie-based baskets? I'm having a hard time finding info about it in the docs.

Thanks.

sasha0:
Fixed by PR #1871.

PR #1871, opened 2015-10-14
Add a message if baskets are merged

john-parton:
Fix for https://github.com/django-oscar/django-oscar/issues/1615

Adds a simple message if baskets have to be merged.

It might be better in the future to generate a more accurate/verbose message letting the user know exactly how the cart changed.

### DIFF
--- a/src/oscar/apps/basket/middleware.py
+++ b/src/oscar/apps/basket/middleware.py
@@ -1,6 +1,8 @@
 from django.conf import settings
+from django.contrib import messages
 from django.core.signing import Signer, BadSignature
 from django.utils.functional import SimpleLazyObject, empty
+from django.utils.translation import ugettext_lazy as _
 
 from oscar.core.loading import get_model
 from oscar.core.loading import get_class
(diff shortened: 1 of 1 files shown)


maikhoepfel:
Hey John, thank you for this. I agree with the idea! But if I'm not mistaken, basket merging will never remove products and is safe, so I think a more positive message is appropriate. Something like `We have merged your baskets; the contents might have changed.` What do you think?

Also, I'd like to see the message assigned once to a variable instead of repeated. That reduces the change of one message being updated while the other isn't. Sounds sensible?

john-parton:
You are correct that merging baskets will never remove items, but I would not say that adding items is "safe." If you add an item to a basket without the user expecting it, it's possible that they will accidentally purchase the item. This isn't usually an issue if you add two different items, because it's usually pretty "obvious" when a second line is added to the cart. However, the merging algorithm _adds_ line quantities together for identical items, so two items could become three.

The message I wrote is a little verbose, for sure, but I was trying to make it as clear as possible to a less savvy user. "We have merged your baskets" prompts me to ask "What baskets? I only had one basket" which is why I added the verbiage "from a previous session".

How is this for a message:

"We have merged a basket from a previous session. Its contents may have changed."

I see what you mean about the possibility of a duplicate message. I'll see if I can get that changed.

maikhoepfel:
Lovely, that works for me. Thanks!

Issue #1875, opened 2015-10-19
Dashboard has a horizontal scrollbar


john-parton:
Looks like an element in the navigation is being positioned slightly too far to the right. (Negative margin is too high.)

I'll submit a fix.

mvantellingen:
Thanks!

PR #1876, opened 2015-10-19
Fix horizontal scrolling in dashboard

john-parton:
Fixes https://github.com/django-oscar/django-oscar/issues/1875

For some reason, my lessc rounded the percentages to a different precision. Maybe we should document which version of lessc to actually use to prevent these sorts of annoying cases?


john-parton:
@maikhoepfel Could you take a quick look at this, please? I patched around the issue with the diff rounding percentages incorrectly.

mvantellingen:
Thanks!

Issue #1878, opened 2015-10-19
Inline styles used in templates/oscar/dashboard/orders/order_detail.html


john-parton:
`style="text-align: right"` is used instead of `class="text-right"`

PR #1879, opened 2015-10-19
Use class=text-right instead of inline style.

john-parton:
Fixes https://github.com/django-oscar/django-oscar/issues/1878

Uses the `text-right` Bootstrap class.


mvantellingen:
Thanks

Issue #1881, opened 2015-10-20
Email to anonymous users cannot be saved


john-parton:
Basically `AbstractEmail.user` has `null=False` so there's no way to just simply save an email that's not hooked to a `User`.

Somewhat related, the actual email address that the email was sent to is not recorded. This could differ from a user's email because they could have changed it between when the email was generated and the present.

I'm not sure we should change the default behavior to save anonymous emails, but we should definitely allow for it to be overridden.

PR #1882, opened 2015-10-20
Issue/1881/save anonymous email

john-parton:
Fix for https://github.com/django-oscar/django-oscar/issues/1881

The email field is set even if user is not null, so if the user changes their address, we still have a reasonable record of the email.

This does _not_ automatically save anonymous email. Anyone who wants to enable that feature likely has to override `apps.customer.utils.Dispatcher` or modify some view code. This is intentional, as I'm trying to keep this PR as small as possible.


sasha0:
Closed in favor of PR #2299, which is updated version of this PR and includes test.

Issue #1883, opened 2015-10-20
customer.utils.Dispatcher needs return values


john-parton:
Right now `customer.utils.Dispatcher` has a bunch of methods that return `None` where it really makes sense to have a return value. It would make hooking into or overriding `Dispatcher` a lot easier.

Might make sense to have some tests for `Dispatcher` as well.

PR #1884, opened 2015-10-20
Issue/1883/dispatcher return values

john-parton:
Fix for https://github.com/django-oscar/django-oscar/issues/1883

The idea is to make `Dispatcher` more modular so that more elements can be customized. I'll be able to make some more useful additions after https://github.com/django-oscar/django-oscar/pull/1882 is merged.

I should probably write some tests. The `Dispatcher` class only has a few integration tests. Docs might be nice.

### DIFF
--- a/src/oscar/apps/customer/utils.py
+++ b/src/oscar/apps/customer/utils.py
@@ -23,60 +23,71 @@ class Dispatcher(object):
 
     def dispatch_direct_messages(self, recipient, messages):
         """
-        Dispatch one-off messages to explicitly specified recipient(s).
+        Dispatch one-off messages to explicitly specified recipient.
         """
         if messages['subject'] and messages['body']:
-            self.send_email_messages(recipient, messages)
+            return self.send_email_messages(recipient, messages)
 
-    def dispatch_order_messages(self, order, messages, event_type=None,
-                                **kwargs):
+    def dispatch_order_messages(self, order, messages, event_type=None, **kwargs):
         """
-        Dispatch order-related messages to the customer
+        Dispatch order-related messages to the customer.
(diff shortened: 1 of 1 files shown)


src/oscar/apps/customer/utils.py
```diff
@@ -23,64 +23,75 @@ def __init__(self, logger=None):
...
+        Dispatch order-related messages to the customer.
         """
         if order.is_anonymous:
-            if 'email_address' in kwargs:
-                self.send_email_messages(kwargs['email_address'], messages)
-            elif order.guest_email:
-                self.send_email_messages(order.guest_email, messages)
-            else:
-                return
+            email = kwargs.get('email_address', order.guest_email)
```
mvantellingen: To keep the same behaviour shouldn't we add `if not email: return`?

mvantellingen:
Looks good @john-parton! One minor question added :-)

src/oscar/apps/customer/utils.py
john-parton: I'm not sure. I tried to keep the logic exactly the same.

There's a check in `send_user_email_messages`

sasha0: I think early return is not necessary, since both methods have it.

john-parton: I'm not totally sure what the question is. The behavior should be the same.

Maybe we need a test for the correct behavior?

sasha0: Test will be useful in any case.

sasha0:
Closing in favor of PR #2303, which is rebased version of this PR.

Issue #1885, opened 2015-10-23
Email templates incorrectly escape product names


john-parton:
If you have a product that has a quote `"` character, it's rendered as `&quot;` in the emails.

The simplest solution is to use a `|safe`. Wrapping the entire body or subject in `{% autoescape off %}` is another option, but that could have other effects.

aleprovencio:
Hello @john-parton, unfortunately for my product titles to show up correctly (not empty) on the alert emails I have to take the `|safe` out from it. What can be wrong?

PR #1886, opened 2015-10-23
Update templates so that the product title isn't automatically escaped.

john-parton:
Here's an easy one.

Fixes https://github.com/django-oscar/django-oscar/issues/1885


mvantellingen:
Thanks again :)

craigloftus:
@john-parton @mvantellingen Just a heads up that this PR did not address the issue as you cannot use template filters inside {% blocktrans %}. I have pushed another attempt, #1943.

john-parton:
@craigloftus Thanks for the feedback. I'll try to keep that issue in mind for future pull requests.

PR #1887, opened 2015-10-23
Fix dashboard overriding alert style

john-parton:
This is basically a resubmission of https://github.com/django-oscar/django-oscar/pull/1252

The buttons are now correctly styled.


mvantellingen:
Hi John, is it possible to fix the commit to ignore the decimal rounding errors?

mvantellingen:
@john-parton can you try to generate the css via #1891 ?

john-parton:
@mvantellingen Try that. It's still not perfect, but I used version 1.7.5 of lessc instead of the one I happened to have installed on my system.

mvantellingen:
Looks good, thanks

john-parton:
@mvantellingen There's a regression with this, actually. Error alerts apply a `alert-error` css class, but the bootstrap class is called `alert-danger`.

Maybe the simplest fix is to update the settings:

```python
# settings.py
from django.contrib.messages import constants as messages
MESSAGE_TAGS = {
    messages.ERROR: 'error danger',
}
```

That way `messages.error` will set both tags.

Alternatively, we can update `templates/partials/alert_messages.html` and `templates/dashboard/partials/alert_messages.html` to output the `alert-danger` class if the `error` tag is set.

Let me know if you have a preferred way you want it fixed and I'll submit a new PR. I can open a separate issue if you would like.

Thanks.

PR #1889, opened 2015-10-25
Fixed a bug in Dashboard Catalogue filter

DNX:
Fixed a bug in Dashboard Catalogue when filter by upc and have multiple results.
Where is the right place to write some tests for this?

### DIFF
--- a/src/oscar/apps/dashboard/catalogue/views.py
+++ b/src/oscar/apps/dashboard/catalogue/views.py
@@ -128,15 +128,15 @@ class ProductListView(SingleTableMixin, generic.TemplateView):
             # that contain the UPC
             matches_upc = Product.objects.filter(upc=data['upc'])
             qs_match = queryset.filter(
-                Q(id=matches_upc.values('id')) |
-                Q(id=matches_upc.values('parent_id')))
+                Q(id__in=matches_upc.values('id')) |
+                Q(id__in=matches_upc.values('parent_id')))
 
             if qs_match.exists():
                 queryset = qs_match
             else:
                 matches_upc = Product.objects.filter(upc__icontains=data['upc'])
                 queryset = queryset.filter(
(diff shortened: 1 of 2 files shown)


DNX:
With Django 1.8.5 and Postgres 9.4.5 the sql query in this case will be:
```
SELECT "catalogue_product"."id", "catalogue_product"."structure", "catalogue_product"."upc", "catalogue_product"."parent_id", "catalogue_product"."title", "catalogue_product"."slug", "catalogue_product"."description", "catalogue_product"."product_class_id", "catalogue_product"."rating", "catalogue_product"."date_created", "catalogue_product"."date_updated", "catalogue_product"."is_discountable", "catalogue_product"."points", "catalogue_product"."points_calculator", "catalogue_productclass"."id", "catalogue_productclass"."name", "catalogue_productclass"."slug", "catalogue_productclass"."requires_shipping", "catalogue_productclass"."track_stock" FROM "catalogue_product" LEFT OUTER JOIN "catalogue_productclass" ON ( "catalogue_product"."product_class_id" = "catalogue_productclass"."id" ) WHERE ("catalogue_product"."parent_id" IS NULL AND ("catalogue_product"."id" = (SELECT U0."id" FROM "catalogue_product" U0 WHERE UPPER(U0."upc"::text) LIKE UPPER('%123%')) OR "catalogue_product"."id" = (SELECT U0."parent_id" FROM "catalogue_product" U0 WHERE UPPER(U0."upc"::text) LIKE UPPER('%123%')))) ORDER BY "catalogue_product"."date_created" DESC
```

DNX:
@mvantellingen where is the right place to write a test in this case? I didn't find any that test the Dashboard Catalogue Views.

mvantellingen:
The dashboard tests can be found here: https://github.com/django-oscar/django-oscar/tree/master/tests/functional/dashboard

DNX:
I wrote some tests that will fail without this PR fix

mvantellingen:
Thanks again!

PR #1890, opened 2015-10-26
Remove django-compressor

maikhoepfel:
django-compressor served it's purpose, but given the advent of
more and more frontend JS work, and the accompanying build tools (grunt,
gulp, etc.), it makes less and less sense to use compressor. Jannis
Leidel himself also stepped back from maintaining it.

In my experience with Oscar projects, it has also become a bit of a
liability to have compressor as a dependency. Most projects don't use it
any more, but it's not easy to disable/override because all our
templates include it.

So, let's keep things simple and remove it as a dependency. That paves
the way for using other tools, while it's always easy to add back.

### DIFF
--- a/docs/source/howto/how_to_handle_statics.rst
+++ b/docs/source/howto/how_to_handle_statics.rst
@@ -39,31 +39,10 @@ LESS/CSS
 
 By default, CSS files compiled from their LESS sources are used rather than the
 LESS ones.  To use Less directly, set ``USE_LESS = True`` in your settings file.
-You will also need to ensure that the ``lessc`` executable is installed and is
-configured using a setting like::
-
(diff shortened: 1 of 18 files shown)


maikhoepfel:
@mvantellingen, what do you think?

mvantellingen:
Yeah seems like a good first step to me! We should move in the future move to gulp like for example wagtail does https://github.com/torchbox/wagtail/tree/master/gulpfile.js

maikhoepfel:
Yeah, I fully agree. Is that a LGTM? :)

mvantellingen:
yeah :+1:

maikhoepfel:
Thanks a lot for the review!

PR #1891, opened 2015-10-26
Implement simple gulp task to generate less files

mvantellingen:
Usage:
  npm install
  npm run compile
  npm run start

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -93,12 +93,8 @@ compiledmessages:
 	cd src/oscar; django-admin.py compilemessages
 
 css:
-	# Compile CSS files from LESS
-	lessc --source-map --source-map-less-inline src/oscar/static/oscar/less/styles.less src/oscar/static/oscar/css/styles.css
-	lessc --source-map --source-map-less-inline src/oscar/static/oscar/less/dashboard.less src/oscar/static/oscar/css/dashboard.css
-	# Compile CSS for demo site
-	lessc --source-map --source-map-less-inline sites/demo/static/demo/less/styles.less sites/demo/static/demo/css/styles.css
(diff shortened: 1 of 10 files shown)


mvantellingen:
@maikhoepfel perhaps this is a good starting point.

Makefile
```diff
@@ -93,12 +93,8 @@ compiledmessages:
...
 
 css:
-	# Compile CSS files from LESS
-	lessc --source-map --source-map-less-inline src/oscar/static/oscar/less/styles.less src/oscar/static/oscar/css/styles.css
-	lessc --source-map --source-map-less-inline src/oscar/static/oscar/less/dashboard.less src/oscar/static/oscar/css/dashboard.css
-	# Compile CSS for demo site
-	lessc --source-map --source-map-less-inline sites/demo/static/demo/less/styles.less sites/demo/static/demo/css/styles.css
-	lessc --source-map --source-map-less-inline sites/demo/static/demo/less/responsive.less sites/demo/static/demo/css/responsive.css
+	npm install
+	npm run compile
```
maikhoepfel: I know we used to call it `compile`, mostly out of ignorance. But I think a name like `preprocess-css` is much more appropriate.

maikhoepfel:
Cool! And now live-reloading, linting of the JS, Babel for ES6... one can dream :+1: and this is a good foundation for it.

Makefile
mvantellingen: I'll change it

mvantellingen:
Need to fix a couple of small things noted by a colleague and then i'll merge it

Makefile
mvantellingen: Ended up naming it 'build'. It's commonly used, and the task will possibly do non css related tasks in the future (javascript etc)

PR #1894, opened 2015-10-27
Correct allow_failures for Travis

maikhoepfel:
Django 1.8 supports Python 3.5. It does raise a DeprecationWarning, which is
safe to ignore. We only care about DeprecationWarnings in our own code.

And it seems that Travis wants all allowed failures listed explicitly, so we
do that. It should now pass again.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,6 +53,8 @@ matrix:
         - python: 3.4
           env: "DJANGO=Django==1.9b1"
         - python: 3.5
+          env: "DJANGO=Django==1.8.5"
+        - python: 3.5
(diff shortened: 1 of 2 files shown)


maikhoepfel:
Err, I'm not sure why Travis isn't picking this up.

mvantellingen:
Travis ignored the allow_failures line due to case issues in the Django 1.9 module name (my bad). I've fixed this and also re-added Python 3.5 / Django 1.8 to the allow_failures since travis still failed with a DeprecationError. It isn't officially supported yet but afaik this will happen in Django 1.8.6

PR #1902, opened 2015-11-04
Some review form fields do not always exist

solarissmoke:
For logged in users, the `name` and `email` fields do not exist in the form. Check whether they exist before attempting to render them.


maikhoepfel:
That's a good change, thank you very much!

PR #1905, opened 2015-11-08
Added options attribute addition documentation

vkaracic:
I was tasked to create a product that would be out-of-the-box available to our users, so I added the attributes in a data migration but I had a hard time finding any resource that describes adding options via code and had to dig through Oscar's code on GitHub to figure this out. People would benefit from an addition like this, at least save time from digging through code to figure this out.

### DIFF
--- a/docs/source/topics/modelling_your_catalogue.rst
+++ b/docs/source/topics/modelling_your_catalogue.rst
@@ -43,6 +43,34 @@ they have a ``size`` attribute::
     > shirt.attr.size
     <AttributeOption: Large>
 
+You can assign ``option`` s to your product. For example you want a Language attribute
+to your product, and a couple of options to choose from, for example English and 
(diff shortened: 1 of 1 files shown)


docs/source/topics/modelling_your_catalogue.rst
```diff
@@ -43,6 +43,34 @@ they have a ``size`` attribute::
...
+    > )
+    > AttributeOption.objects.create(
+    >     group=language,
+    >     option='Croatian'
+    > )
+
+Finally assign the Language options group to your product as an attribute::
+
+    > ProductAttribute.objects.create(
+    >     product_class=klass,
```
jstarcher: klass variable hasn't been defined

maikhoepfel:
Lovely addition, thank you very much.

vkaracic:
@maikhoepfel I don't see this addition in the Oscar documentation. Is something wrong with it?

PR #1907, opened 2015-11-11
Use correct field names in VoucherForm

maikhoepfel:
Fixes a bug in VoucherForm's clean method. The logic to validate sensible values for the start and end time of the voucher would always be skipped because it was forgiving and the wrong field names were used.


maikhoepfel:
Travis passed; merging because I'm a cowboy like that.

PR #1908, opened 2015-11-13
Introduce new Voucher.created datetime, deprecate Voucher.date_created

maikhoepfel:
I think almost always datetime fields are more useful than date fields. We
wanted to be able to sort by last created vouchers, which is not useful when
only the day of the year is available.

I'm being a bit handwavy with the migration procedure, because I think it's
overall quite simple to implement, and my guess is that most deployments don't
really care about creation dates for older vouchers. Hence just doing the
switch when Oscar 1.3 comes around should be sufficient.

### DIFF
--- a/docs/source/releases/v1.2.rst
+++ b/docs/source/releases/v1.2.rst
@@ -47,7 +47,14 @@ What's new in Oscar 1.2?
 Minor changes
 ~~~~~~~~~~~~~
  - Fix missing page_url field in the promotions form (`#1816`_)
-
+ - ``Voucher.date_created`` and ``VoucherApplication.date_created``
+   are date fields that have been deprecated.
+   Their intended replacements are the new ``.created`` datetime
+   fields, which we hope are more generally useful. If you have been happy
+   with just having the date, we suggest to keep using the old fields
+   for now, and only switch to the new field when ``date_created`` is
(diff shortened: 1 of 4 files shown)


maikhoepfel:
Review welcome!

mvantellingen:
@maikhoepfel I agree, ran into the same issue a few times. I do think that we need a data migration to copy the `date_created` values to the `created` column (with time 00:00)

maikhoepfel:
Okay, I should find the time to get this added.

maerteijn:
In Oscar, only in the Wishlist and Voucher models the `date_created` fields are DateFields. In all other models this field is already a DateTimeField. I would suggest to migrate the existing 'date_created` fields (which are DateFields) to DateTimeFields so the naming (and field types) are consistent throughout Oscar.

mvantellingen:
I agree with @maerteijn, any objections to do it this way @maikhoepfel ?

maikhoepfel:
Fully agree with @maerteijn! Don't think I'll have time to do any work before February though.

sasha0:
Perhaps we could check other fields naming for consistency.

SalahAdDin:
:+1: Up!

sasha0:
Closing in favor of #2305, which changes fields types and includes necessary migrations.

PR #1909, opened 2015-11-13
Update dependencies on pillow, factory-boy and django-widget-tweaks

mvantellingen:
Instead of forcing a version range for theses packages specificy only
the minimal version number. This packages are stable enough and forcing
upper version limits can cause issues with other (non oscar)
dependencies.


mvantellingen:
Personally I think the oscar releases are too infrequent. Documenting package numbers which are tested by us is imho a better method. I keep running into issues with other dependencies which require a newer version of for example factory boy.

PR #1911, opened 2015-11-16
Remove US and demo sites

mvantellingen:
Remove the US and Demo sites as they are not up-to-date anymore. The demo site might be moved to a separate repository in the future.

Discussed with both David and Maik.


codeinthehole:
Thanks for this.

PR #1913, opened 2015-11-22
Make product unavailable if price is empty to avoid exception when adding product without price to basket

### LINKED ISSUE #1912: Exceptions when price_excl_tax is unset
The Oscar product admin allows the `price_excl_tax` field to be left unset, however an exception occurs when adding a product to the basket if this field is empty.

Similarly, if I have a product in my cart, and then clear the `price_excl_tax` field, an exception occurs when I try to view my basket.

Bottom line: `price_excl_tax` needs to be mandatory.

sasha0:
Fixes #1912

### DIFF
--- a/src/oscar/apps/partner/strategy.py
+++ b/src/oscar/apps/partner/strategy.py
@@ -210,7 +210,7 @@ class StockRequired(object):
     """
 
     def availability_policy(self, product, stockrecord):
-        if not stockrecord:
+        if not stockrecord or stockrecord.price_excl_tax is None:
             return Unavailable()
         if not product.get_product_class().track_stock:
             return Available()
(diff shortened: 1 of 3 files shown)


mvantellingen:
Shouldn't we use the pricing policy to determine if a price is available?

sasha0:
@mvantellingen right, I'll move check to pricing police. But in general, what do you think about solution?

src/oscar/apps/partner/strategy.py
```diff
@@ -204,7 +204,7 @@ class StockRequired(object):
     def availability_policy(self, product, stockrecord):
         if not stockrecord:
             return availability.Unavailable()
-        if not product.get_product_class().track_stock:
+        if not product.get_product_class().track_stock and stockrecord.price_excl_tax:
```
mvantellingen: This seems to make products which are free (price is 0) unavailable. Not sure we want, perhaps:

    if not stockrecord or stockrecord.price_excl_tax is None:
        return availability.Unavailable()

sasha0: Right. Otherwise looks good?

sasha0:
@mvantellingen updated PR. Please check.

PR #1914, opened 2015-11-22
Fix recommended products ordering on the frontend

### LINKED ISSUE #1901: Recommended Products have no order
We can see ranking on Recommended products in Dashboard only.
There are no ordering on Product detail page.

Oscar master, Django 1.8

sasha0:
Fixes #1901

### DIFF
--- a/src/oscar/apps/catalogue/abstract_models.py
+++ b/src/oscar/apps/catalogue/abstract_models.py
@@ -647,6 +647,12 @@ class AbstractProduct(models.Model):
     def num_approved_reviews(self):
         return self.reviews.approved().count()
 
+    @property
+    def sorted_recommended_products(self):
+        """Keeping order by recommendation ranking."""
+        return [r.recommendation for r in self.primary_recommendations
+                                              .select_related('recommendation').all()]
+
 
 class AbstractProductRecommendation(models.Model):
     """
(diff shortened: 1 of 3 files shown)


src/oscar/apps/catalogue/abstract_models.py
```diff
@@ -678,6 +678,12 @@ def num_approved_reviews(self):
         return self.reviews.filter(
             status=self.reviews.model.APPROVED).count()
 
+    @property
+    def sorted_recommended_products(self):
+        """Keeping order by recommendation ranking."""
+
+        return [r.recommendation for r in self.primary_recommendations.all()]
```
mvantellingen: This needs to be `self.primary_recommendations.select_related('recommendation').all()` I think, otherwise additional SQL queries are executed.

Also the empty line should be remove

sasha0: @mvantellingen ok done.

sasha0:
@mvantellingen already updated PR, please take a look.

sasha0:
@vdboor do you think it's ready for the merge? Thanks.

mvantellingen:
Can we add a unittest for this?

sasha0:
@mvantellingen yep, added test.

PR #1915, opened 2015-11-22
Create simple dockerfile to run the sandbox

mvantellingen:
This is still really basic (we can add elasticsearc etc in the future).

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,6 @@
+.cache
+.tox
+.tx
+__pycache__
+*.pyc
+node_modules/
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,23 @@
+FROM python:2.7
+ENV PYTHONUNBUFFERED 1
+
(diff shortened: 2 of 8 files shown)


sasha0:
@mvantellingen we had actually another similar PR in progress - #1366 .

mvantellingen:
@sasha0 Yeah I remember. Looking at that PR again it seems to do too much, I think we should just use this as a starting point and add improvements along the way.

mvantellingen:
Also perhaps good to mention is that one of the reasons I've added it is that we might want to run http://latest.oscarcommerce.com/ via the docker instance. It seems it is still running an old version.

Automatically building the sandbox via docker hub and deploying it somewhere might result in a simple to manage workflow.

mvantellingen:
Images of this branch are now automatically build and published at https://hub.docker.com/r/oscarcommerce/django-oscar-sandbox/

Simply run via:

    docker run -p 8080:8080  oscarcommerce/django-oscar-sandbox

PR #1916, opened 2015-11-22
fix a typo in layout.html

anentropic:


mvantellingen:
The original version seems to be correct, see https://github.com/django-oscar/django-oscar/commit/a1ce6ac78838b024a8b5bd8dd1dad251e852c672 and http://lesscss.org/#client-side-usage

Is this causing other issues?

anentropic:
oh ok, in that case I fixed the wrong one :)

see amended commit

mvantellingen:
Thanks :P

PR #1917, opened 2015-11-22
Add missing end slash - fix error when APPEND_SLASH is set to True.

michaelkuty:
RuntimeError: You called this URL via POST, but the URL doesn't end in a slash and you have APPEND_SLASH set. Django can't redirect to the slash URL while maintaining POST data. Change your form to point to host.cz/en/user-profile/orders/100030/1/ (note the trailing slash), or set APPEND_SLASH=False in your Django settings.

### DIFF
--- a/src/oscar/apps/customer/app.py
+++ b/src/oscar/apps/customer/app.py
@@ -101,7 +101,7 @@ class CustomerApplication(Application):
             url(r'^orders/(?P<order_number>[\w-]*)/$',
                 login_required(self.order_detail_view.as_view()),
                 name='order'),
-            url(r'^orders/(?P<order_number>[\w-]*)/(?P<line_id>\d+)$',
+            url(r'^orders/(?P<order_number>[\w-]*)/(?P<line_id>\d+)/$',
                 login_required(self.order_line_view.as_view()),
                 name='order-line'),
 


mvantellingen:
This seems like a workaround for a bug somewhere else. There is no reason that this url has to end with a slash.

michaelkuty:
Here is the form which cause my problem https://github.com/django-oscar/django-oscar/blob/master/src/oscar/templates/oscar/customer/order/order_detail.html#L135
This makes /en/user-profile/orders/100030/1 and POST raise RuntimError

mvantellingen:
I've justed tested it locally with the latest version of master and it works here. The url generated is `/en-gb/accounts/orders/100001/2` so it should just hit the view. APPEND_SLASH is only used when no initial route matched

mvantellingen:
See https://docs.djangoproject.com/en/1.8/ref/settings/#append-slash

michaelkuty:
Ok, thanks for your time and sorry for my mistake. 
By the way, I have one another question about oscar-accounts, If i tries to install via pip install git+..
i have error about missing README.md file in the installation, i tries to inspect, MANIFEST is there and other stuff looks good. Do you not have a similar problem ?

mvantellingen:
Regarding oscar accounts, I'm going to work with it in the coming weeks. But perhaps you can give https://github.com/django-oscar/django-oscar-accounts/pull/39 a try in the meantime?

michaelkuty:
Sorry, but I work on integration all oscar applications to Leonardo CMS which makes me totally busy.

PR #1920, opened 2015-11-23
Fix approved product review queryset

### LINKED ISSUE #395: Create template tag for reviews

sasha0:
Fixes #395

### DIFF
--- a/docs/source/releases/v1.2.rst
+++ b/docs/source/releases/v1.2.rst
@@ -66,7 +66,7 @@ Backwards incompatible changes in Oscar 1.2
   were not up-to-date anymore. These might return in the future as 
   separate repositories.
 
-- The `RecentReviewsManager`, `TopScoredReviewsManager` and 
+- The `ApprovedReviewsManager`, `RecentReviewsManager`, `TopScoredReviewsManager` and
   `TopVotedReviewsManager` managers are removed from the reviews app 
   since they were broken and unused.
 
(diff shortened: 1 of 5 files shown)


mvantellingen:
I've solved it locally using the following construct:

class ProductReviewQuerySet(models.QuerySet):
    use_for_related_fields = True

    def approved(self):
        return self.filter(status=self.model.APPROVED)


class AbstractProductReview:
    # ...
    objects = ProductReviewQuerySet.as_manager()
    # ...

This way we can use product.reviews.approved().all(). IMHO a cleaner way

sasha0:
@mvantellingen yeah, agree. I'll rework PR.

sasha0:
@mvantellingen done - reworked and added test.

mvantellingen:
Nice! It does break backwards compatibility by removing the `approved` manager. We might get away with a message in the release notes?

sasha0:
@mvantellingen Mb we could firstly merge you PR #1919 and then I'll update release notes?

mvantellingen:
Done :-)

sasha0:
@mvantellingen done too :-).

sasha0:
@maikhoepfel @mvantellingen already updated PR, please take a look.

mvantellingen:
Hi @sasha0 sorry for not looking faster at this. Could you update the PR with changes to 1.3? We can finally release this since our dependencies are now all 1.9 compatible

mvantellingen:
Rebased and merged

PR #1922, opened 2015-11-23
Fix errors with shipping discounts on free weightbased shipping.

vdboor:
The standard shipping methods all inherit from `shipping.methods.Base`, which has a `discount()` method.
The shipping models inherit from `shipping.abstract_models.AbstractBase`, which *does not* have a `discount()` method! This PR makes sure that standard shipping methods, and model-based shipping methods have the same base.

---
extra background:

The `Repository.apply_shipping_offer()` method returns the current shipping method object when there is no additional discount for an order. That works with the standard shipping methods that decent from `Base`, because `Base.discount()` nicely returns 0. (meaning: the shipping method does not provide an additional discount).

The `WeightBased` model does not have a working `discount()` method. It uses some copy-pasted code from `TaxInclusiveOfferDiscount`, but that does not make sense. Nor did it work, because `self.offer` would never be assigned.

Instead, it the model versions (`WeightBased`, `OrderAndItemCharges`) should have a similar default `discount()` method in their base class (`AbstractBase`) that also returns 0.

### DIFF
--- a/src/oscar/apps/shipping/abstract_models.py
+++ b/src/oscar/apps/shipping/abstract_models.py
@@ -39,6 +39,13 @@ class AbstractBase(models.Model):
     def __str__(self):
         return self.name
 
+    def discount(self, basket):
+        """
+        Return the discount on the standard shipping charge
+        """
+        # This method is identical to the Base.discount().
+        return D('0.00')
+
 
 class AbstractOrderAndItemCharges(AbstractBase):
     """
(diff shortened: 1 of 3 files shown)


maikhoepfel:
I am not surprised the (relatively rarely used) weight-based shipping methods weren't tested with shipping discounts. 
Your PR has more text than the commit changes, your code more docstrings than code, and you add tests. Looks good, thanks Diederik!

Issue #1929, opened 2015-12-02
Difficult to customize basket message logic (offer gained, offer lost, new total)


john-parton:
`apps.basket.views.get_messages` and `apps.basket.views.apply_messages` aren't customizable. It's possible to change the messages by overriding the templates, but if you want to pass in some additional context or disable a message completely, there's no good way to do that.

One simple solution would be to move `get_messages` and `apply_messages` to a class and then load in the class with `get_class`.

maikhoepfel:
Fixed by https://github.com/django-oscar/django-oscar/issues/1930

mvantellingen:
I assume we can close this then? Otherwise please re-open :-)

PR #1930, opened 2015-12-02
Basket Message Generator

john-parton:
Fix for https://github.com/django-oscar/django-oscar/issues/1929

The basket message functions were really opaque. I really don't want to customize `BasketView` AND `BasketAddView` AND  `SavedView` just to tweak some message logic.


maikhoepfel:
I like this. Thanks!

PR #1921, opened 2015-11-23
Fieldnames are not correct

maerteijn:

PR #1937, opened 2015-12-09
Fixes for dashboard offers

maerteijn:
* Fixes an issue with timezone aware datetime objects (django.utils.timezone takes care of this)
* When not all steps are completed in the wizard (eg directly save an edited offer) you would get a `'NoneType' object has no attribute 'save'` exception. This is because there is no serialized version of these objects (yet) as these forms are not initialized when pressing 'save' in the first wizard step.
* Added tests for pull request #1921


maikhoepfel:
Lovely work, thanks.

PR #1940, opened 2015-12-10
Remove unique=True on Benefit class

### LINKED ISSUE #685: Non abstract Benefit model with unique proxy_class is hard to modify or extend
offer/Benefit is a non abstract model so it can't be overridden/extended by overriding the application and making some new migrations for locally changed Benefit model. Second problem is unique proxy_class which prevents adding more than one Voucher with custom proxy_class (and  I have to set them as... Benefit is not abstract).

I've make manual migration to remove the unique from proxy_class and now I can use multiple vouchers with a proxy class ("most expensive for free") for different product sets with no problems. So the question is why Benefit model is done as it is? Why it is non-abstract and why proxy_class is unique?

maikhoepfel:
I ran into the issues detailed in #685. Basically, the uniqueness constraint
makes the custom benefit classes a lot less useful. We want to create slightly
altered versions of PercentageDiscounts with different values, and can't,
because we can only ever create one benefit with the custom class.

I don't see how removing the constraint could cause a problem.

Closes #685.

### DIFF
--- a/docs/source/releases/v1.2.rst
+++ b/docs/source/releases/v1.2.rst
@@ -47,6 +47,11 @@ What's new in Oscar 1.2?
 Minor changes
 ~~~~~~~~~~~~~
  - Fix missing page_url field in the promotions form (`#1816`_)
+ - Custom benefits now don't enforce uniqueness on the ``proxy_class``
+   field, making them more useful (`#685`_).
+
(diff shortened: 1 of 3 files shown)


maikhoepfel:
Merged because nobody complained (:

PR #1942, opened 2015-12-15
Range.all_products can return duplicates - ensure it doesn't

solarissmoke:
`Range.all_products` does a few OR joins on ManytoMany relationships. Because of the OUTER JOINs that are generated by this query, it is possible to end up with duplicates in the resulting queryset.

To reproduce, create a range which has only one product (which has been added using its SKU rather than by applying a category to the range). When the range is saved, the product will appear three (or four) times in the queryset returned by `all_products`.

This patch just adds a `distinct()` to the end of the query to ensure that this doesn't happen.


maikhoepfel:
That makes perfect sense. Thanks for testing and fixing this.

PR #1943, opened 2015-12-15
Revert use of safe template filter in product alert email templates

craigloftus:
1b549d1 made product titles disappear from the product alert emails because they use the `safe` filter inside `{% blocktrans %}`.

From the [Django docs](https://docs.djangoproject.com/en/1.9/topics/i18n/translation/#blocktrans-template-tag):
> To translate a template expression – say, accessing
> object attributes or using template filters – you need to bind the expression
> to a local variable for use within the translation block.

The templates in question are txt emails so I have used `{% autoescape off %}`
on the whole content as any escaping will be unwelcome and of no benefit. An
alternative would be `{% blocktrans title=alert.product.get_title|safe %}`

Note, I have also standardised on the use of `Product.get_title` in these
templates as that seems to be the more common than using `Product.__str__`?


maikhoepfel:
Thank you for the PR. I agree with most of your reasoning, and am marking this a bug because we need to resolve it before the next release. But I'd like to strongly discourage disabling autoescaping. Your alternative is the correct approach. I also absolutely agree with using `get_title` instead of `__str__`.

Could you update the PR to use your alternative syntax? Thank you.

craigloftus:
I will update the PR later today, but just for interest, what is the reason against using `{% autoescape off %}` in plain text emails? I actually copied this approach from the commtype plain text emails.

mvantellingen:
I think that it might be that some email clients will still try to parse plain text as html, but don't pin me on that. Always escaping is the safe way to go imho

PR #1945, opened 2015-12-17
Fix postcodes_regex.

dragonjun:
According to http://en.wikipedia.org/wiki/List_of_postal_codes,
Korea, South(KR) post code rule was changed on August 1, 2015.


maikhoepfel:
Excellent spot, thank you!

Issue #1910, opened 2015-11-14
Long rendered page when a lot of categories products


amirrpp:
_Sorry for my english._

When we have a lot of categories (and subcategories) the list of products render a long time because Django search products in all subcategories of current category.
#### I have:
- 1,100 Categories and subcategories
- 40,000 products
- The average depth of 4 categories
- In the navbar need to show categories first and second level of nesting. In this category you need to show only those that have products (at least one of the subcategories). Therefore, it is necessary to once again pass through the tree categories. Navbar appear on each page.
- On pages /catalogue/\* it is necessary to show the sidebar (the children of the currently selected category), the logic of both the navbar.
#### Possible solution cache search at a template. But:
- In the category affected by external factors
- Sidebar for each of your categories
- A list of products will be formed is still a long
### The best decision:

Show products only selected category. When we create the product and choose him a category, it must be added that the goods in the category-parents.
For example like this:

``` python
class AbstractProduct(models.Model):

    def save(self, *args, **kwargs):
        super(AbstractProduct, self).save(*args, **kwargs)
        product_categories = ProductCategory.objects.filter(product=self)
        for product_category in product_categories:
            product_category.save()


class AbstractProductCategory(models.Model):

    def save(self, *args, **kwargs):
        super(AbstractProductCategory, self).save(*args, **kwargs)
        product = self.product
        category = self.category
        parent = category.get_parent()
        if not parent:
            return
        ProductCategory.objects.update_or_create(product=product, category=parent)
```

Of course this has its problems.

What do you think?

mvantellingen:
Hi, this might be related to #1828 I've promised @maikhoepfel to make a test case for this issue

maikhoepfel:
And I promised to help fix this instead of reverting to the old code. I'm happy to leave this open until the issue resolved. @amirrpp, thank you for the detailed write-up. This helps us think about the issue.

shimulch:
Hi,

I have experienced same issue with v1.5. It was due to ```get_ancestors()``` call while ```get_absolute_url()``` has got called. So I have done a hack in my overriden Category class as following -

```python
    def _get_ancestors_cache_key(self):
        paths = [
            self.path[0:pos]
            for pos in range(0, len(self.path), self.steplen)[1:]
        ]
        key = str(paths)
        return key.replace(',', '').replace('\'', '').replace('[','N').replace(']','N').replace(' ', '')

    def get_ancestors(self):
        key = self._get_ancestors_cache_key()
[...]

Issue #1828, opened 2015-08-20
Performance issues 1.1


mvantellingen:
Since 1.1 the category slug and full_name are not denormalised anymore, which results in a massive number (5000+) of queries in the dashboard and the frontend.

The relevant commits are:
- https://github.com/django-oscar/django-oscar/commit/283fb600abacfb9a4b2eebb867b6e64f0297bd73
- https://github.com/django-oscar/django-oscar/commit/6f2119ffeb52109cff977ad8ae5b723128bd6c09

I think the only option is to start denormalising the values again. But if anyone has a better proposal then pleat let me know.

john-parton:
Which views/urls are generating the most queries?

I might have an idea for a solution.

john-parton:
The most straight forward solution is just to cache the property.

Something like this:
```python
@property
def full_slug(self):
    """
    Returns a string of this category's slug concatenated with the slugs
    of it's ancestors, e.g. 'books/non-fiction/essential-programming'.
    ...
    """
    full_slug = cache.get('Category.%s.full_slug' % self.pk, None)
    if full_slug is None:
        full_slug = self._slug_separator.join(category.slug for category in self.get_ancestors_and_self())
        cache.set(cache_key, full_slug, 60 * 60)
    return full_slug
```

sasha0:
@john-parton how about cache invalidation?

mvantellingen:
Also 2000 cache lookups are still to heavy. I'll have a local fix to restore the denormalisation again, will push after cleaning it up some more

john-parton:
Any movement on this?

You could use `django.utils.lru_cache` instead of using the cache backend. Performance would obviously be much better than using the cache backend. It has a `cache_clear()` method for cache invalidation as well.

john-parton:
@mvantellingen I just submitted PR https://github.com/django-oscar/django-oscar/pull/1946

@sasha0 Cache invalidation is handled in the PR.

mvantellingen:
@maikhoepfel: i've finally created an anonymized set of categories which causes the issues. It can be found here:
https://gist.githubusercontent.com/mvantellingen/c4aa2480c9317d2c3df2/raw/d2844da24c10b02faa0c8d17532cae8d63487e73/oscar-categories.sql

To load the data run `make sandbox` and then execute `sqlite3 sites/sandbox/db.sqlite < file.sql`.

mvantellingen:
[...]

PR #1946, opened 2015-12-17
Wrap Category.full_name and Category.full_slug properties with lru_cache

john-parton:
This is supposed to help mitigate some performance issues. Fix for https://github.com/django-oscar/django-oscar/issues/1910 and partial fix for https://github.com/django-oscar/django-oscar/issues/1828

I think we might want to add a setting: `CATEGORY_LRU_MAXSIZE` or something instead of hard coding 256 in.

Cache invalidation is handled in `Category.save()`. 

@mvantellingen Could you please test this and provide some feedback?

### DIFF
--- a/src/oscar/apps/catalogue/abstract_models.py
+++ b/src/oscar/apps/catalogue/abstract_models.py
@@ -16,6 +16,7 @@ from django.db.models import Count, Sum
 from django.utils import six
 from django.utils.encoding import python_2_unicode_compatible
 from django.utils.functional import cached_property
+from django.utils.lru_cache import lru_cache
 from django.utils.html import strip_tags
 from django.utils.safestring import mark_safe
 from django.utils.translation import ugettext_lazy as _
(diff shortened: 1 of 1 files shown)


mvantellingen:
This seems like a workaround imho. And it doesn't do cache invalidation properly when you scale to multiple app instances

john-parton:
@mvantellingen Thanks for the feedback. I was just trying to submit something other than reverting to the previous method.

Good point about cache invalidation. My production setup actually has two app instances, so I should be more careful about these sorts of issues.

maikhoepfel:
I agree a fix is needed, but have to also agree with @mvantellingen about this not being thread-safe. I boldly proclaimed me normalising the category handling is a good move and everything else can be solved by caching, so it's somewhat upon me to prove that. I'm quite busy in January, but am itching strongly to do some Oscar work, and am confident I'll manage to.

PR #1948, opened 2015-12-20
exclude current product from recently viewed

anentropic:
I found it a bit weird viewing a product detail page and having the same product listed underneath as recently viewed. A change to the templatetag makes it possible to exclude it.


sasha0:
:+1:

maikhoepfel:
That's a lovely change, thanks!

PR #1956, opened 2015-12-28
Add range to ConditionFactory

vkaracic:
In our project we get the products like this ``....condition.range.all_products()``. The ``Condition`` model has the ``range`` attribute which isn't added in the ``ConditionFactory``.


maikhoepfel:
I agree that the count condition needs a range. Thanks for the fix!

PR #1963, opened 2016-01-07
Add default ordering for Partner instances

maikhoepfel:
I was toying with the partner dashboard when I realized the order isn't 
consistent. So I fixed it.

It probably would make sense to go through all Oscar models and check that
they have an ordering defined. It should make sense for most of them.


maikhoepfel:
@mvantellingen Does this look fine to you?

PR #1964, opened 2016-01-08
address summary should use printable name

anentropic:
Currently when the shipping address is displayed in e.g. the checkout complete 'thank you' page it uses the full official name of the country.

For UK this comes out as "United Kingdom of Great Britain and Northern Ireland".

This looks a bit weird, arguably the address summary should use `printable_name` instead, which in this case would read "United Kingdom"


maikhoepfel:
I agree. Thank you for the PR! I will merge, and then fix the failing tests.

PR #1965, opened 2016-01-09
shipping - getting weight of product's parent if weight is not set for targeted product

kisiel:


maikhoepfel:
This is a good improvement. Thank you.

PR #1969, opened 2016-01-11
Enforce order of basket, order, wishlist lines

maikhoepfel:
When working on an Oscar project, I noted that the order of the basket lines
was different from the order of the order lines created from that basket. This
made me investigate how exactly Oscar approaches this.

It turned out that Oscar had no notion of ordering those lines. That means,
the order of them was determined by the database. That will work in most
cases, because at least PostgreSQL tends to return them in the order of
creation - but this is not guaranteed. From 
http://www.postgresql.org/docs/9.1/static/queries-order.html:

> If sorting is not chosen, the rows will be returned in an unspecified
> order. The actual order in that case will depend on the scan and join
> plan types and the order on disk, but it must not be relied on. A
> particular output ordering can only be guaranteed if the sort step is
> explicitly chosen.

Luckily, Django's AutoIntegerField guarantees that the primary keys are 
incremented. So to get the behaviour we most likely want, we just enforce
ordering by primary keys. I also documented how one would go about switching
to an Amazon-style handling. Implementing that is left as an exercise to the
reader.

### DIFF
--- a/docs/source/releases/v1.2.rst
+++ b/docs/source/releases/v1.2.rst
@@ -49,6 +49,9 @@ What's new in Oscar 1.2?
 Minor changes
 ~~~~~~~~~~~~~
  - Fix missing page_url field in the promotions form (`#1816`_)
+ - The order of basket, order and wishlist lines in now guaranteed
+   to be in the order of creation. Previously, this wasn't guaranteed,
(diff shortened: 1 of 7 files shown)


mvantellingen:
:+1:

PR #1973, opened 2016-01-13
Use option name instead of slug as field label

maikhoepfel:
When building the AddToBasketForm for a product with options, Oscar didn't
specify a label for the form field. Hence the field name was used, which is
the rather human-unfriendly slug. Options have a name field, so we can just
use that.


maikhoepfel:
Before:
ss1

After:
ss2

maikhoepfel:
@mvantellingen Do you want to have a look? Otherwise I'll just merge; I think it's quite safe.

sasha0:
Due to the conflicts, remote branch needs to be rebased against upstream master.

sasha0:
Alright, I've resolved conflicts by myself.

PR #1975, opened 2016-01-18
Move formset classes to separate modules to make it easier to override them

### LINKED ISSUE #1957: Formsets are hard to override
For instance, it's [hard to override](https://groups.google.com/forum/?fromgroups#!searchin/django-oscar/stockrecord$20form/django-oscar/sW7t_Y_b8GE/Yh8dswl1IL0J) the dashboard's `StockRecordForm`. If one wants to add or remove a field, one has to inherit the field, but also copy across `StockRecordFormSet`. 
Most likely, we want to move all of Oscar's formsets into `formsets` modules, and dynamically load the related forms.

sasha0:
Fixes #1957


sasha0:
Any suggestions about backwards compatibility @mvantellingen @vdboor ?

vdboor:
So am I right to understand that formsets need to move to a separate file, so they use `get_class()` to find the form, instead of hard-linking to it (because both were in the same file)? (that took me some effort to find out)

To preserve compatibility, I think oscar's `get_class()` should have a list of "moves" so it knows where to look;
* first in the custom app
* next in the old location of a custom app (new, using the "moves" table)
* in the standard oscar location

sasha0:
I suppose we'd better use "moves" for the oscar location rather then in custom app, since formset classes moved initially there.

sasha0:
@vdboor I've finally added *moves*, implemented fallback for loading and covered it with the tests. Thus, would be glad to know your opinion about that. 😉

solarissmoke:
This looks good to me - my only comment would be that there is no deprecation timeline for the backward-compatibility hack. I think we should state clearly when this will be removed (say Oscar 1.7), stick to that to minimise technical debt.

PR #1979, opened 2016-01-21
Remove Django 1.7 support

maikhoepfel:
We only support two versions of Django. That policy might change given that
Django has started doing LTS releases. But for now, it only makes sense to
remove Django 1.7 support.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,19 +32,15 @@ env:
         - PIP_WHEEL_DIR=$HOME/.cache/pip/wheels
         - PIP_FIND_LINKS=file://$HOME/.cache/pip/wheels
     matrix:
-        - DJANGO=Django==1.7.11
-        - DJANGO=Django==1.8.10
+        - DJANGO=Django==1.8.11
         - DJANGO=Django==1.9.3
 
 
 matrix:
     exclude:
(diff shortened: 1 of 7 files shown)


.travis.yml
```diff
@@ -32,19 +32,15 @@ env:
...
-        - DJANGO=Django==1.7.11
         - DJANGO=Django==1.8.8
         - DJANGO=Django==1.9.1
 
 
 matrix:
     exclude:
-        - python: 3.2
-          env: "DJANGO=Django==1.9.1"
+        # Python 3.3 is not supported by Django 1.9
```
john-parton: Did you mean Python 3.2 isn't supported by Django 1.9?

mvantellingen:
@maikhoepfel are you able to rebase this on master?

.travis.yml
mvantellingen: Django 1.9 only supports Python 3.4 and 3.5.

> Django 1.9 requires Python 2.7, 3.4, or 3.5. We highly recommend and only officially support the latest release of each series.
> 
> The Django 1.8 series is the last to support Python 3.2 and 3.3.

mvantellingen:
Rebased on master

PR #1981, opened 2016-01-22
Let error pages inherit from layout.html, not base.html

fghaas:
`layout.html` is the template that pulls in all the pretty Bootstrap bits and other styling. Error pages previously inherited just from `base.html`, which made them profoundly ugly. Let them inherit from `layout.html` instead for a more professional overall experience.


mvantellingen:
Thanks! Looks much better indeed :D

PR #1982, opened 2016-01-24
Fix postcodes_regex for NI (Nicaragua)

javierwilson:
According to http://en.wikipedia.org/wiki/List_of_postal_codes
Also according to official site http://www.correos.gob.ni/index.php/codigo-postal it's just 5 digits.


mvantellingen:
Ok, i'll take your word on it ;-)

PR #1983, opened 2016-01-26
Use a datepicker widget in some more forms in the dashboard.

solarissmoke:
Currently the date range inputs on the Reports and Fulfilment Statistics pages on the dashboard don't have a datepicker widget - the user has to manually enter the date in YYYY-MM-DD format with no cues that this is the format they should use.

This patch adds the datepicker widget to these fields. It also adds a small styling hack so that these display inline correctly - otherwise the label sits on its own line.


mvantellingen:
Thanks!

PR #1993, opened 2016-02-02
addressbook > address book

### LINKED ISSUE #1984: In shipping address checkout view, "addressbook" should be two words
The shipping_address template has "An address from your addressbook?" as the header above the user's address book. "addressbook" should be changed to "address book".
shipping_address___checkout___flyshop_-_welcome_to_surround_video_

cowlicks:
Closes #1984 (lol @ issue number)

I did this with:
```bash
$ grep -r --binary-files=without-match addressbook . | cut --delimiter=":" --field=1 | xargs -n1 sed -i 's/addressbook/address book/'

$ grep -r addressbook | cut --delimiter=" " -f3 | xargs -n1 sed -i 's/addressbook/address book/'
```
It is a two-liner because grep gives a differently formatted output for the "binary files". 

I'm not sure how localization with Transifex works, so please let me know if I did something wrong here. There is no semantic change so I don't thing the translations should change.


codeinthehole:
Thanks.

The transifex thing should be fine as the message files are generated from the same string in the template

PR #1995, opened 2016-02-05
Unicode slugs for Django 1.9

### LINKED ISSUE #1994: Unicode slugs
Since `SlugField` supports unicode since Django 1.9 (https://docs.djangoproject.com/es/1.9/ref/models/fields/#django.db.models.SlugField.allow_unicode), which is already supported by latest Oscar, we should probably think about proper unicode slugs support in the Oscar too.

Firstly, we could make optional conversion from unicode to ASCII in the slugify function - https://github.com/django-oscar/django-oscar/blob/master/src/oscar/core/utils.py#L41.

Also, perhaps there's no need in custom `oscar.models.fields.AutoSlugField`, since original one from `django_extensions` already support custom slugify function - https://github.com/django-extensions/django-extensions/blob/master/django_extensions/db/fields/__init__.py#L109.

Thanks to @pwalsh for the help in discovering this issue.

sasha0:
Fixes #1994 

Initial feature design has trade-offs in favor of backwards compatibility.
Improvement suggestions welcomed.

### DIFF
--- a/src/oscar/apps/basket/abstract_models.py
+++ b/src/oscar/apps/basket/abstract_models.py
@@ -14,6 +14,7 @@ from oscar.apps.offer import results
 from oscar.apps.partner import availability
 from oscar.core.compat import AUTH_USER_MODEL
 from oscar.core.utils import get_default_currency
+from oscar.models.fields.slugfield import SlugField
 from oscar.templatetags.currency_filters import currency
 
 
@@ -586,7 +587,7 @@ class AbstractLine(models.Model):
     # We can't just use product.id as you can have customised products
     # which should be treated as separate lines.  Set as a
     # SlugField as it is included in the path for certain views.
-    line_reference = models.SlugField(
+    line_reference = SlugField(
         _("Line Reference"), max_length=128, db_index=True)
 
     product = models.ForeignKey(
--- a/src/oscar/core/utils.py
+++ b/src/oscar/core/utils.py
@@ -6,11 +6,12 @@ import logging
 from django.conf import settings
 from django.shortcuts import redirect, resolve_url
 from django.template.defaultfilters import date as date_filter
-from django.template.defaultfilters import slugify as django_slugify
 from django.utils import six
 from django.utils.http import is_safe_url
 from django.utils.module_loading import import_string
+from django.utils.text import slugify as django_slugify
 from django.utils.timezone import get_current_timezone, is_naive, make_aware
+
 from unidecode import unidecode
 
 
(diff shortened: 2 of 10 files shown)


pwalsh:
Would be great to get this looked at. Allows us to move forward with Django 1.9 which (finally!) accepts that people on the web actually do use unicode slugs.

src/oscar/models/fields/slugfield.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class SlugField(DjangoSlugField):
+    def __init__(self, *args, **kwargs):
+        # not override parameter if it was passed explicitly,
+        # so passed parameters takes precedence over the setting
+        if getattr(settings, 'OSCAR_SLUG_ALLOW_UNICODE', None) and \
+                'allow_unicode' not in kwargs:
+            kwargs['allow_unicode'] = True
+
```
mvantellingen: `kwargs.setdefault('allow_unicode', settings.OSCAR_SLUG_ALLOW_UNICODE)`

src/oscar/core/utils.py
```diff
@@ -39,9 +44,11 @@ def slugify(value):
         slugifier = import_string(slugifier)
 
     # Use unidecode to convert non-ASCII strings to ASCII equivalents where
-    # possible.
-    value = slugifier(unidecode(six.text_type(value)))
+    # possible if unicode is not allowed to contain in slug.
+    if not getattr(settings, 'OSCAR_SLUG_ALLOW_UNICODE', None):
```
mvantellingen: The `settings.OSCAR_SLUG_ALLOW_UNICODE` should always be set right? So better to use it directly (also on the other places)

src/oscar/core/utils.py
```diff
@@ -6,7 +6,7 @@
 from django.conf import settings
 from django.shortcuts import redirect, resolve_url
 from django.template.defaultfilters import date as date_filter
-from django.template.defaultfilters import slugify as django_slugify
+from django.utils.text import slugify as django_slugify
```
mvantellingen: import order

src/oscar/apps/basket/abstract_models.py
```diff
@@ -10,6 +10,7 @@
 from django.utils.translation import ugettext_lazy as _
 
 from oscar.apps.basket.managers import OpenBasketManager, SavedBasketManager
+from oscar.models.fields.slugfield import SlugField
```
mvantellingen: import order

mvantellingen:
Looks really nice @sasha0 ! Few minor things and perhaps you can add some small text in the release notes for 1.2?

sasha0:
Sure, will do.

pwalsh:
Thanks for reviewing @mvantellingen

sasha0:
@mvantellingen I've updated PR, please take a look. Thanks.

PR #1997, opened 2016-02-06
Ran $(yes | isort) in the root-folder of the project

atombrella:
Although a previous commit (I've been browsing the history) was performed, lots of files were neglected, or import were not sorted again after new commits. Maybe an isort-test, similar to Django's testsuite, would be a nice amendment.


mvantellingen:
This actually formats the imports incorrectly. The order (see https://www.python.org/dev/peps/pep-0008/#imports) is:

1. standard library imports
2. related third party imports
3. local application/library specific imports

PR #1998, opened 2016-02-08
Serve parent image if child image is missing

maikhoepfel:
When child products are added to the basket in stock Oscar, there never will
be an image associated with them because the Oscar dashboard doesn't support
it to keep things simple.

But if that is the case, we can easily attempt to serve the parent product
image, which is the better default.


sasha0:
:+1:

sasha0:
I think we could merge this PR. Do you agree @vdboor @mvantellingen ?

PR #2000, opened 2016-02-10
Prevent trying to save a negative quantity to a Basket

crgwbr:
basket.Line.quantity is a positive integer field, but the quantity sumation in
basket.Basket.add_product made it possible to try and save negative values. This
makes add_product never go below 0-quantity, thereby preventing a db exception
from being thrown. Also adds a test for the described behavior.

### DIFF
--- a/src/oscar/apps/basket/abstract_models.py
+++ b/src/oscar/apps/basket/abstract_models.py
@@ -217,7 +217,7 @@ class AbstractBasket(models.Model):
                 line.attributes.create(option=option_dict['option'],
                                        value=option_dict['value'])
         else:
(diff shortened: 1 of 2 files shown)


mvantellingen:
:+1:

bufke:
@mvantellingen is it intentional that a 0 quantity line can exist? I've created some funky scenarios with oscar-api by using  api-basket-add-product with a -1 quantity. It kind of makes sense I could use that to go from 2 to 1. Going to 1 to 0 while keeping the line is really odd. I'm unsure if this should be addressed on django-oscar or django-oscar-api.

PR #2001, opened 2016-02-11
Fixing links for Twitter Bootstrap (it was pointing to an outdated ve…

atombrella:
…rsion) and for django-oscar-eway


sasha0:
:+1: I'd suggest to rename commit message to something like "Fix outdated links for Bootstrap and django-oscar-eway".

codeinthehole:
Thanks for the corrections.

PR #2002, opened 2016-02-15
Re-add Category.full_name and Category.full_slug

john-parton:
Here we go again!

Trying to fix https://github.com/django-oscar/django-oscar/issues/1828

This PR adds `full_slug` and `full_name` attributes.

I think the big reason that they were removed before is because they're messy and it's hard to understand how they work. I fixed that by putting all of the denormalization logic into a single `classmethod` `denormalise_fields` which I then wired up to a signal.

I added a management command that wraps that method. It will be necessary to call it if you migrate from `1.1`.

@maikhoepfel Do you think an approach like this would be a reasonable compromise?

Please let me know if you have any questions.

Thanks.
John

### DIFF
--- a/src/oscar/apps/catalogue/abstract_models.py
+++ b/src/oscar/apps/catalogue/abstract_models.py
@@ -1,12 +1,12 @@
 import logging
 import os
 from datetime import date, datetime
+from itertools import count
 
 from django.conf import settings
 from django.contrib.contenttypes.fields import GenericForeignKey
 from django.contrib.contenttypes.models import ContentType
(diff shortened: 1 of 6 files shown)


sasha0:
@john-parton I think this PR could be rebased against upstream master to exclude merge commits.

john-parton:
@sasha0 Other than that, are there any changes that need to be made? Tests?

Does my overall approach make sense here?

sasha0:
Closing since we'll use another approach to address this, similar to Wagtail, per our latest discussion with @mvantellingen.

PR #2003, opened 2016-02-16
Fix Customizing the Base Dashboard App

prikhi:
When customizing the base `dashboard` app, the `config_name` should be `DashboardConfig`, not `DashboardDashboardConfig`.

### DIFF
--- a/src/oscar/core/customisation.py
+++ b/src/oscar/core/customisation.py
@@ -37,7 +37,7 @@ def subfolders(path):
 
 
 def inherit_app_config(local_app_path, app_package, app_label):
-    if 'dashboard' in app_label:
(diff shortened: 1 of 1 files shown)


src/oscar/core/customisation.py
```diff
@@ -37,7 +37,7 @@ def subfolders(path):
 
 
 def inherit_app_config(local_app_path, app_package, app_label):
-    if 'dashboard' in app_label:
+    if 'dashboard' in app_label and app_label != 'dashboard':
```
sasha0: Second check is redundant, they producing the same result.

sasha0:
True, newly forked dashboard app config has `DashboardDashboardConfig` config class, inherited from `config.DashboardDashboardConfig`. Thanks for catching that.

PR #2010, opened 2016-02-28
Limit a percentage discount to 100%

moloughlin:
Noticed that a voucher could apply a >100% discount, leading to a negative total.


codeinthehole:
Good spot - thanks.

Issue #2013, opened 2016-03-02
Allow site override in /apps/order/utils.py:OrderCreator.create_order_model


mattdrayer:
Hi there!

I'm running into a site-related issue with the OrderCreator.  We are trying to pass in a site entity as an override, but the current implementation appears to ignore any provided value in favor of the Site's get_current() workflow.  Typically this wouldn't be an issue, however this is a multi-host scenario and we'd like to be able to indicate the site as part of the request to this class.  See here:

https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/order/utils.py#L104

I'd like to propose that the setting of the "site" key be moved from its current location to after the "if extra_order_fields..." conditional on line 120.  We would perform a quick check to see if the site key had not been set via extra_order_fields -- if not we'd execute Site._default_manager.get_current() as is currently happening.

I am happy to make this change and submit a pull request to the django-oscar project if that's the best way to move forward?  If you could let me know that would be great.

Thanks!

Matt

sasha0:
Closing since related PRs #2014 and #2080 already merged.

PR #2014, opened 2016-03-02
mattdrayer/issue/2013: Allow site override in OrderCreator

mattdrayer:
Accompanying change set for Issue #2013 -- happy to revise.

Cheers,

Matt

### DIFF
--- a/src/oscar/apps/order/utils.py
+++ b/src/oscar/apps/order/utils.py
@@ -101,7 +101,6 @@ class OrderCreator(object):
         """
         order_data = {'basket': basket,
                       'number': order_number,
-                      'site': Site._default_manager.get_current(),
                       'currency': total.currency,
                       'total_incl_tax': total.incl_tax,
                       'total_excl_tax': total.excl_tax,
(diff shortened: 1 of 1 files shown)


src/oscar/apps/order/utils.py
```diff
@@ -119,6 +118,8 @@ def create_order_model(self, user, basket, shipping_address,
             order_data['status'] = status
         if extra_order_fields:
             order_data.update(extra_order_fields)
+        if 'site' not in order_data:
+            order_data['site'] = Site._default_manager.get_current()
```
sasha0: Could be replaced with `order_data.setdefault('site', Site._default_manager.get_current())`
Original idea by @mvantellingen :wink:

mvantellingen:
Thanks, can you also add a PR with a unittest for this? This makes sure we don't break it again in the future.

src/oscar/apps/order/utils.py
mvantellingen: Using setdefault() here could cause an unnecessary call to get_current()

sasha0:
@mvantellingen added tests in the PR #2080.

mattdrayer:
Sorry guys, I have been meaning to get to this but time's been a bit tight due to a big project @ edX.

sasha0:
@mattdrayer no worries - I suppose my tests will be enough for now.

PR #2019, opened 2016-03-23
Fixed performance issue in offer dashboard detail page

vdboor:
There was an N-query issue while displaying all orders the offer was applied to.

On one of our offers, it was a jump from 1000+ queries back to 24.


mvantellingen:
Thanks. I'm preparing a 1.2.1 release, will add this

PR #2020, opened 2016-03-23
Proposal: show 404 pages for inactive offers

vdboor:
Offers that are suspended are not visible in the list view, but will still be shown as detail page.
I wonder whether this is intended, or accidentally.

One of my clients requested to hide such pages (as I've done in the project code).

With this PR I'd hope to receive feedback on whether this should be the default, or not.


sasha0:
Agree. How about you @mvantellingen ?

mvantellingen:
Agreed, merging 👍

PR #2021, opened 2016-03-23
Widen the Source.reference field 

jbzdak:
Widen the `Source.reference` field to match Stripe payment processor maximal length for payment tokens.

Tests and migrations ran well on my local machine, this change should have no ill effects whatsoever. 

Please review :)!


sasha0:
Could you please refer to the token length requirement? Because from the docs https://stripe.com/docs/api#token_object and my experience tokens are not longer than current limit (128 chars), usually it's >30 chars.

Thanks.

jbzdak:
Here is the API Forward compatibility page: https://stripe.com/docs/upgrades#what-changes-does-stripe-consider-to-be-backwards-compatible, specifically, following is considered as a forward compatible change: 

> Changing the length or format of object IDs or other opaque strings. 
This includes adding or removing fixed prefixes (such as ch_ on charge IDs). 
You can safely assume object IDs we generate will never exceed 255 characters, but you should be able to handle IDs of up to that length. If for example you’re using MySQL, you should store IDs in a VARCHAR(255) COLLATE utf8_bin column (the COLLATE configuration ensures case-sensitivity in lookups).

PS. @bradenmacdonald this is one of the off-shots of Stripe payment processor, could you take over?

bradenmacdonald:
@jbzdak No problem, I can follow-up.

mvantellingen:
Seems good to me!

PR #2022, opened 2016-03-24
Fixes re-ordering of products in ranges when CSRF cookie is HTML-Only

pjstevns:


sasha0:
Looks good.

mvantellingen:
Dankje! :+1:

PR #2035, opened 2016-04-13
Use get_class for imports to partner.prices and partner.availability

crgwbr:
Makes codebase more consistent and extensible.


sasha0:
Looks good to me. What do you think @mvantellingen @vdboor ?

mvantellingen:
👍

sasha0:
@crgwbr thanks for the contribution!

PR #2038, opened 2016-04-23
Set BOM in UnicodeCSVWriter in PY2 for better csv reader compatibility

SteinRobert:
We noticed that some Users, which are using MS Excel on Mac, have troubles with the encoding of the files created by the `UnicodeCSVWriter`. Setting the byte order mark at the beginning of the file solved the issue.


mvantellingen:
Looks nice. Are you able to add a unittest?

SteinRobert:
Sure I can do that.

sasha0:
Closed in favor of PR #2300 which includes test.

PR #2047, opened 2016-05-18
Send signals on order and order line status change.

michaelkuty:
Make possible to handle status change for order and their lines.
Needs documentation, but I can make it if you say that these signals are useful and will be merged.


sasha0:
I think this PR now contains changes, unrelated to the signals introduction.

michaelkuty:
Sorry my bad, new PR #2075

PR #2050, opened 2016-05-27
Add missing migration to ProductReview

solarissmoke:
There seems to be a missing migration after the changes introduced in [Remove Django 1.7 support](https://github.com/django-oscar/django-oscar/commit/afe08db40965aee8df797d55c64a8de7fc0f3e50).


mvantellingen:
Thanks. It should be fixed in master and i've added a unittest to test for missing migrations

PR #2051, opened 2016-05-30
Issue/675/per variant image support

akash1234:


vdboor:
Hi! Thanks for making big contributions to oscar.

In all honestly, I wonder whether this is the path to take. Product variants are a `Product` model too, so they can have different images. That's the technique I've applied at https://waka-waka.com/store/catalogue/wakawaka-light_15/ for example. That URL is the "parent" product, while the selectbox selects the variant to buy. There is no need for a new M2M relation to achieve all that.

Could you comment on what this code is meant to achieve differently?

sasha0:
Closing since it's duplicate of #1643. Let's continue discussion in the original PR.

PR #2054, opened 2016-06-03
Allow haystack 2.5.0.dev1

mvantellingen:
This will finally allow us to release a Django 1.9 compatible version


mvantellingen:
This requires ``pip install --pre``

sasha0:
Perhaps we should add this flag into `Makefile` too.

mvantellingen:
Fixed in master

PR #2056, opened 2016-06-03
Fix bootstrap column layout in shipping_address template

solarissmoke:
The [port to Bootstrap 3](https://github.com/django-oscar/django-oscar/commit/d702f47f127081e00dfadb35de0f0615191134f5#diff-f5387c20752230674cc3d54950d8d388L20) was a little broken for this template. The intention was for addresses to appear two-per-row on large screens but this wasn't actually happening and they all appeared in one column on the left of the page. 

This fixes the column layout to work as intended.

### DIFF
--- a/src/oscar/templates/oscar/checkout/shipping_address.html
+++ b/src/oscar/templates/oscar/checkout/shipping_address.html
@@ -20,12 +20,12 @@
     {% if user.is_authenticated %}
         {% if addresses %}
             <h3>{% trans "An address from your address book?" %}</h3>
-            <div class="row">
-                <div class="choose-block col-sm-6">
-                    <ul class="list-unstyled">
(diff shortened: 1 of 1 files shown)


mvantellingen:
Nice one thanks!!, I remember I had the same issue 👍

mvantellingen:
@solarissmoke Seems the form below the addresses from the address book has some alignment issues (now) (with class new_shipping_address). See the new sandbox site: http://d1nv2fhv7mlo67.cloudfront.net/nl/checkout/shipping-address/ (dns will be fixed asap)

solarissmoke:
@mvantellingen there were a number of issues in the template that were messing up the layout on this page. I have pushed a fix to #2073.

PR #2057, opened 2016-06-14
Update the currency filter to support the same params as babel.numbers.format_currency

craigloftus:
The current implementation of the `currency` filter does not allow all the parameters provided by `format_currency` to be configured, this PR adds to settings `OSCAR_CURRENCY_DIGITS` and `OSCAR_CURRENCY_FORMAT_TYPE`.

As `currency` is a filter I've used settings rather than arguments to configure the parameters.

### DIFF
--- a/docs/source/ref/settings.rst
+++ b/docs/source/ref/settings.rst
@@ -368,6 +368,29 @@ the ``format_currency`` function from the `Babel library`_.
 
 .. _`Babel library`: http://babel.pocoo.org/docs/api/numbers/#babel.numbers.format_currency
 
+``OSCAR_CURRENCY_DIGITS``
+-------------------------
+
+Default: ``True``
+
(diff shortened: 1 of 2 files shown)


vdboor:
Thanks for your effort and pull request!

Before we merge this, I wonder, shouldn't this be handled differently to support multiple currencies?
For example:

```
OSCAR_CURRENCY_FORMAT = {
    'USD': {
        'currency_digits': 2,
        'currency_digits': "...",
    },
    'EUR': { .. }
}
```

Those values could serve as base for the ``kwargs``; for example `kwargs.update(OSCAR_CURRENCY_FORMAT.get(currency, {}))`

Otherwise we keep adding settings that prevent multicurrency usage.

sasha0:
Closing in favor of PR #2284 which is updated version of this PR with multi-currency support. Thanks.

PR #2061, opened 2016-07-01
passing request object to place order view

mjrulesamrat:
It is possible to use multiple sites in the project. While placing order, oscar uses site object form 'SITE_ID' settings from project settings. I have passed request object through mixins and it does not affect any functionality but it enables developer to get current_site from SITE model.

If you feel this should be included in oscar then please accept my pull request.

Any suggestions are welcome.

Thanks,
Jay Modi


sasha0:
Thanks for your contribution.

Please rebase your pull request against upstream master and fix pylint warnings.

I would also suggest to implement test which will verify that submitted order has proper site.

Thanks!

sasha0:
Closed in favor of #2256, which is rebased version of this PR.

PR #2062, opened 2016-07-06
Added a default ordering so get_queryset will not query the database again

maerteijn:
I tried customizing the basket view by adjusting the ``basket_content.html`` to display the discounts of each line. I encountered that the discounts were not there, however, the summary with the totals below it shows the discounts.

After some debugging I figured out that the queryset that is used (``basket.all_lines()``) contains the discounts applied as it uses a cached resultset with all offers applied:
```python
# oscar/apps/basket/views.py
class BasketView(ModelFormSetView):
    model = get_model('basket', 'Line')
    ...
    def get_queryset(self):
        return self.request.basket.all_lines()
```

This Formset which is used in the view is using this queryset, so the question is why the discount values are empty:
```python
# oscar/apps/basket/forms.py
class BaseBasketLineFormSet(BaseModelFormSet):

    def __init__(self, strategy, *args, **kwargs):
        self.strategy = strategy
        super(BaseBasketLineFormSet, self).__init__(*args, **kwargs)
        ...
```
The BaseModelFormSet is using the following code for getting the queryset:
```python
# django/forms/models.py
class BaseModelFormSet(BaseFormSet):
    ...
    def get_queryset(self):
        if not hasattr(self, '_queryset'):
            if self.queryset is not None:
                qs = self.queryset
            else:
                qs = self.model._default_manager.get_queryset()

            # If the queryset isn't already ordered we need to add an
            # artificial ordering here to make sure that all formsets
            # constructed from this queryset have the same form order.
            if not qs.ordered:
                qs = qs.order_by(self.model._meta.pk.name)

            # Removed queryset limiting here. As per discussion re: #13023
            # on django-dev, max_num should not prevent existing
            # related objects/inlines from being displayed.
            self._queryset = qs
        return self._queryset
```
So when the queryset is not ordered, it will query the database again with an ``order_by``, so then all discount values are lost. 

So that's the reason of this pull request :)

### DIFF
--- a/src/oscar/apps/basket/abstract_models.py
+++ b/src/oscar/apps/basket/abstract_models.py
@@ -125,7 +125,8 @@ class AbstractBasket(models.Model):
                 self.lines
                 .select_related('product', 'stockrecord')
                 .prefetch_related(
-                    'attributes', 'product__images'))
+                    'attributes', 'product__images')
+                .order_by(self._meta.pk.name))
         return self._lines
 
     def is_quantity_allowed(self, qty):
--- a/tests/integration/basket/form_tests.py
+++ b/tests/integration/basket/form_tests.py
@@ -1,14 +1,32 @@
+from decimal import Decimal as D
+
 from django.test import TestCase
 from django.conf import settings
 import mock
 
 from oscar.apps.basket import forms
+from oscar.apps.offer.utils import Applicator
+from oscar.core.loading import get_model
 from oscar.test import factories
+from oscar.test.basket import add_product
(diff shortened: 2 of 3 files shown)


vdboor:
This seems like a reasonable solution, given that the `Line` model already orders to `id` by default.

@mvantellingen would a unit test be needed here to make sure this is not broken, or just a small comment to mention why the extra `.order_by()` is there?

maerteijn:
@vdboor I added some tests for it

mvantellingen:
Looks good to me!

PR #2065, opened 2016-07-24
Update debug toolbar.

pabu:
Require debug toolbar 1.5.0 rather than 1.4.0 and thereby avoid a bug in the sql panel that breaks the entire sandbox default installation.


mvantellingen:
Thanks!

PR #2067, opened 2016-07-25
Fix test requirements for tox and travis.

sasha0:


mvantellingen:
I actually moved it to setup.py to minimize the number of requirement files: see https://github.com/django-oscar/django-oscar/commit/21ebd1178db012575f1d315bc3e6598d966cbd74

This allows you to use ``pip install -e .[test]``

sasha0:
Right. But then test and docs section where removed from the main dependencies - https://github.com/django-oscar/django-oscar/commit/0693f1cb35dc44909d573507e7c67388a140d3ae#diff-b4ef698db8ca845e5845c4618278f29aL5

Intention is to have test requirements in one file. Maybe we could just add `-e .[test]` to the requirements_tests.txt instead of package list to avoid duplication?

mvantellingen:
But the test requirements are already in one file: setup.py. Can you explain the use-case? I might be missing it. Or let's discuss via slack tomorrow :-)

maerteijn:
Change:
```ini
 [testenv]
 commands = python runtests.py
 deps = 
    -r{toxinidir}/requirements_tests.txt
    1.8: django==1.8.13
    1.9: django==1.9.7
```
to:
```ini
 [testenv]
 install_command = pip install -e .[test]
 commands = python runtests.py
 deps = 
    -r{toxinidir}/requirements_tests.txt
    1.8: django==1.8.13
    1.9: django==1.9.7
```
And you both achieve what you want right? :)

mvantellingen:
@maerteijn yeah that correct. And then just call py.test instead of runtests.py

sasha0:
Won't work:

`tox.ConfigError: ConfigError: 'install_command' must contain '{packages}' substitution`

maerteijn:
Something like this then:
```ini
[testenv]
 commands =
     pip install -e .[test]
     python runtests.py
 deps = 
    1.8: django==1.8.13
    1.9: django==1.9.7
```

Otherwise you are right to move the test dependencies to a separate file (and read this file in ``setup.py``) to avoid duplication

mvantellingen:
Last one should work. Also you can add ``deps = -rrequirements.txt`` so that tox can re-use the envs

sasha0:
How about this:
```
[testenv]
whitelist_externals = make
commands =
     make install
     python runtests.py
deps = 
    1.8: django==1.8.13
    1.9: django==1.9.7
```

sasha0:
Alright, let's use previous one. @maerteijn thanks for the suggestion!

PR #2068, opened 2016-07-30
Prevent product attribute value reset during validation.

### LINKED ISSUE #2015: Cannot edit attribute values when the product class has an `entity` type attribute
To reproduce (I just tested on repo's master branch, but I found it on the 1.1.1 release)
1. Create a product class with an a `text` type attribute and a `entity` type attribute.
2. Create a product of this class and set some value for the `text` attribute.
3. Save.
4. Set the text attribute to a different value.
5. Save it, but the value won't get updated.

It seems that the existence of an entity type attribute on the product class messes something up

sasha0:
Fixes #2015

### DIFF
--- a/src/oscar/apps/catalogue/product_attributes.py
+++ b/src/oscar/apps/catalogue/product_attributes.py
@@ -19,12 +19,15 @@ class ProductAttributesContainer(object):
         self.product = product
         self.initialised = False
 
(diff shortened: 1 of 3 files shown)


sasha0:
Will soon add tests for product attributes save in the dashboard.

sasha0:
Test is ready finally.

PR #2073, opened 2016-08-02
Layout fixes in shipping_address template.

solarissmoke:
Follow-up from #2056 to fix a few more issues in this template:

- Add an outer column class to the main block, because the parent template expects one.
- Remove use of UL in the address list, because it causes styling issues.
- Fix div closure logic in for loop.

There is also one line of dashboard CSS that was not compiled from the LESS.


mvantellingen:
Works perfectly! Thanks for the quick response :-)

PR #2075, opened 2016-08-02
Send signals on order and order line status change.

michaelkuty:
New PR related to #2047


mvantellingen:
hi @michaelkuty, so this is a pretty late response ;-), would like to merge this but it needs unittests Are you able to add these?

sasha0:
Closed in favor of PR #2270, which includes tests.

PR #2078, opened 2016-08-02
Change the default argument of ProductReview.status to a callable.

solarissmoke:
The `OSCAR_MODERATE_REVIEWS` setting is [used to set the default value](https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/catalogue/reviews/abstract_models.py#L56-L59) on the `status` field in `AbstractProductReview`.

The problem with this approach is that if a project changes `OSCAR_MODERATE_REVIEWS` then the `status` field definition changes and Django expects you to create a migration. This is problematic if you have not forked the reviews app - and you shouldn't have to just to change the setting.

This patch changes `default` to use a callable which will fetch the setting at runtime. I've also added some tests.


sasha0:
Thanks for the contribution!

PR #2083, opened 2016-08-04
Feature/dashboard style fix

mvantellingen:
This updates a number of dashboard templates to use panels instead of table captions. Looks a bit cleaner.

It also removes the background image (the dots)


src/oscar/static/oscar/css/dashboard.css
```diff
@@ -5203,9 +5203,6 @@ button.close {
 .alert .alertinner p:first-child {
   margin: 0;
 }
-body {
-  background: url("../img/ui/dashboard/bg_subtle_dots.png") repeat scroll 0 0 #fff;
```
solarissmoke: Probably should delete the image file as well.

sasha0: True. Removed.

mvantellingen:
I stopped working on this due to the completely new approach, see http://oscar.mvantellingen.nl/
I'm still not really happy with the sidebar menu, but perhaps this is something we can continue on?

sasha0:
I was going to finalize this styling improvement, since it has left not too much, and continue to work on the new dashboard layout, which is going to take major efforts and does not seem to feasible to finish for the next release. WDYT?

mvantellingen:
I think we should just move to the new layout directly instead of implementing a cleanup first. Otherwise people need to update their custom dashboard layouts twice.

If you want start working on the new layout then we should discuss it briefly via slack? :-)

sasha0:
Agreed. Shall we close this PR then?

PR #2103, opened 2016-08-17
Fix determination of hurry_mode for alerts

### LINKED ISSUE #2100: Hurry mode for customer alerts is never True
[The code](https://github.com/django-oscar/django-oscar/blob/62014cf6e36f5e1e612adc47dcbd7f859db3be4b/src/oscar/apps/customer/alerts/utils.py#L68-L77) to determine if hurry mode is set for a customer alert is checking that the number of active alerts are _less_ than the total number of items in stock for a product. 

```
if num_stockrecords == 1:
    num_in_stock = stockrecords[0].num_in_stock
    # hurry_mode is false if num_in_stock is None
    hurry_mode = num_in_stock is not None and num_alerts < num_in_stock
else:
    result = stockrecords.aggregate(max_in_stock=Max('num_in_stock'))
    hurry_mode = result['max_in_stock'] is not None and \
        num_alerts < result['max_in_stock']
```

I think it should be the reverse - check if the number of active alerts for a product are _greater_ than the number in stock. I'll submit a pull request to fix this.

digicase:
Hurry mode should be True if there are more active alerts than available stock for a product. Fixes #2100

### DIFF
--- a/src/oscar/apps/customer/alerts/utils.py
+++ b/src/oscar/apps/customer/alerts/utils.py
@@ -70,11 +70,11 @@ def send_product_alerts(product):
     if num_stockrecords == 1:
         num_in_stock = stockrecords[0].num_in_stock
         # hurry_mode is false if num_in_stock is None
-        hurry_mode = num_in_stock is not None and num_alerts < num_in_stock
+        hurry_mode = num_in_stock is not None and num_alerts > num_in_stock
     else:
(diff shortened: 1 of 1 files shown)


sasha0:
Thanks for the contribution. How about adding a test? Thanks.

sasha0:
Closing in favor of #2304, which is similar to this one and includes tests.

Issue #2114, opened 2016-09-08
productIndex should have a template with relative path instead of the absoute oscar path


kkx:
[ProductIndex.text: template_name field](https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/search/search_indexes.py#L15) 
with relative path, it is easier to override the search index template(item_text.txt) without the need of subclassing the entire search module.
I will submit a PR.

sasha0:
Merged. Thanks.

PR #2115, opened 2016-09-08
change productIndex text templete to a relative path

kkx:
This PR is related to the issue [2114](https://github.com/django-oscar/django-oscar/issues/2114)


sasha0:
Thanks for the contribution.

Issue #2116, opened 2016-09-09
user search logics is wrong in dashboard.users page


kkx:
As described in [issue 1337](https://github.com/django-oscar/django-oscar/issues/1337) too, Oscar assumed that [users only have 2 parts as firsname and last name](https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/dashboard/users/views.py#L85), which is not true. For example: Pablo Ruiz Picasso has 3 parts both Ruiz Picasso are his lastname, in the current version, this user would not be searched in the dashboard.
I am going to submit a PR for this fix

sasha0:
Closing since PR #2117 already merged.

PR #2117, opened 2016-09-09
create a dynamic user name filter on name parts

kkx:
This PR is related to [issue 2116](https://github.com/django-oscar/django-oscar/issues/2116).
With this approach, user results now depends on how specific the query is. With the complete user names, the result would be more exactly, which gets most of use cases.

### DIFF
--- a/src/oscar/apps/dashboard/users/views.py
+++ b/src/oscar/apps/dashboard/users/views.py
@@ -82,12 +82,11 @@ class IndexView(BulkEditMixin, SingleTableMixin, FormMixin, TemplateView):
             # If the value is two words, then assume they are first name and
             # last name
             parts = data['name'].split()
-            if len(parts) == 2:
-                condition = Q(first_name__istartswith=parts[0]) \
-                    | Q(last_name__istartswith=parts[1])
-            else:
-                condition = Q(first_name__istartswith=data['name']) \
-                    | Q(last_name__istartswith=data['name'])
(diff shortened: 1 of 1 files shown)


kkx:
Set the initial filter condition to True(will filter all results), then we can use something like “condition &= new_condition” in a for loop.

src/oscar/apps/dashboard/users/views.py
```diff
@@ -82,12 +82,11 @@ def apply_search_filters(self, queryset, data):
...
             # last name
             parts = data['name'].split()
-            if len(parts) == 2:
-                condition = Q(first_name__istartswith=parts[0]) \
-                    | Q(last_name__istartswith=parts[1])
-            else:
-                condition = Q(first_name__istartswith=data['name']) \
-                    | Q(last_name__istartswith=data['name'])
+            # always true filter
+            condition = Q(pk__isnull=False)
```
kkx: Set the initial filter condition to True(like an empty filter), then we can use something like “condition &= new_condition” in a for loop to combine more conditions(something like a>1 and b<2 and ...)

sasha0: Initial condition could be rather just `Q()`, than this one.

kkx:
@sasha0 Thanks for pointing it out!

sasha0:
Thanks for the contribution.

PR #2121, opened 2016-09-15
Updated code for Django>=1.8

sudhirmishra:
A new setting was introduced in Django 1.8: TEMPLATES. All existing template-related settings were deprecated. The documentation is outdated for customising templates.

[Upgrading templates to Django 1.8](https://docs.djangoproject.com/en/1.10/ref/templates/upgrading/)


sasha0:
Thanks for the contribution!

Could you please squash commits into one? Thanks.

PR #2125, opened 2016-09-29
Update getting_started.rst

richlloydmiles:
## Scope

As per https://docs.djangoproject.com/en/1.10/topics/http/middleware/ MIDDLEWARE_CLASSES has changed to MIDDLEWARE.

Oscar still uses MIDDLEWARE_CLASSES and it should be noted to avoid confusion during setup.

Alternatively if this change isn't made it throws the following error when running `python manage.py rusnerver`

`VariableDoesNotExist: Failed lookup for key [basket] in u"<WSGIRequest: GET '/'>"`


sasha0:
Thanks for the contribution.

Issue #2127, opened 2016-10-01
Product search returns products that contain any *char* from the query


anentropic:
If I search for `red dress` I get any products with `r` or `e` or `d` or `s` in them.

I am using Elasticsearch backend.

Can anyone provide any suggestion of how to begin solving this?

I'm not even sure right now what to google for

anentropic:
Is it the tokeniser?  I don't want to use exact matching, I want `red AND dress`

anentropic:
yeah I think this is because of use of `EdgeNgramField` in `ProductIndex`

I assume that was needed to make auto-complete work but it stuffs up the normal search results

so I think these features should use separate fields

at the moment both `title` and `text` are using `EdgeNgramField` ...I'm not sure where autocomplete is used but maybe it's enough to just have the ngram on `title`? Then the document field `text` could just be a `CharField` so as to return sensible results from search.

I have done this successfully in my own oscar project, overriding the field on my `ProductIndex`, and it works as desired... does the current default setup make any sense?

solarissmoke:
Fixed via #2128.

PR #2128, opened 2016-10-01
don't use ngram for main search field, returns irrelevant results

anentropic:
see issue #2127 

I assume `EdgeNgramField` was chosen to support autocomplete type of functionality.  I could not find where that was actually implemented anywhere though. In the meantime, using ngram field here causes lots of irrelevant results to appear in search.  So I think this is the right change.

If autocomplete is implemented I think it maybe makes sense to autocomplete on product titles only (rather than words from description) and the `title` field is still defined as another `EdgeNgramField` so that would work.


solarissmoke:
Strikes me that we should mention this in the release notes. Will add a task to do that.

Issue #1778, opened 2015-06-12
Implement hash navigation in dashboard


jakobholmelund:
It's really annoying that you can't navigate the dashboard properly. Every time you hit back, you are reset to the default tab position.

For example:

I'm looking at an order, i go to the shipping tap, i press edit, i regret for a moment and press back/cancel/breadcrumb. When i get back the Order contents tap is now active

This is just one example, variant editing comes to mind as another pain point in this regard

john-parton:
This would be nice to have. I have a hack that adds some history in, but specifically only for variant editing.

maikhoepfel:
I totally agree. We're using Bootstrap's tabs, I'm not sure if they support it? Any way, a PR is most welcome!

bmihelac:
Added PR fot this, let me know if needs any change.

sasha0:
Fixed by f6d16fc53e3c0d8012890bda04e1267957982b04.

PR #2136, opened 2016-10-13
Fixes Implement hash navigation in dashboard #1778

bmihelac:
This is my first encounter with django-oscar codebase, so it is possible that this should go to some .js file rather than html.

Code is adapted from http://stackoverflow.com/a/9393768/238628


mvantellingen:
Thanks! looks good

PR #2137, opened 2016-10-14
Update docs to reflect using Django Migrations

bmihelac:
instead of South migrations


sasha0:
Thanks for the contribution!

PR #2138, opened 2016-10-16
Fix sending direct messages.

michaelkuty:
Use same logic like has send_user_messages.

### DIFF
--- a/src/oscar/apps/customer/utils.py
+++ b/src/oscar/apps/customer/utils.py
@@ -25,7 +25,7 @@ class Dispatcher(object):
         """
         Dispatch one-off messages to explicitly specified recipient(s).
         """
-        if messages['subject'] and messages['body']:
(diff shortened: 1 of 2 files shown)


sasha0:
Could you please refer to related issue? Thanks!

michaelkuty:
Hi, thanks for checking PR, i havent any issue yet, but this fix sending messages when user does not exists, i use same behavior as is on line https://github.com/michaelkuty/django-oscar/blob/e6021815dbf38e2da1cc399a6267dc131dd10c72/src/oscar/apps/customer/utils.py#L55

When we has same templates/messages for non registered users then this method skips the message without any warning or exceptions

sasha0:
Alright, could you please add test for this? Thanks.

michaelkuty:
Ok working tests added.

sasha0:
Thanks for the contribution!

Issue #1872, opened 2015-10-16
Search results sorting returns results in wrong order


bharling:
I'm having lots of problems with all types of sorting of search results. First off I'm using the haystack backend. Sorting by price gives me results ordered by the first digit in the price eg: £7.30 consistently comes before £12.30 in a descending search. Also title search seems to be reversed. I'm using oscar 1.2. 

We have a custom Strategy and pricing policy in place, but they differ very little from the default implementation. My search indexes are up to date. 

Has anyone else hit this ?

bharling:
Just finished setting up SOLR in my dev environment and sorting magically working fine. I think the docs could be clearer that actually SOLR is the only search engine of any use with oscar ?

john-parton:
What backend were you using first? Whoosh?

Generally, elasticsearch is "as good" as solr.

bharling:
I had tried whoosh and elasticsearch, both with the same result when it came to ordering of search results. Solr is working great for me at the moment, but is quite a lot more painful to configure than the other options ( spent a good portion of today configuring staging and production environments with it :( )

I guess this could be closed really, but I'd still suggest that the docs could be clearer that you're going to get a lot better results with solr than other options.

digicase:
I tried elasticsearch first of all and had trouble with the search facets - they basically didn't work at all. As soon as I switched to solr the faceting worked. So I don't think elasticsearch can be considered "as good" as solr when it comes to Oscar support. It would be nice if elasticsearch was better supported by Oscar, as I found the same as @bharling that solr was harder to set up.

john-parton:
Elasticsearch works fine. It's just that Oscar and Haystack "out of the box" don't really take full advantage of it.

I use elastic as my backend, but I had to [almost] completely replace the search app. I used [elasticsearch-dsl-py](https://github.com/elastic/elasticsearch-dsl-py) to actually generate the elastic queries, and a bunch of custom classes to glue it all together. More info on the google group: https://groups.google.com/forum/#!msg/django-oscar/Sl3_jq4KBBc/QPfFgEY_BCYJ

You can try it out [here](https://pillowcentral.com/search/) if you would like.

Mohl:
This is a known issue for almost 5 years now - https://github.com/django-haystack/django-haystack/issues/398

digicase:
Thanks @Mohl that worked for me with Solr 4.10.4 and Haystack 2.4.1

[...]

PR #2142, opened 2016-10-21
- fixed sorting by title

vaskinyy:
Sorting product by title doesn't work with Elasticsearch backend, because Elasticearch doesn't sort ngram fields correctly.
This request adds unanalyzed title CharField that is sorted correctly by all backends.

p.s. partially addresses #1872


sasha0:
Thanks for the contribution.

PR #2146, opened 2016-10-25
Feature/django 110

dszmaj:
some errors in testing, but doesnt seem to be django 1.10 related, sandbox on 1.10 works good


dszmaj:
oh, and found this

MyModel._meta.get_fields_with_model() becomes:
```python
[
    (f, f.model if f.model != MyModel else None)
    for f in MyModel._meta.get_fields()
    if not f.is_relation
        or f.one_to_one
        or (f.many_to_one and f.related_model)
]
```

SalahAdDin:
What's about this?

dszmaj:
nothing, i didn't see this earlier, i will take out all errors later so you could release and i could upgrade to 1.10 for needed features

mvantellingen:
Added 1.10 via another PR in master. Thanks for the PR though, really appreciated.

PR #2150, opened 2016-11-03
- added documentation for facet sorting

vaskinyy:
django-oscar provides an easy way of sorting/filtering faceted items, you just pass sorting/filtering parameter through 'options'. 
Yet, the documentation says almost nothing about it, so it took a while for me to understand it. By this documentation improvement I want to make it a bit more clear how to do facet sorting/filtering.


sasha0:
Thanks for the contribution.

PR #2159, opened 2016-11-07
Use class name as unique identifier for custom benefits and conditions.

michaelkuty:


src/oscar/apps/offer/custom.py
```diff
@@ -38,8 +38,10 @@ def create_condition(condition_class, **kwargs):
     """
     Create a custom condition instance
     """
-    return Condition.objects.create(
-        proxy_class=_class_path(condition_class), **kwargs)
+    created, condition = Condition.objects.create(
+        proxy_class=_class_path(condition_class), defaults=kwargs)
```
mvantellingen: ``Model.objects.create()`` doesn't return a tuple i think?

sasha0: No it doesn't. Maybe it was meant to use `Condition.objects.get_or_create()`.

michaelkuty:
Yes, you are right

michaelkuty:
You cannot call these metods when app starts because you must check duplicities, from my point of view this is better for custom offers use cases, but agree that change is not important for all users

Issue #2096, opened 2016-08-13
Apache Solr integration documentation outdated


SalahAdDin:
Hi guys

As[ i couldn't integrate](https://github.com/django-oscar/django-oscar/issues/2089) Oscar with `elasticsearch` search engine, i had chosen use `apache solr` instead.

```
 I test it in the same way that i test `elasticsearch`: seeking in the home page all products, and i have this bug:
[13/Aug/2016 01:43:49] "GET / HTTP/1.1" 200 28031
Failed to query Solr using '*:*': Solr responded with an error (HTTP 404): [Reason: Error 404 Not Found]
Traceback (most recent call last):
  File "/home/salahaddin/Proyectos/demo-oscar/lib/python3.5/site-packages/haystack/backends/solr_backend.py", line 134, in search
    raw_results = self.conn.search(query_string, **search_kwargs)
  File "/home/salahaddin/Proyectos/demo-oscar/lib/python3.5/site-packages/pysolr.py", line 710, in search
    response = self._select(params, handler=search_handler)
  File "/home/salahaddin/Proyectos/demo-oscar/lib/python3.5/site-packages/pysolr.py", line 411, in _select
    return self._send_request('get', path)
  File "/home/salahaddin/Proyectos/demo-oscar/lib/python3.5/site-packages/pysolr.py", line 386, in _send_request
    raise SolrError(error_message % (resp.status_code, solr_message))
pysolr.SolrError: Solr responded with an error (HTTP 404): [Reason: Error 404 Not Found]
Failed to query Solr using '*:*': Solr responded with an error (HTTP 404): [Reason: Error 404 Not Found]
Traceback (most recent call last):
  File "/home/salahaddin/Proyectos/demo-oscar/lib/python3.5/site-packages/haystack/backends/solr_backend.py", line 134, in search
    raw_results = self.conn.search(query_string, **search_kwargs)
  File "/home/salahaddin/Proyectos/demo-oscar/lib/python3.5/site-packages/pysolr.py", line 710, in search
    response = self._select(params, handler=search_handler)
  File "/home/salahaddin/Proyectos/demo-oscar/lib/python3.5/site-packages/pysolr.py", line 411, in _select
    return self._send_request('get', path)
  File "/home/salahaddin/Proyectos/demo-oscar/lib/python3.5/site-packages/pysolr.py", line 386, in _send_request
    raise SolrError(error_message % (resp.status_code, solr_message))
[...]

PR #2171, opened 2016-11-17
Update How to setup Solr with Oscar (#2096)

bmihelac:
Docs are updated to create schema.xml instead of reckon on config that no longer exists.


SalahAdDin:
But, are you sure about delete it?

sasha0:
@bmihelac thanks for the contribution!

PR #2185, opened 2016-12-14
Add 'oscar/' prefix when extending base.html or layout.html.

### LINKED ISSUE #1378: Rework template include path to avoid clashes
The way Oscar handles templates can clash with other apps. For instance, when you use the `OSCAR_MAIN_TEMPLATE_DIR` folder in `TEMPLATE_DIRS` (as you need to do to install Oscar), then templates like `base.html` and `layout.html` will resolve to Oscar's template folder.  Buy other apps often define a `base.html` (eg Mezzanine) which will clash.

Oscar uses this additional template path to allow templates to be customised without forking the whole template.

I think we can work around this issue by changing how we use template roots. We should:
- Change `OSCAR_MAIN_TEMPLATE_DIR` to point to the `oscar` folder itself so the `base.html` template can be resolved against paths `oscar/base.html` and `templates/oscar/base.html` (but no longer `base.html`).
- Change all of Oscar's templates to extend the 'oscar'-namespaced version (eg `{% extends 'oscar/layout.html' %}` instead of `{% extends 'layout.html' %}`.

This will allow projects to customise Oscar's templates by extending the corresponding template but using the longer path:

``` html
# myproject/templates/oscar/base.html

{% extends 'templates/oscar/base.html' %}

{% block content %}

{% endblock %}
```

Previously, this customising template would have been located in `myproject/templates/base.html` and would have extended `oscar/base.html` and so this change will be backwards incompatible for projects that customise Oscar's templates.

kamzil:
- {% extends "base.html" %} -> {% extends "oscar/base.html" %}
- {% extends "layout.html" %} -> {% extends "oscar/layout.html" %}

This allows having independent base.html and layout.html in template root.

Other templates remain untouched, and e.g. creating template dashboard/layout.html will still override oscar/dashboard/layout.html. This might still cause problems if the project has another app named dashboard, for example.

Fixes #1378

### DIFF
--- a/src/oscar/templates/oscar/basket/basket.html
+++ b/src/oscar/templates/oscar/basket/basket.html
@@ -1,4 +1,4 @@
-{% extends "layout.html" %}
+{% extends "oscar/layout.html" %}
 {% load thumbnail %}
 {% load i18n %}
 
--- a/src/oscar/templates/oscar/catalogue/detail.html
+++ b/src/oscar/templates/oscar/catalogue/detail.html
@@ -1,4 +1,4 @@
-{% extends "layout.html" %}
+{% extends "oscar/layout.html" %}
 
 {% load history_tags %}
 {% load currency_filters %}
--- a/src/oscar/templates/oscar/catalogue/reviews/review_detail.html
+++ b/src/oscar/templates/oscar/catalogue/reviews/review_detail.html
@@ -1,4 +1,4 @@
-{% extends "layout.html" %}
+{% extends "oscar/layout.html" %}
 {% load i18n %}
 
 {% block title %}
--- a/src/oscar/templates/oscar/catalogue/reviews/review_list.html
+++ b/src/oscar/templates/oscar/catalogue/reviews/review_list.html
@@ -1,4 +1,4 @@
-{% extends "layout.html" %}
+{% extends "oscar/layout.html" %}
 
 {% load history_tags %}
 {% load i18n %}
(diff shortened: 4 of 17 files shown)


djkartsa:
As mentioned in the commit message, this does not solve the problem if there is e.g. an unrelated app named dashboard that has layout.html or base.html and the templates are located in templatefolder/dashboard.

This is because if I understood correctly, the recommended way to customize Oscar is to have this kind of structure and prepending "oscar/" in front of every template reference could complicate things.

Please correct me if I'm wrong.

solarissmoke:
Thanks for this PR. I've had a good long look at how the template resolution currently works and think we need to do a bit more work here, basically along the lines of what is proposed in #1378. My view is that:

1. Oscar internally has to be consistent about how it loads templates. Currently with this PR some templates are extended with an `oscar/` prefix, but others are not. This is problematic because different template loaders will be used in template resolution (anything with the `oscar/` prefix will go through the Filesystem loader; anything without will go through the AppDirectories loader). This inconsistency will cause all sorts of confusion with projects that have apps that collide with any Oscar templates.

2. Oscar should *not* hog the global template namespace, and should be prefixing everything with `oscar/`.

On this basis I think the best approach is to do pretty much was #1378 says:

1. Change *all* Oscar templates to extend `oscar/` prefixed templates.

2. Change `OSCAR_MAIN_TEMPLATE_DIR` to point to `src/oscar` instead of `src/oscar/templates/oscar`. 

This is a backwards-incompatible change, but we probably just have to make it at some point.

Input from @mvantellingen would be useful here as well.

sasha0:
I personally agree with the approach and since new dashboard layout is in progress in PR #2302 it might be right time to include this one too.

sasha0:
Per the latest discussion in Slack, PR has partials changes and not in the state we can merge it "as is". Since it's backwards incompatible change, it will make much more sense to introduce it within new dashboard layout, as noted above.

PR #2187, opened 2016-12-26
Fix facet form scripting

okfish:
This allows to make facet form more flexible for css-styling (as our UX guys has a real headashes styling it) and also i could not find a way to partially extend/override this UI script :)


wa3x:
the default behavior (and decor) does not change, and (for example) awesome-bootstrap-checkbox work perfect too.

mvantellingen:
thanks

Issue #2183, opened 2016-12-05
python manage.py oscar_populate_countries throws error -


eve1234:
Hello,

while running the following command I get this error -

`./manage.py oscar_populate_countries`

or

`python manage.py oscar_populate_countries -
`

This' the log.

```
./manage.py oscar_populate_countries
/Users/akos/oscar/lib/python2.7/site-packages/django/utils/six.py:808: RemovedInDjango110Warning: SubfieldBase has been deprecated. Use Field.from_db_value instead.
  return meta(name, bases, d)

/Users/akos/oscar/lib/python2.7/site-packages/django/core/management/__init__.py:345: RemovedInDjango110Warning: OptionParser usage for Django management commands is deprecated, use ArgumentParser instead
  self.fetch_command(subcommand).run_from_argv(self.argv)

Traceback (most recent call last):
  File "./manage.py", line 10, in <module>
    execute_from_command_line(sys.argv)
  File "/Users/akos/oscar/lib/python2.7/site-packages/django/core/management/__init__.py", line 353, in execute_from_command_line
    utility.execute()
  File "/Users/akos/oscar/lib/python2.7/site-packages/django/core/management/__init__.py", line 345, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/Users/akos/oscar/lib/python2.7/site-packages/django/core/management/base.py", line 348, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/Users/akos/oscar/lib/python2.7/site-packages/django/core/management/base.py", line 399, in execute
    output = self.handle(*args, **options)
  File "/Users/akos/oscar/lib/python2.7/site-packages/oscar/management/commands/oscar_populate_countries.py", line 58, in handle
    for country in pycountry.countries]
  File "/Users/akos/oscar/lib/python2.7/site-packages/pycountry/db.py", line 22, in __getattr__
    raise AttributeError
AttributeError


```

I have been google to find the answer. And evening ended up creating a whole new project again. What Im I doing wrong?

Regards

adlh:
Same for me on a 100% fresh installation. 

These are my installed packages in the venv:
```
$ pip freeze
Babel==2.3.4
Django==1.9
django-extra-views==0.6.4
django-haystack==2.5.1
django-oscar==1.3
django-tables2==1.0.7
django-treebeard==4.1.0
django-widget-tweaks==1.4.1
factory-boy==2.8.1
Faker==0.7.5
mock==2.0.0
[...]

PR #2198, opened 2017-01-08
Support pycountry >= 16.10.23rc1

tobiase:
Fix issue #2183 
Since pycountry >= 16.10.23rc1: 

> Note that "alpha2", "alpha4", etc. are now using an underscore as that's the pattern in the upstream packages. So it's "alpha_2" now.
[Source](https://bitbucket.org/flyingcircus/pycountry/src/4b3a8c9ca4604270b4aa6d92f7b53bfd1eb7454d/HISTORY.txt?at=default&fileviewer=file-view-default)


tobiase:
I created a new pull request without support for pycountry pycountry < 16.10.23rc1

PR #2199, opened 2017-01-08
Support pycountry >= 16.10.23rc1

tobiase:
Fix issue #2183 
Without support for pycountry < 16.10.23rc1
I couldn't think of any good reason for supporting older pycountry versions.


mvantellingen:
thanks!

PR #2211, opened 2017-01-26
Add CACHES Redis as alternative

alexfromvl:


mvantellingen:
It now uses django-environ, see https://github.com/django-oscar/django-oscar/blob/master/sandbox/settings.py#L43

PR #2212, opened 2017-01-28
added 127.0.0.1 to ALLOWED_HOSTS to avoid initial startup error

flycal6:


solarissmoke:
Thanks for this - the project structure changed since your PR so have applied the fix in the new settings file.

PR #2157, opened 2016-11-05
Upgrade django_tables2, fix dashboard views to support it.

sasha0:
Fixes #2084

PR #2222, opened 2017-02-05
Fix various deprecation warnings

mvantellingen:
Would be good to get this in before 1.4.

It fixes the following (1500) deprecation warnings:
- RemovedInDjango20Warning: Using user.is_authenticated() and user.is_anonymous() as a method is deprecated. Remove the parentheses to use it as an attribute.
 - ResourceWarning: unclosed file ..
 -  Direct assignment to the forward side of a many-to-many set is deprecated due to the implicit save() that happens. Use offers.set() instead.

It also ignores warnings in django-webtest and sorl, see setup.cfg

The last few should be fixed by #2157 (django-tables2)

### DIFF
--- a/src/oscar/apps/offer/abstract_models.py
+++ b/src/oscar/apps/offer/abstract_models.py
@@ -1120,7 +1120,7 @@ class AbstractRangeProductFileUpload(models.Model):
         Extract all SKU- or UPC-like strings from the file
         """
         with open(self.filepath, 'r') as fh:
-            for line in fh.readlines():
+            for line in fh:
                 for id in re.split('[^\w:\.-]', line):
                     if id:
                         yield id
--- a/src/oscar/apps/offer/abstract_models.py
+++ b/src/oscar/apps/offer/abstract_models.py
@@ -1119,10 +1119,11 @@ class AbstractRangeProductFileUpload(models.Model):
         """
         Extract all SKU- or UPC-like strings from the file
         """
-        for line in open(self.filepath, 'r'):
-            for id in re.split('[^\w:\.-]', line):
-                if id:
-                    yield id
+        with open(self.filepath, 'r') as fh:
+            for line in fh.readlines():
+                for id in re.split('[^\w:\.-]', line):
+                    if id:
+                        yield id
 
     def delete_file(self):
         os.unlink(self.filepath)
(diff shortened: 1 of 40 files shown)


src/oscar/core/compat.py
```diff
@@ -77,6 +77,27 @@ class Meta:
     MiddlewareMixin = object
 
 
+def user_is_authenticated(user):
+    if django.VERSION >= (1, 10):
+        return user.is_authenticated
+    else:
+        return user_is_authenticated(user)
```
solarissmoke: Shouldn't this be `return user.is_authenticated()`?

mvantellingen: Yeah, search-and-replace gone wrong. Thanks :-)

src/oscar/apps/offer/abstract_models.py
```diff
@@ -1119,10 +1119,11 @@ def extract_ids(self):
         """
         Extract all SKU- or UPC-like strings from the file
         """
-        for line in open(self.filepath, 'r'):
-            for id in re.split('[^\w:\.-]', line):
-                if id:
-                    yield id
+        with open(self.filepath, 'r') as fh:
+            for line in fh.readlines():
```
solarissmoke: `readlines()` loads the entire contents of the file into memory. I think it's better to just loop over the file object itself: `for line in fh` which is memory-efficient.

tests/settings.py
```diff
@@ -68,15 +69,29 @@
...
+
+        'oscar.apps.basket.middleware.BasketMiddleware',
+    ]
+else:
+    MIDDLEWARE = [
+        'django.middleware.common.CommonMiddleware',
+        'django.contrib.sessions.middleware.SessionMiddleware',
+        'django.middleware.csrf.CsrfViewMiddleware',
+        'django.contrib.auth.middleware.AuthenticationMiddleware',
+        'django.contrib.auth.middleware.SessionAuthenticationMiddleware',
```
solarissmoke: `SessionAuthenticationMiddleware` [no longer does anything](https://github.com/django/django/blob/1.10.5/django/contrib/auth/middleware.py#L27) and is removed in Django 2.0 - can remove it from here.

src/oscar/apps/offer/abstract_models.py
mvantellingen: Fixed

tests/settings.py
mvantellingen: Removed, thanks

PR #2236, opened 2017-02-13
Remove break lines from some templates.

jklapuch:


mvantellingen:
First of all thanks for the PR. I'm wondering what the reason for this is. If we merge this then all translations for that block will be lost, so we can't take this too lightly.

jklapuch:
Oh, sorry, I have forgotten the translations.  I have two shipping methods and one offer for free shipping. Without modifications my results is

image

PR #2241, opened 2017-02-17
Fixed string formatting bug in VoucherRemoveView

clintonb:
The pk kwarg on views is a string type. The message format string was expecting a numeric value. This updates that format string to expect a string value, which will always display correctly.


solarissmoke:
Thanks!

Issue #1811, opened 2015-07-19
Is there a race condition when allocating stock?


anentropic:
there are several methods in the `partner.StockRecord` model that look a bit like this:

```
def allocate(self, amount, reference='', status=''):
        """
        Convenience method for ring-fencing money against this source
        """
        self.amount_allocated += amount
        self.save()
```

I didn't dig too deep but I didn't find any use of `select_for_update` in the source so I think even knowing that everything's wrapped in a per-request transaction there's a possible race condition here?

Should the methods do something like this to be safer?

```
def allocate(self, amount, reference='', status=''):
        """
        Convenience method for ring-fencing money against this source
        """
        self.__class__.objects.select_for_update().filter(pk=self.pk).update(
            amount_allocated=F('amount_allocated') + amount)
```

john-parton:
As far as I can tell, `select_for_update` isn't used any where in the django-oscar code base at all.

This one might be better solved with a `transaction.atomic` decorator, though.

```python
@transaction.atomic
def allocate(self, amount, reference='', status=''):
        """
        Convenience method for ring-fencing money against this source
        """
        self.amount_allocated += amount
        self.save()
```

Could you please write a test case where the current implementation fails due to a race condition?

anentropic:
`atomic` wouldn't solve the potential race condition in your example, as the instance has already been loaded from the db, so the value of `self.amount_allocated` can't be trusted any more.

even if the instance was loaded from the db _within_ the atomic transaction I don't think you can trust it as the transaction does not preclude another user session from changing the value behind your back (assuming default isolation level https://docs.djangoproject.com/en/1.8/ref/databases/#isolation-level http://www.postgresql.org/docs/current/static/transaction-iso.html#XACT-READ-COMMITTED) 

this is why `select_for_update` exists - i.e. in the case where you `get()` the row then later `save()` it, the db would ensure that row is locked

however _I think_ I have gone too far in my example and `select_for_update` is not necessary when doing an operation like `.filter(pk=self.pk).update(amount_allocated=F('amount_allocated') + amount)` since this accomplishes the update within a single query, so it's enough to just be within an atomic transaction

anentropic:
either way, I agree a test case should be made :)

john-parton:
Excellent info on `select_for_update`. 

Looking at it, you might be right that the `select_for_update` is unnecessary, because it doesn't actually read anything, and just does a single `UPDATE` directly.

[...]

PR #2246, opened 2017-02-20
Concurrent checkout.

sasha0:
Refs #1811

### DIFF
--- a/tests/functional/checkout/test_concurrent_checkout.py
+++ b/tests/functional/checkout/test_concurrent_checkout.py
@@ -1,0 +1,31 @@
+from threading import Thread
+
+from oscar.core.loading import get_model, get_class
+from oscar.test.testcases import WebTestCase
+from . import CheckoutMixin
+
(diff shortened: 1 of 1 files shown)


tests/functional/checkout/test_concurrent_checkout.py
```diff
@@ -0,0 +1,25 @@
...
+class TestConcurrectCheckout(CheckoutMixin, WebTestCase):
+
+    def do_checkout(self):
+        self.add_product_to_basket()
+        self.enter_shipping_address()
+        self.place_order()
+
+    def test_num_allocated(self):
+        t = Thread(target=self.do_checkout)
+        t.start()
```
mvantellingen: We should join the threads in a finally clause i think ?

sasha0: Yep.

mvantellingen:
Closing as it is now a duplicate of #2281

PR #2256, opened 2017-02-26
passing request object to place order view

sasha0:


src/oscar/apps/checkout/mixins.py
```diff
@@ -146,7 +146,9 @@ def place_order(self, order_number, user, basket, shipping_address,
             shipping_charge=shipping_charge,
             total=order_total,
             billing_address=billing_address,
-            status=status, **kwargs)
+            status=status,
+            request=getattr(self, 'request', None),
```
solarissmoke: This will break things is someone has passed a `request` to `kwargs`? Need to pop it from `kwargs` if it exists.

sasha0: Indeed it would. Thanks for the find.

sasha0: Fixed.

solarissmoke:
Looks good to me.

PR #2264, opened 2017-03-13
Update context processors in getting started docs.

jklapuch:
Load i18n context processor from django.template instead of django.core as fix for Django 1.10


SalahAdDin:
@sasha0 This is very important!

sasha0:
@jklapuch thanks for the contribution.

PR #2268, opened 2017-03-21
Add 'isort:skip' to model imports in forked apps

skirsdeda:
Since `from xxx.models import *` must always be after custom model definitions in forked apps, isort should skip that line.


sasha0:
Thanks for the contribution.

PR #2271, opened 2017-03-23
Add note to the customization doc about Oscar applications order.

sasha0:


docs/source/topics/customisation.rst
```diff
@@ -78,6 +78,27 @@ list of additional apps, they will be used to replace the Oscar core apps.
 In the above example, ``yourproject.order`` will be returned instead of
 ``oscar.apps.order``.
 
+.. note::
+
+    Dashboard applications should follow core applications (basket, catalogue etc),
```
solarissmoke: Suggest a slight tweak to the wording:

> Overrides to dashboard applications should be listed after any overrides to core applications (`basket`, `catalogue`, etc.), as they depend on models declared in the core applications. Declaring them before core overrides will cause problems with Oscar's dynamic model loading.

sasha0: Thanks for the review and suggestion.

Issue #2040, opened 2016-04-26
MultipleObjectsReturned is possible when logging in with duplicate email


bufke:
Related to #295 but that was marked fixed so maybe this is a regression. Same issue as before. Can be reproduced by
1. Create 2 users with the same email in django admin
2. Log in using /store/accounts/login/

```
MultipleObjectsReturned: There are multiple users with the given email address and password
  File "django/views/generic/base.py", line 89, in dispatch
    return handler(request, *args, **kwargs)
  File "oscar/apps/customer/views.py", line 125, in post
    return self.validate_login_form()
  File "oscar/apps/customer/views.py", line 152, in validate_login_form
    if form.is_valid():
  File "django/forms/forms.py", line 184, in is_valid
    return self.is_bound and not self.errors
  File "django/forms/forms.py", line 176, in errors
    self.full_clean()
  File "django/forms/forms.py", line 393, in full_clean
    self._clean_form()
  File "django/forms/forms.py", line 417, in _clean_form
    cleaned_data = self.clean()
  File "django/contrib/auth/forms.py", line 157, in clean
    password=password)
  File "django/contrib/auth/__init__.py", line 74, in authenticate
    user = backend.authenticate(**credentials)
  File "oscar/apps/customer/auth_backends.py", line 54, in authenticate
    "There are multiple users with the given email address and "
```

maerteijn:
I can confirm this one

bufke:
I ended up working around this with some monkey patching auth.User to make the email field unique and force lower caps. The only clean way I could think to do it would be replacing the user model outright - which is a ton of code for a very small change. 

Not sure the best solution here. Minimally Oscar could document the issue and suggest some work-arounds.

mvantellingen:
PR welcome

sasha0:
I neither don't see so far better option as to document this. Enforcing custom user model is not backwards compatible with existing Oscar installations. Besides, using generic Oscar user creation flow it's not available to create multiple users with the same email address.

sasha0:
The only other option I have in mind is to override user admin with the custom model form which will enforce email address uniqueness. WDYT @solarissmoke @mvantellingen ?

solarissmoke:
I've added to the documentation in #2354 to state that Oscar assumes a unique `email` field. Not sure whether it is feasible to do anything to fix this at the code level - the assumption of uniquess is baked pretty hard into Oscar.

sasha0:
Agree. Closing.

PR #2272, opened 2017-03-26
Fixed race condition handling logic in RegisterUserMixin

solarissmoke:
A user must be authenticated (or have a backend set) before you can log them in with `auth.login()`. Currently in `RegisterUserMixin`'s race condition handling logic, this doesn't happen, which means you end up with another exception and the point of catching the first exception is lost.

Changed this to re-authenticate the user after cleaning up the duplicates.

### DIFF
--- a/src/oscar/apps/customer/mixins.py
+++ b/src/oscar/apps/customer/mixins.py
@@ -74,10 +74,14 @@ class RegisterUserMixin(object):
             # ignore capitalisation when looking up an email address.
             # We might otherwise accidentally mark unrelated users as inactive
             users = User.objects.filter(email=user.email)
-            user = users[0]
             for u in users[1:]:
                 u.is_active = False
                 u.save()
+                
+            # Try again to authenticate
+            user = authenticate(
+                username=user.email,
+                password=form.cleaned_data['password1'])
 
         auth_login(self.request, user)
 


solarissmoke:
@sasha0 I've tried... but it is incredibly difficult to write a test that lets me override the `AUTH_USER_MODEL` (which is what I need to do in order to test this - Oscar's custom user model enforces unique email address which means the case I'm trying to test would never arise as the DB would not allow two users with the same email address).

For some reason applying `@override_settings(AUTH_USER_MODEL='auth.user')` doesn't work - the test database is missing the `auth_user` table.

SalahAdDin:
Wow, isn't works.

sasha0:
@SalahAdDin please stop leaving irrelevant comments. Thank you.

@solarissmoke did you review Django tests for custom user model?

solarissmoke:
@sasha0, yes, but I didn't find anything there to help. 

Let me try and explain the problem.  The whole Oscar test suite is [initialised with Oscar's custom user model](https://github.com/django-oscar/django-oscar/blob/master/tests/settings.py#L37). This bit of code needs to use Django's core model. I have tried doing this:

```python
@override_settings(AUTH_USER_MODEL='auth.user')
class CustomerMixinTestCase(TestCase):

    def test_race_condition_handling(self):
        User = ...
```

This doesn't work. The problem is that the test database is still initialised with the original (`myauth.auth` user model). When you then try to use the overridden model here, you get an error:

`OperationalError: no such table: auth_user`

Because the table that was created was `myauth_user`.

Any suggestions?

sasha0:
Why custom test user model is not suitable? Since it's inherited from abstract user model.

solarissmoke:
The custom test model enforces `unique=True` on `email`, which means you get an integrity error when you attempt to create users with the same email address - i.e, there is no way to create the race condition.

Actually the whole logic here is broken and doesn't work as intended. Even in a model where `email` is unique you will get a 500 error (due to database integrity error) when this bit of code is triggered. I think more substantial refactoring is required to properly fix this.

sasha0:
Is this related with #2040? Since major reworking required - I'd suggest to open new PR when it's finished and close this one.

Issue #2101, opened 2016-08-17
Vouchers can be reused under concurrent checkout


TWarszawski:
When two customers check out concurrently using a voucher that should only be able to be used once by one customer, the voucher can be used twice.

Steps to reproduce:
1.  Start sandbox, create two customers, create/pick test product, create test voucher with usage setting “can be used once by one customer”.
2.  Both customers add the item to their baskets, apply the voucher, and perform the  checkout concurrently, making sure both customers finish checkout (click the Place Order button) as close to the same time as possible.

We have reproduced this behavior on a single machine, by performing the above steps in rapid succession with one customer in one browser window and another customer in a second browser window.

Expected Result:
Only one checkout successfully completes using the voucher.

Actual Result:
Both checkouts successfully complete using the voucher.

sasha0:
Fixed in 8fb5ce932beeffc3df1482dce9198a0978a6855b

PR #2281, opened 2017-03-31
Race conditions

mvantellingen:
These changes fix two race conditions:
 - One in the stock allocation process (see #1811)
 - One when using single-usage vouchers (#2101)

### DIFF
--- a/src/oscar/apps/analytics/receivers.py
+++ b/src/oscar/apps/analytics/receivers.py
@@ -25,6 +25,8 @@ def _update_counter(model, field_name, filter_kwargs, increment=1):
     Efficiently updates a counter field by a given increment. Uses Django's
     update() call to fetch and update in one query.
 
+    TODO: This has a race condition, we should use UPSERT here
+
     :param model: The model class of the recording model
     :param field_name: The name of the field to update
     :param filter_kwargs: Parameters to the ORM's filter() function to get the
@@ -37,8 +39,8 @@ def _update_counter(model, field_name, filter_kwargs, increment=1):
             filter_kwargs[field_name] = increment
             model.objects.create(**filter_kwargs)
     except IntegrityError:
-        # get_or_create sometimes fails due to MySQL's weird transactions, fail
-        # silently
+        # get_or_create has a race condition (we should use upsert in supported)
+        # databases. For now just ignore these errors
         logger.error(
             "IntegrityError when updating analytics counter for %s", model)
 
--- a/src/oscar/apps/order/utils.py
+++ b/src/oscar/apps/order/utils.py
@@ -50,11 +50,8 @@ class OrderCreator(object):
             order_number = generator.order_number(basket)
         if not status and hasattr(settings, 'OSCAR_INITIAL_ORDER_STATUS'):
             status = getattr(settings, 'OSCAR_INITIAL_ORDER_STATUS')
-        try:
-            Order._default_manager.get(number=order_number)
-        except Order.DoesNotExist:
-            pass
-        else:
+
+        if Order._default_manager.filter(number=order_number).exists():
             raise ValueError(_("There is already an order with number %s")
                              % order_number)
 
--- a/src/oscar/apps/order/utils.py
+++ b/src/oscar/apps/order/utils.py
@@ -2,6 +2,7 @@ from decimal import Decimal as D
 
 from django.conf import settings
 from django.contrib.sites.models import Site
+from django.db import transaction
 from django.utils.translation import ugettext_lazy as _
 
 from oscar.core.compat import user_is_authenticated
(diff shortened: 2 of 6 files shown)


src/oscar/apps/analytics/receivers.py
```diff
@@ -37,8 +39,8 @@ def _update_counter(model, field_name, filter_kwargs, increment=1):
             filter_kwargs[field_name] = increment
             model.objects.create(**filter_kwargs)
     except IntegrityError:
-        # get_or_create sometimes fails due to MySQL's weird transactions, fail
-        # silently
+        # get_or_create has a race condition (we should use upsert in supported)
+        # databases. For now just ignore these errors
```
maikhoepfel: Ahaha. So that was the cause for that age-old comment.  

There's a misplaced `)`in the comment.

src/oscar/apps/partner/abstract_models.py
```diff
@@ -189,11 +191,27 @@ def allocate(self, quantity):
...
+        # Atomic update
+        (self.__class__.objects
+            .filter(pk=self.pk)
+            .update(num_allocated=(
+                Coalesce(F('num_allocated'), Value(0)) + quantity)))
+
+        # Make sure the current object is up-to-date
         if self.num_allocated is None:
             self.num_allocated = 0
         self.num_allocated += quantity
```
maikhoepfel: Shouldn't we just read `num_allocated` from the instance here? If there indeed was a race condition, `self.num_allocated` would be wrong?

src/oscar/apps/order/utils.py
```diff
@@ -50,44 +51,48 @@ def place_order(self, basket, total,  # noqa (too complex (12))
...
+                user, basket, shipping_address, shipping_method, shipping_charge,
+                billing_address, total, order_number, status, **kwargs)
+            for line in basket.all_lines():
+                self.create_line_models(order, line)
+                self.update_stock_records(line)
+
+            for voucher in basket.vouchers.select_for_update():
+                available_to_user, msg = voucher.is_available_to_user(user=user)
+                if not voucher.is_active() or not available_to_user:
+                    raise ValueError(msg)
```
mvantellingen: These lines were added, the rest was moved to the transaction.atomic().

src/oscar/apps/partner/abstract_models.py
```diff
@@ -189,11 +191,27 @@ def allocate(self, quantity):
...
-        self.save()
+
+        # Send the post-save signal
+        signals.post_save.send(
```
solarissmoke: Is it potentially confusing that `post_save` is sent, but not `pre_save`?

mvantellingen: Yeah good point, i'll add it

src/oscar/apps/analytics/receivers.py
mvantellingen: This sentence doesn't make much sense now indeed, i'll fix it :-)

src/oscar/apps/partner/abstract_models.py
mvantellingen: We explicitly want to reuse the current value in the database not on the instance because the instance value might be old . See https://docs.djangoproject.com/en/1.10/ref/models/expressions/#avoiding-race-conditions-using-f

tests/integration/order/test_creator.py
```diff
@@ -218,3 +224,96 @@ def test_multi_sites(self):
...
+        product = factories.ProductFactory()
+        voucher = factories.VoucherFactory(usage=Voucher.SINGLE_USE)
+        voucher.offers.add(factories.create_offer(offer_type='Voucher'))
+
+        # Make the order creator a bit more slow too reliable trigger
+        # concurrency issues
+        org_create_order_model = OrderCreator.create_order_model
+        def new_create_order_model(*args, **kwargs):
+            time.sleep(0.5)
+            return org_create_order_model(creator, *args, **kwargs)
```
sasha0: Not sure why do we need to make a delay.

src/oscar/apps/partner/abstract_models.py
maikhoepfel: Sorry if I wasn't clear. I think we mean the same thing. With your code, the database will have the correct value, but if there is a race condition, wouldn't `self.num_allocated` still contain the "faulty" quantity?

src/oscar/apps/analytics/receivers.py
anentropic: it might be possible to fallback to the `update` in the `except IntegrityError` block

that block only triggers when you try to `create` and lose the race... at this point you know the other record exists

could fail again if you are racing with a concurrent delete too - are deletes possible in this data model?

src/oscar/apps/partner/abstract_models.py
mvantellingen: Yup true, although I don't think we can reliably get the correct quantity though?

src/oscar/apps/partner/abstract_models.py
mvantellingen: Added

tests/integration/order/test_creator.py
mvantellingen: See the comments around it, The idea is to add a delay so that we can consistently trigger the race condition.

src/oscar/apps/analytics/receivers.py
mvantellingen: Let's open a new PR for this change as it's outside the scope of this PR

PR #1947, opened 2015-12-19
add javascript to enable client-side product image sorting in dashboard

anentropic:
I've been using this in my project, I think this is roughly the minimum needed to get it working

PR #2287, opened 2017-04-06
add javascript to enable client-side product image sorting in dashboard

sasha0:
Rebased version of PR #1947.


sasha0:
@anentropic could you please suggest what is the purpose of containers cloning in db27a62f78c48150f5fe435c07b3abaa0d642e6a? Thanks.

anentropic:
@sasha0 wow, it's a long time ago I wrote this, err...

I think basically the idea was the form always has an extra 'empty' image field (including various elements to display the thumbnail, not just the form field)

So when we add a new photo (or multiple photos) we clone that empty field, using it as a template, and add the new photos to the product.  Then if we want to add some more photos in a second operation, the empty template is still there for us, to clone again as needed.

sasha0:
@solarissmoke I re-generated migrations and fixed tests.

solarissmoke:
I'm seeing a few issues:

1. The `caption` and `reorder` fields are only available for existing images, not for new ones. Strangely, on master, the `caption` field doesn't appear at all, but here it appears for existing images.

2. For some reason, the `delete` checkbox does not appear on all the images - it is missing on some of the new ones.

The thumbnails are also failing to display for me in the edit view - I am also getting this in master. Not too sure what is going on there but something doesn't seem right.

nordbit:
For my need this is a important part:
remove:
```
unique_together = ("product", "display_order")
```
This is the real blocker if someone want make a custom sortable feature

sasha0:
Alright. I'll replace it to the separate PR.

sasha0:
@nordbit here you go #2338

sasha0:
> on master, the caption field doesn't appear at all, but here it appears for existing images.

Caption field was just hidden previously.

> For some reason, the delete checkbox does not appear on all the images - it is missing on some of the new ones.

Yes, it's to do with the JS code which copies the last element in the row.

> The thumbnails are also failing to display for me in the edit view - this happens on master as well. Are you getting the same?

Works fine for this changeset as well - re-checked on the brand new environment.

sasha0:
Finally finalized this functionality, had to rework extra form addition and decided not to introduce multiple files upload (through `multiple` attribute of the file input in the last extra form). Also made captions visible and eventually it looks like this:

Скриншот 2019-03-30 10 28 57

Would appreciate feedback on that @django-oscar/core.

solarissmoke:
Will try and review/test this next week

docs/source/releases/v2.0.rst
```diff
@@ -47,6 +47,9 @@ What's new in Oscar 2.0?
 - Added a ``filter_by_attributes`` method on the ``ProductQuerySet``,
   to allow database-level filtering of products by attribute.
 
+- Added images re-order to the dashboard product edit page, implement extra empty form addition to the images formset
```
solarissmoke: Grammar - suggest "Added support for re-ordering product images in the dashboard, and for adding an arbitrary number of additional images into the formset."

src/oscar/templates/oscar/dashboard/catalogue/product_update.html
```diff
@@ -172,6 +172,9 @@ <h3>{% trans "Upload, change or remove images" %}</h3>
                                                 {% include "dashboard/partials/product_images.html" with form=image_form %}
                                             {% endfor %}
                                         </ol>
+                                        <p class="help-block">
```
solarissmoke: I don't think this help text is necessary. The user will realise as soon as they add one image that the formset grows automatically.

src/oscar/apps/dashboard/catalogue/forms.py
```diff
@@ -300,25 +300,26 @@ class ProductImageForm(forms.ModelForm):
...
-        obj.save()
-        return obj
+    def __init__(self, data=None, *args, **kwargs):
+        self.prefix = kwargs.get('prefix', None)
+        instance = kwargs.get('instance', None)
+        if not instance:
+            initial = {'display_order': self.get_display_order()}
+            initial.update(kwargs.get('initial', {}))
+            kwargs['initial'] = initial
+        super(ProductImageForm, self).__init__(data, *args, **kwargs)
```
solarissmoke: Please use `super()` without any arguments, for consistency across codebase now that we are on Python 3. This means we only use the long form where we are explicitly calling super on something other than the parent class.

tests/functional/dashboard/test_product.py
```diff
@@ -212,3 +220,62 @@ def test_product_update_attribute_values(self):
         self.assertFalse(self.product.attr.boolean)
         self.assertEqual(self.product.attr.richtext, 'article')
         self.assertEqual(self.product.attr.date, datetime.date(2016, 10, 10))
+
+
+class TestProductImages(ProductWebTest):
+
+    def setUp(self):
+        super(TestProductImages, self).setUp()
```
solarissmoke: Please use `super()`.

tests/functional/dashboard/test_product.py
```diff
@@ -212,3 +220,62 @@ def test_product_update_attribute_values(self):
...
+        page = self.get(self.url)
+        product_form = page.form
+        product_form['images-0-display_order'] = '5'
+        product_form['images-1-display_order'] = '3'
+        product_form['images-2-display_order'] = '4'
+        product_form.submit()
+        self.product = Product.objects.get(pk=self.product.id)
+        display_orders = list(self.product.images.values_list('display_order', flat=True))
+        image_ids = list(self.product.images.values_list('id', flat=True))
+        self.assertEqual(display_orders, [3, 4, 5])
```
solarissmoke: This doesn't look right to me. Elsewhere (e.g., here: https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/catalogue/abstract_models.py#L1202-L1206) Oscar expects that the primary image will always have `display_order` of `0` - this means that the form always has to re-number the images so that they start from 0. See also the `delete()` method on `AbstractProductImage` which also enforces this logic.

src/oscar/apps/dashboard/catalogue/forms.py
sasha0: This and above super calls left from original implementation. Fixing.

src/oscar/templates/oscar/dashboard/catalogue/product_update.html
sasha0: > I don't think this help text is necessary. The user will realise as soon as they add one image that the formset grows automatically.

Can't agree with that. Behavior of the formset will change and help text should help to avoid user confusion.

In short, I prefer to leave explicit instructions and also not sure how would user make conclusion about image re-order.

solarissmoke: OK - in that case please translate the string and change grammar to "Drag images to re-order them. Space for additional images will appear when images are added."

docs/source/releases/v2.0.rst
sasha0: Fixed.

tests/functional/dashboard/test_product.py
sasha0: Ok fixed.

PR #2288, opened 2017-04-06
Dynamic loading fixes.

sasha0:


src/oscar/apps/basket/middleware.py
```diff
@@ -7,7 +7,7 @@
 from oscar.core.compat import MiddlewareMixin, user_is_authenticated
 from oscar.core.loading import get_class, get_model
 
-Applicator = get_class('offer.utils', 'Applicator')
+Applicator = get_class('offer.applicator', 'Applicator')
```
vdboor: While this is the proper location for new projects (which left me confused why my `Applicator` was not used), this could break older projects that still place the override in `offer.utils`. Is there some kind of backwards compatibility for class "moves", or should this be mentioned in the release notes?

sasha0: In fact `offer.utils` already backward compatible, since it has import of `Applicator` class - https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/offer/utils.py#L6.  Will add a note to the release notes about proper location too.

sasha0: Added a note to the release docs.

docs/source/releases/v1.5.rst
```diff
@@ -55,6 +55,9 @@ Minor changes
    addresses tracked separately: billing address in ``UserAddress.num_orders_as_billing_address`` field
    and shipping address in ``UserAddress.num_order_as_shipping_address`` accordingly.
 
+- Majority of the Oscar imports made using dynamic loading, instead of direct importing for better extensibility and
+  customizability.
+
```
mvantellingen: can we keep documentation this to max 79 chars ?

src/oscar/apps/basket/views.py
```diff
@@ -12,19 +12,20 @@
...
+Applicator = get_class('offer.applicator', 'Applicator')
 (BasketLineForm, AddToBasketForm, BasketVoucherForm, SavedLineForm) = get_classes(
     'basket.forms', ('BasketLineForm', 'AddToBasketForm',
                      'BasketVoucherForm', 'SavedLineForm'))
 BasketLineFormSet, SavedLineFormSet = get_classes(
     'basket.formsets', ('BasketLineFormSet', 'SavedLineFormSet'))
-Repository = get_class('shipping.repository', ('Repository'))
+Repository = get_class('shipping.repository', 'Repository')
+basket_addition, voucher_addition, voucher_removal = get_classes(
+    'basket.signals', ('basket_addition', 'voucher_addition', 'voucher_removal'))
```
mvantellingen: What is the reason to make signals dynamic?

docs/source/releases/v1.5.rst
sasha0: Sure.

src/oscar/apps/basket/views.py
sasha0: For consistency with other signals imports, for instance https://github.com/django-oscar/django-oscar/blob/6ce56ccbc033341a72838b0cb06935d6ba498c75/src/oscar/apps/customer/receivers.py#L7.

sasha0:
Moved `Applicator` loading fix to the separate PR - #2312.

sasha0:
During latest discussion we decided that dynamic signals loading might be not quite safe, despite we already load them dynamically in few places. 

It does not makes much sense also because there are no ways to customize signal itself - user can already pass additional parameters via `**kwargs`. Hence, to make imports consistent we could switch back to direct imports for signals and leave a note about this in the documentation.

SalahAdDin:
:+1:

src/oscar/apps/analytics/receivers.py
```diff
@@ -4,16 +4,16 @@
...
+from oscar.apps.order.signals import order_placed
 from oscar.apps.search.signals import user_search
 from oscar.core.compat import user_is_authenticated
-from oscar.core.loading import get_class, get_classes
+from oscar.core.loading import get_classes
 
 UserSearch, UserRecord, ProductRecord, UserProductView = get_classes(
     'analytics.models', ['UserSearch', 'UserRecord', 'ProductRecord',
                          'UserProductView'])
 product_viewed = get_classes('catalogue.signals', ['product_viewed'])
```
solarissmoke: `product_viewed` also needs to be changed to direct import

docs/source/releases/v1.6.rst
```diff
@@ -36,6 +36,9 @@ Removal of deprecated features
 
 Minor changes
 ~~~~~~~~~~~~~
+ - Majority of the Oscar imports now using dynamic loading, instead of
+   direct importing for better extensibility and customizability, except
+   signals.
```
solarissmoke: I think it's worth adding a note to the backwards incompatible changes to say that signals are no longer dynamically loaded.

src/oscar/apps/analytics/receivers.py
sasha0: Overlooked this one, thanks!

docs/source/releases/v1.6.rst
sasha0: Ok added.

PR #2290, opened 2017-04-08
Translations. Simplified Chinese zh-cn becomes zh-hans

shlinx:
Same as what happened in rest framework: https://github.com/encode/django-rest-framework/issues/2688
Currently on the sample site https://latest.oscarcommerce.com/, the "简体中文" language option is not working.


sasha0:
Thanks for the contribution!
How about backward compatibility with Django 1.8? Thanks.

shlinx:
@sasha0 Thanks for reminding. Sorry, this PR is closed by accident. I didn't find a solution to make it compatible with 1.8, it seems Django itself is not compatible with its old version. Do you have any thought that can give me some light? Please are you able to re-open this PR if it has some value? Thanks.

sasha0:
I think I'll have to create manage this on the Transifex side and then we can switch sandbox site to the new locale, since its already using latest Django.

SalahAdDin:
Yes, transifex is the best way.

sasha0:
At the meantime, could you create an corresponding issue this PR aimed to fix? Thanks.

shlinx:
@sasha0 Issue created. #2318  Thanks.

jxltom:
I believe compatibility to django 1.8 is not a problem since django 1.8.8 supports all followings. So let's just add support for  zh-hans and zh-hant. The ```zh-cn``` can still be used.

```
 ('zh-cn', gettext_noop('Simplified Chinese')),
 ('zh-hans', gettext_noop('Simplified Chinese')),
 ('zh-hant', gettext_noop('Traditional Chinese')),
```

PR #2295, opened 2017-04-12
Order statistics fixes.

### LINKED ISSUE #1579: order statistics status link loses date range
1. In /dashboard/orders/statistics/ apply a date filter.
   - the order statuses now count only orders within the given period
2. Click on one of the order statuses

_Actual_: A list with all orders with that status is shown.

_Expected_: A list with orders within the given period and with that status is shown.

sasha0:
Fixes #1579

### DIFF
--- a/src/oscar/templates/oscar/dashboard/orders/statistics.html
+++ b/src/oscar/templates/oscar/dashboard/orders/statistics.html
@@ -1,6 +1,7 @@
 {% extends 'dashboard/layout.html' %}
 {% load currency_filters %}
 {% load i18n %}
+{% load django_tables2 %}
 
 {% block body_class %}{{ block.super }} orders{% endblock %}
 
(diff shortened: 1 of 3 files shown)


src/oscar/templates/oscar/dashboard/orders/statistics.html
```diff
@@ -59,7 +60,7 @@
             </tr>
             {% for dict in order_status_breakdown %}
                 <tr>
-                    <td><a href="{% url 'dashboard:order-list' %}?order_status={{ dict.status }}">{{ dict.status }}</a></td>
+                    <td><a href="{% url 'dashboard:order-list' %}{% querystring 'order_status'=dict.status %}">{{ dict.status }}</a></td>
```
solarissmoke: This should be `'status'=dict.status`.

sasha0: Right, fixed.

PR #2299, opened 2017-04-13
Issue/1881/save anonymous email

### LINKED ISSUE #1881: Email to anonymous users cannot be saved

sasha0:
Fixes #1881


src/oscar/apps/customer/migrations/0004_auto_20170413_1853.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('customer', '0003_update_email_length'),
+    ]
+
+    operations = [
+        migrations.AddField(
```
solarissmoke: Please can you rebuild these migrations? Seems like you changed the `email` field in between this migration and the next one.

sasha0: Not quite:

- add optional field `email` (0004_auto_20170413_1853.py);
- fill in `email` field with data taken from `User.email` field (0005_auto_20170413_1857.py);
- make `email` field mandatory since data is in place (0006_auto_20170413_1901.py).

solarissmoke: Ah yes, my bad. Looks good to merge.

PR #2300, opened 2017-04-14
Set BOM in UnicodeCSVWriter in PY2 for better csv reader compatibility

sasha0:


solarissmoke:
The unicode standard says "Use of a BOM is neither required nor recommended for UTF-8" - i.e., it explicitly discourages use of the BOM. There some interesting discussion [here](http://stackoverflow.com/questions/2223882/whats-different-between-utf-8-and-utf-8-without-bom). 

I get that MS Excel etc are broken in this regard and there may be a case for supporting them, but we need to make sure we don't break compatibility for other more standards-compliant CSV readers.

Also why is it only being done in Python 2? Does Python 3 add a BOM by default?

sasha0:
> I get that MS Excel etc are broken in this regard and there may be a case for supporting them, but we need to make sure we don't break compatibility for other more standards-compliant CSV readers.
How would we achieve this? Shall we improve current test or etc?

> Also why is it only being done in Python 2? Does Python 3 add a BOM by default?
Not sure, this is the question to original PR author - @SteinRobert.

SteinRobert:
Well, it's been some time since I wrote that PR.  I recall that the tests für Python 3 failed and I ended up excluding Python 3 for the BOM fix. Anyways I can't tell you why it's only Python 2. It seems like the test results of the old PR are gone, at least I can't find them.
Maybe we include an optional flag for setting the BOM. In that case everyone could decide what is the better way to write the CSV files, depending on what CSV reader the users of that Oscar installation use.
What do you think about that?

solarissmoke:
> How would we achieve this? Shall we improve current test or etc?

@sasha0 the problem is it depends on client. MS Excel on Windows behaves differently from MS Excel on Mac. As far as I can tell, it is only Excel that is the problem here. To make things worse, this fix apparently [will not work on Excel on Mac anyway](https://answers.microsoft.com/en-us/msoffice/forum/msoffice_excel-mso_mac/mac-excel-converts-utf-8-characters-to-underlines/7c4cdaa7-bfa3-41a2-8482-554ae235227b) - you have to use UTF-16 for it to work there.

> Maybe we include an optional flag for setting the BOM. In that case everyone could decide what is the better way to write the CSV files, depending on what CSV reader the users of that Oscar installation use.

The issue is that this is not a site-level thing, but a user-level thing. This fix might work for someone using Excel on Windows, but won't work for someone with Excel on Mac.

Personally I think we should just stick to the unicode standard, and perhaps put a note in the documentation saying that if you have trouble with Excel, then you need to manually choose the encoding using the import wizard rather than asking it to auto-detect.

sasha0:
Yeah, agreed. I inclined to add note to the documentation and leave test.

SteinRobert:
> To make things worse, this fix apparently will not work on Excel on Mac anyway - you have to use UTF-16 for it to work there.
We have tried that, but that didn't work for us. Only the BOM setting seemed to work.
@solarissmoke I agree its completely Excel-related and it seems that there is some weird implementation on how CSV files are read. However, one should ignore the spread and impact of Excel for business people and end-users.

sasha0:
Alright, after discussion with @solarissmoke we agreed that we'll initially try to re-create issue and test fix on Python2 and Python3 as well.

sasha0:
Closing in favor of #2306

PR #2303, opened 2017-04-15
Issue/1883/dispatcher return values 

### LINKED ISSUE #1883: customer.utils.Dispatcher needs return values

sasha0:
Fixes #1883.

### DIFF
--- a/src/oscar/apps/customer/utils.py
+++ b/src/oscar/apps/customer/utils.py
@@ -24,61 +24,72 @@ class Dispatcher(object):
 
     def dispatch_direct_messages(self, recipient, messages):
         """
-        Dispatch one-off messages to explicitly specified recipient(s).
+        Dispatch one-off messages to explicitly specified recipient.
         """
         if messages['subject'] and (messages['body'] or messages['html']):
-            self.send_email_messages(recipient, messages)
+            return self.send_email_messages(recipient, messages)
 
-    def dispatch_order_messages(self, order, messages, event_type=None,
-                                **kwargs):
+    def dispatch_order_messages(self, order, messages, event_type=None, **kwargs):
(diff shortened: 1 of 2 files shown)


solarissmoke:
This looks good to me, but I think tests are necessary because there is some obscure stuff going on with tuples being passed around in various places.

sasha0:
Done, added tests.

src/oscar/apps/customer/utils.py
```diff
@@ -24,65 +24,76 @@ def __init__(self, logger=None):
...
 
-        # Create order communications event for audit
-        if event_type is not None:
-            CommunicationEvent._default_manager.create(
-                order=order, event_type=event_type)
+        self.create_communication_event(order, event_type, dispatched_messages)
+
+    def dispatch_anonymous_messages(self, email, messages):
+        dispatched_messages = {}
+        if email:
```
solarissmoke: In what situations can `email` be `None`? Is it possible for `order.guest_email` to be missing?

sasha0: I suppose only in case of email delivery failure.

sasha0: I don't think `guest_email` can be `None` - we check set it on the checkout gateway page https://github.com/django-oscar/django-oscar/blob/31ea8dcbcce4ccb6bcbdbb08c3fe9b01103fae2c/src/oscar/apps/checkout/views.py#L80-L82.

PR #2304, opened 2017-04-16
Fix logic for determining 'hurry mode' on alerts. Fixes #2100, #2103.

### LINKED ISSUE #2100: Hurry mode for customer alerts is never True

solarissmoke:
Same as the proposed changes in #2103, with slightly improved code logic and tests.

### DIFF
--- a/tests/functional/customer/test_alert.py
+++ b/tests/functional/customer/test_alert.py
@@ -1,7 +1,9 @@
 from django_webtest import WebTest
 from django.core.urlresolvers import reverse
 from django.core import mail
+from django.test import TestCase
 
+from oscar.apps.customer.alerts.utils import send_product_alerts
 from oscar.apps.customer.models import ProductAlert
 from oscar.test.factories import create_product, create_stockrecord
 from oscar.test.factories import UserFactory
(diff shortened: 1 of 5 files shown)


tests/functional/customer/test_alert.py
```diff
@@ -78,3 +80,33 @@ def test_can_create_a_stock_alert(self):
...
+        ProductAlert.objects.create(user=self.user, product=self.product)
+
+        send_product_alerts(self.product)
+
+        self.assertEqual(1, len(mail.outbox))
+        self.assertNotIn('Beware that the amount of items in stock is limited',
+            mail.outbox[0].body)
+
+    def test_hurry_mode_set_when_stock_low(self):
+        # Two alerts, 1 item in stock. Better hurry.
```
sasha0: Mb "Better to hurry up."?

solarissmoke: Both are correct English :). But I've reworded to be more straightforward. Also added tests for the case where a product has multiple stock records.

sasha0:
Looks good! Please also describe alert in the documentation and add a note to the release docs.

PR #2305, opened 2017-04-17
Make `Voucher.date_created` and `VoucherApplication.date_created` DateTimeField.

sasha0:
PR includes step-by-step migration for data safety during date field conversion to datetime:

- rename `date_created` into `date_created_original`;
- create optional field `date_created`;
- copy field values from `date_created_original` to date_created`;
- make `date_created` mandatory, remove `date_created_original`.


mvantellingen:
Why are all these steps required if only the field type is changed from a DateField to a DateTimeField?

sasha0:
During local testing on SQLite field lost data after migration. On PostgreSQL and MySQL worked fine.

mvantellingen:
Can't we do something smarter for sqlite in that case? Detect vendor name for the db driver, and do some sql statements in case sqlite is used? This migration seems unnessary heavy for projects which have large amounts of vouchers (I have some projects which have this)

sasha0:
Agreed. 👍

src/oscar/apps/voucher/migrations/0004_auto_20170417_0831.py
```diff
@@ -0,0 +1,28 @@
...
+
+from django.db import migrations
+
+
+def move_date_created_to_datetime_field(apps, schema_editor):
+    Voucher = apps.get_model('voucher', 'Voucher')
+    VoucherApplication = apps.get_model('voucher', 'VoucherApplication')
+    for voucher in Voucher._default_manager.all():
+        voucher.date_created = voucher.date_created_original
+        voucher.save()
```
mvantellingen: Besides my earlier remark, this should always be done via a simple update statement:

```python
Voucher.objects.update(date_created=F('date_created_original'))
VoucherApplication.objects.update(date_created=F('date_created_original'))
```
But I think we really should just alter the type of the field with one single migration instead.

sasha0:
Alright, left single migration and added note with warning about SQLite migration limitations.

PR #2311, opened 2017-04-20
Re-do the migrations for the customer email change.

mvantellingen:
The copying of the email address within the datamigration isn't really
optimal when a site has a large number of users and send emails.


src/oscar/apps/customer/abstract_models.py
```diff
@@ -117,7 +117,7 @@ class AbstractEmail(models.Model):
         related_name='emails',
         verbose_name=_("User"),
         null=True)
-    email = models.EmailField(_('Email Address'))
+    email = models.EmailField(_('Email Address'), null=True, blank=True)
```
solarissmoke: `null=True` doesn't have any effect on text fields

mvantellingen: Yes but since this is a new field in a potentially very large table adding a new column with a default value ('' in this case) will lock the database.

solarissmoke: Okay, makes sense. Rest looks fine to me.

PR #2315, opened 2017-04-22
Refactor product alerts to use Communication Events

### LINKED ISSUE #2310: Product alert emails should use the Dispatcher, and provide an HTML template
`customer.alerts.utils.send_product_alerts` could do with some refactoring to be more consistent with the rest of the code base. Specifically:

- Use `customer.utils.Dispatcher.send_direct_messages` to send the emails.
- Provide and use an HTML body template (currently it is plain text only, and method cannot be overridden)

I suspect the reason that `Dispatcher` wasn't used originally was because this method sends all the emails using a single SMTP connection. This use case is not unique and it might be better to let the various `Dispatcher` methods accept a `connection` which they can use if provided.

solarissmoke:
There are two key changes in this PR:

1. Allow `customer.utils.Dispatcher` to accept a `mail_connection` when instantiated, which it will use to send messages. This is useful if you are sending a lot of messages through the dispatcher and don't want to open hundreds of new SMTP connections.

2. Refactor the product alert code to use Communication Events and the Dispatcher. This means moving and renaming the email templates. The old ones are still supported for projects that have overridden them, but this raises a warning.

Fixes #2310.


sasha0:
Please add release notes and update documentation.

solarissmoke:
@sasha0 I've already added a test for the change to the Dispatcher, and the alert tests still pass. Is there anything else that needs testing?

sasha0:
Yes - if product alert was properly sent via `Distpatcher`, `Email` object created etc.

solarissmoke:
Done

PR #2320, opened 2017-04-24
Cleanup unused benefits and conditions after conditional offer removal.

### LINKED ISSUE #1493: Deleting an offer should delete related benefit/conditions
If the ConditionalOffer is deleted then it should also delete the Benefit and Condition models if they are no longer used by other offers (with exception of custom benefits/conditions)

I'll create a PR hopefully next week

samitnuk:
Fixes #1493


tests/settings.py
```diff
@@ -6,14 +6,19 @@
...
 
 ALLOWED_HOSTS = ['test', '.oscarcommerce.com']
 
 DATABASES = {
     'default': {
-        'ENGINE': os.environ.get('DATABASE_ENGINE', 'django.db.backends.postgresql_psycopg2'),
+        'ENGINE': os.environ.get('DATABASE_ENGINE',
+                                 'django.db.backends.postgresql_psycopg2'),
         'NAME': os.environ.get('DATABASE_NAME', 'oscar'),
+        'USER': 'oscaradmin',
```
solarissmoke: These changes are breaking the tests - you shouldn't have to modify this?

samitnuk: These changes are accidentally here. This will be modified asap.

src/oscar/apps/offer/signals.py
```diff
@@ -0,0 +1,27 @@
...
+Condition = get_model('offer', 'Condition')
+Benefit = get_model('offer', 'Benefit')
+
+
+@receiver(post_delete, sender=ConditionalOffer)
+def delete_related_useless_conditions_and_benefits(instance, **kwargs):
+    offer = instance  # the object is no longer in the database
+
+    condition_id = offer.condition_id
+    condition = Condition.objects.filter(id=condition_id).first()
```
solarissmoke: The more conventional query here would be `Condition.objects.get(id=conditiong_id)`. Same for the benefit query below.

solarissmoke:
One minor comment on the query syntax, otherwise this looks good to me. Also just need to check that the test failure on Django 1.8/Python 3 goes away.

Thanks!

PR #2321, opened 2017-04-25
Fix EmailBackend.authenticate() signature for Django 1.11.

solarissmoke:
In Django 1.11, `authenticate` is passed a required `request` argument, which means the current method fails. This patch modifies the method signature if we're using Django 1.11. 

While I was at it I also removed the `Emailbackend` (lower case b) class which was deprecated in Oscar 1.0.

### DIFF
--- a/tests/integration/customer/test_auth_backend.py
+++ b/tests/integration/customer/test_auth_backend.py
@@ -1,0 +1,26 @@
+import unittest
+
+import django
+from django.test import TestCase
+
+from oscar.apps.customer.auth_backends import EmailBackend
+from oscar.test.factories import UserFactory
+
+
+class AuthBackendTestCase(TestCase):
+
+    def setUp(self):
+        self.user = UserFactory(email='foo@example.com', is_staff=True)
+        self.user.set_password('letmein')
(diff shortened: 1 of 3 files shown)


mvantellingen:
LGTM 👍

tests/integration/customer/test_auth_backend.py
```diff
@@ -0,0 +1,21 @@
...
+    def test_authentication_method_signature(self):
+        user = UserFactory(email='foo@example.com', is_staff=True)
+        user.set_password('letmein')
+        user.save()
+
+        backend = EmailBackend()
+        if django.VERSION < (1, 11):
+            auth_result = backend.authenticate('foo@example.com', 'letmein')
+        else:
+            auth_result = backend.authenticate(None, 'foo@example.com', 'letmein')
```
sasha0: Maybe we can just split it into two conditional tests using `unittest.skipUnless`? That way we can make sure that test for particular Django version passed or failed.

solarissmoke: Yep - I've split into two tests.

PR #2326, opened 2017-04-26
Update supported Django/Python versions and tox configuration.

solarissmoke:
- Drop support for Django 1.9 (EOL)
- Drop support for Python 3.6 with Django 1.8 to 1.10 (these versions of Django don't officially support Python 3.6)


sasha0:
Shall we also update release notes?

solarissmoke:
Good point - updated.

PR #2327, opened 2017-04-26
Updated tinymce version

mightym:
I encountered a bug in oscar's current tinymce version on safari and chrome. The one described here http://stackoverflow.com/questions/16450499/form-with-tinymce-textarea-having-html5-required-attribute-cannot-submit

tl;dr: A required textfield will be hidden via tinymce in the dashboard. Chrome tries to focus it. But it can't. Even if there is text inside. 

I tested the 4.3 Version of tinymce (latest one available on cdn.tinymce.com, current version is 4.5.x.)


mightym:
NVM just realized that the bug is still there, even in the latest version :( 
https://github.com/tinymce/tinymce/issues/2584

PR #2332, opened 2017-04-29
Switch to using Django's password validation framework in Django >=1.9

solarissmoke:
Django 1.9 introduced the `AUTH_PASSWORD_VALIDATORS` setting which allows for more flexible password validation. Switch to using that when it is available, with a fallback to Oscar's previous behaviour in Django 1.8 or when this setting is empty.


sasha0:
LGTM. Shall we also add password validators to the default Oscar settings?

solarissmoke:
> Shall we also add password validators to the default Oscar settings?

The defaults only contain `OSCAR_` specific settings so I didn't want to do that. I have added a fallback in the code so that existing functionality is retained if the project does not define `AUTH_PASSWORD_VALIDATORS`.

PR #2333, opened 2017-04-29
Remove product reviews on product removal.

### LINKED ISSUE #862: Decide if reviews should exist without products
There's an [unresolved discussion](https://groups.google.com/forum/#!msg/django-oscar/1gur_QmbgzY/W2_MOW906cMJ) about whether reviews should be deleted when the product is deleted. This issue has been added to track it.

If it is decided that reviews can live on their own, it should be ensured that the dashboard handles that case okay. The related issue #836 has been fixed now.

sasha0:
Fixes #862


docs/source/releases/v1.5.rst
```diff
@@ -88,6 +88,9 @@ Minor changes
  - Changed product alerts to be sent using the Communication Event framework.
    The old email templates will continue to be supported until Oscar 2.0.
 
+ - Product reviews are no longer kept without products and being removed on
```
solarissmoke: This reads a bit strangely. Suggest "Product reviews are now deleted if the associated user is deleted."

solarissmoke: Other than this it looks good to me.

Issue #2176, opened 2016-11-21
Order filter form in profile throws UnicodeEncodeError


sandino:
Hi,

Order filtering form throws UnicodeEncodeError if site is displayed in language other than English. After debug I found that [mark_safe](https://github.com/django-oscar/django-oscar/blob/1.2.1/src/oscar/forms/widgets.py#L207) should accept unicode instead of string:

    return mark_safe(u'<div class="form-inline">'
                         ...
                         .format(div=div, input=input))

solarissmoke:
I am not able to reproduce this issue - have tried in several languages. Please can you provide more detailed steps to reproduce?

sandino:
@solarissmoke  
Try commenting out [line 24](https://github.com/django-oscar/django-oscar/blob/1.2.1/src/oscar/templates/oscar/customer/order/order_list.html#L24) in template

    {% include "partials/form_fields.html" with form=form style='horizontal' %}

and using something like 

```
{% for field in form %}
    <div class="my-div">{% render_field field class+='my-input' placeholder=field.label %}</div>
{% endfor %}
```

I'm specifying line numbers according to django-oscar==1.2.1, but I also managed to reproduce the error on versions 1.2.2, 1.3 and 1.4.

### Solution

Of course, it cannot be considered as a solution as it doesn't translate field labels, but at least it doesn't throw exceptions:

```
from oscar.apps.customer.forms import OrderSearchForm as CoreOrderSearchForm


class OrderSearchForm(CoreOrderSearchForm):
    def __init__(self, *args, **kwargs):
        super(OrderSearchForm, self).__init__(*args, **kwargs)
        self.fields['date_from'].label = 'From'
        self.fields['date_to'].label = 'To'
```
or we need to override the widget to pass unicode instead of string

```
class DatePickerInput(DateTimeWidgetMixin, forms.DateInput):
        ...
        return mark_safe(u'<div class="form-inline">'
                         u' {div}'
                         u'  {input}'
                         u'  <span class="input-group-addon">'
                         u'   <i class="icon-calendar glyphicon-calendar"></i>'
                         u'  </span>'
                         u' </div>'
                         u'</div>'
                         .format(div=div, input=input))
```

PR #2336, opened 2017-04-30
Use unicode_literals everywhere. Drop all uses of u prefixes on strings.

solarissmoke:
Issues like #2176 are popping up because we don't consistently enforce unicode literals everywhere. Once you start digging, you can find the same issue in several other places. In some cases this has been locally fixed by using `u` prefixes on strings - I don't think this is a good approach.

Given that we are eventually moving to Python 3, this PR implements Django's [recommendation for porting](https://docs.djangoproject.com/en/1.11/topics/python3/#unicode-literals).

It's a pretty huge diff but there are only two changes:

1. Import `unicode_literals` everywhere for consistency.

2. Drop all `u` string prefixes.


mvantellingen:
With Django 2.0 dropping 2.7 support i'm wondering how relevant this still is?

solarissmoke:
Well, Django 1.11 is an LTS release, so in theory we should be supporting Python 2 until 2021...

mvantellingen:
Ok understood. I think however we need to realise that adding unicode_literals to all modules can have some pretty big side-effects on Python 2. Accepting this PR might introduce a lot of new subtle bugs

solarissmoke:
Yeah, I too am beginning to have second thoughts about this. I think it's probably too risky and test coverage is not sufficient to catch potential issues.

sasha0:
We could limit this to only particular places where we spotted unicode-related issues, not to the whole project.

solarissmoke:
> We could limit this to only particular places where we spotted unicode-related issues, not to the whole project.

I think this is also problematic - strings would be unicode or ascii depending on which file they were declared in, and you would have to check the top of the file to see why.

sasha0:
Ok, what do you think would be the proper fix for the #2176?

solarissmoke:
This is too risky it seems, so closing. We can fix individual issues on a case-by-case basis.

PR #2342, opened 2017-05-01
Initiate pricing strategy inside the search index class in order to make it easier to override.

### LINKED ISSUE #1507: search_index strategy should not be instantiated outside of a class
The Strategy selector is instantiated outside of a class in the search_index.py.  This makes it difficult to use a custom strategy (with different arguments to the method) as we cannot override this functionality in the oscar base class

sasha0:
Fixes #1507.

### DIFF
--- a/src/oscar/apps/search/search_indexes.py
+++ b/src/oscar/apps/search/search_indexes.py
@@ -5,7 +5,6 @@ from oscar.core.loading import get_class, get_model
 # Load default strategy (without a user/request)
 is_solr_supported = get_class('search.features', 'is_solr_supported')
 Selector = get_class('partner.strategy', 'Selector')
-strategy = Selector().strategy()
 
 
 class ProductIndex(indexes.SearchIndex, indexes.Indexable):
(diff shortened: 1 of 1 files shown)


mvantellingen:
This seems to initiate the strategy multiple times which can have a negative performance impact

solarissmoke:
> This seems to initiate the strategy multiple times which can have a negative performance impact.

Easiest way to address this would be to locally cache the result of `ProductIndex.get_strategy()` and return that.

sasha0:
Add simple caching. WDYT?

src/oscar/apps/search/search_indexes.py
```diff
@@ -57,7 +56,13 @@ def prepare_rating(self, obj):
     # most common case is for customers to see the same prices and stock levels
     # and so we implement that case here.
 
+    def get_strategy(self):
+        if not self._strategy:
```
solarissmoke: Won't this raise an `AttributeError`? `_strategy` is not defined anywhere.

sasha0: Fixed.

PR #2346, opened 2017-05-02
Updated offer.custom module to load models using get_model()

### LINKED ISSUE #2345: oscar.apps.offer.custom dependency on models prevents customization
The module `oscar.apps.offer.custom` imports the `Benefit` and `Condition` models directly instead of via `get_model()`. If either of these models is overridden by a forked version, this module cannot be imported.

The module should be updated to call `get_model()`.

clintonb:
This change ensures that the module can be loaded if models are customized in forked apps.

Fixes #2345

### DIFF
--- a/src/oscar/apps/offer/custom.py
+++ b/src/oscar/apps/offer/custom.py
@@ -1,9 +1,10 @@
 from django.core import exceptions
 from django.db import IntegrityError
 
-from oscar.apps.offer.models import Benefit, Condition
 from oscar.core.loading import get_model
 
+Benefit = get_model('offer', 'Benefit')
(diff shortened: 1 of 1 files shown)


clintonb:
@sasha0 @mvantellingen please review

sasha0:
Thanks for the contribution!

clintonb:
You're welcome!

clintonb:
@sasha0 any idea when this will be released?

sasha0:
@clintonb don't know exact date yet - we've planned to release 1.5 quite soon and had to postpone due to Django-Haystack incompatibility with Django 1.11. New release will be announced in the Oscar Slack channel.

PR #2349, opened 2017-05-04
Fix logic for checking whether an offer is active based on date.

### LINKED ISSUE #2344: Undefined Behavior when ConditionalOffer.start_datetime is null
`ConditionalOffer.start_datetime` is a nullable, optional field. After thinking about it for a while, I can see two logical ways to handle this:

1. An offer with a null start_datetime is considered to always be after the start threshold.
2. An offer with a null start_datetime is considered to never be after the start threshold.

The dashboard interface doesn't not make it clear which of those behaviors is true. Initially when using the interface I assumed that it behaved as described in point 1. I.E., nulling the start date would disable the date restriction. However after reviewing the code, I noticed the found the following logic.

```python
# oscar.apps.offer.managers.ActiveOfferManager

class ActiveOfferManager(models.Manager):
    def get_queryset(self):
        cutoff = now()
        return super(ActiveOfferManager, self).get_queryset().filter(
            models.Q(end_datetime__gte=cutoff) | models.Q(end_datetime=None),
            start_datetime__lte=cutoff).filter(status=self.model.OPEN)
```

```python
# oscar.apps.offer.applicator.Applicator

    def get_site_offers(self):
        """
        Return site offers that are available to all users
        """
        cutoff = now()
        date_based = Q(
            Q(start_datetime__lte=cutoff),
            Q(end_datetime__gte=cutoff) | Q(end_datetime=None),
        )

        nondate_based = Q(start_datetime=None, end_datetime=None)

        ConditionalOffer = get_model('offer', 'ConditionalOffer')
        qs = ConditionalOffer.objects.filter(
            date_based | nondate_based,
            offer_type=ConditionalOffer.SITE,
            status=ConditionalOffer.OPEN)
        # Using select_related with the condition/benefit ranges doesn't seem
        # to work.  I think this is because both the related objects have the
        # FK to range with the same name.
        return qs.select_related('condition', 'benefit')
```

The logic in ActiveOfferManager results in a query like the following:

```sql
SELECT $COLS
  FROM "offer_conditionaloffer"
 WHERE (
          (
              "offer_conditionaloffer"."end_datetime" >= $NOW
              OR
              "offer_conditionaloffer"."end_datetime" IS NULL
          )
          AND
          "offer_conditionaloffer"."start_datetime" <= $NOW
          AND
          "offer_conditionaloffer"."status" = Open
      )
ORDER BY "offer_conditionaloffer"."priority" ASC;
```

Whereas the code in Applicator results in a query like this:

```sql
SELECT $COLS
  FROM "offer_conditionaloffer"
 WHERE (
          (
              (
                  "offer_conditionaloffer"."start_datetime" <= $NOW
                  AND
                  (
                      "offer_conditionaloffer"."end_datetime" >= $NOW
                      OR
                      "offer_conditionaloffer"."end_datetime" IS NULL
                  )
              )
              OR
              (
                  "offer_conditionaloffer"."start_datetime" IS NULL
                  AND
                  "offer_conditionaloffer"."end_datetime" IS NULL
              )
          )
          AND
          "offer_conditionaloffer"."offer_type" = Site
          AND "offer_conditionaloffer"."status" = Open
      )
ORDER BY "offer_conditionaloffer"."priority" ASC;
```

These two queries are different in an important way: when an offer's start datetime is null, Applicator treats it as active as long as the end datetime is also null. However ActiveOfferManager will always treat it as inactive.

I think there's a few action items here:

1. Choose which behavior should be exhibited by an offer with null start and/or end times. I would propose that nulling the column should disable that restriction, not disable the offer. In other words:
    1. An offer with null start date and valid end date is active as long as the end date is in the future.
    2. An offer with a valid start date and a null end date is active as long as the start date is in the past.
    3. An offer with valid start and end dates is only active when start is in the past and end is in the future.
    4. An offer with null start and end dates is always active.
2. Add help text to the offer restrictions form to explain this behavior.
3. Unify the behavior of `ActiveOfferManager` with `Applicator` and any other places similar queries are done.

solarissmoke:
- If start_datetime is null, then the offer has no start date.
- If end_datetime is null, then the offer never expires.

Use the `ActiveOfferManager` in `Applicator.get_site_offers` instead of reimplementing the same logic there.

Fixes #2344.


tests/integration/offer/test_availability.py
```diff
@@ -59,6 +59,26 @@ def test_is_inactive_after_date_range(self):
     def test_is_active_on_end_datetime(self):
         self.assertTrue(self.offer.is_available(test_date=self.end))
 
+    def test_active_on_null_end_datetime(self):
+        # null end_datetime means offer should never expire
+        offer = models.ConditionalOffer(start_datetime=self.start,
+                                             end_datetime=None)
```
sasha0: Please check indents.

solarissmoke: Fixed, thanks.

PR #2364, opened 2017-05-16
Update requirements.txt

yetkinozturk:
missing dependency sqlparse==0.2.3 is added.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -16,6 +16,7 @@ redis==2.10.5
 requests==2.14.2
 uWSGI==2.0.15
 whitenoise==3.3.0
+sqlparse==0.2.3
 
 # Linting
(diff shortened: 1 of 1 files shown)


solarissmoke:
Hi! `sqlparse` is a dependency of `django-debug-toolbar`, which is already in `requirements.txt` and gets installed with that. Can you clarify why this is needed?

solarissmoke:
Closing due to lack of response from OP. `sqlparse` is already installed as a dependency of debug toolbar as far as I can see.

yetkinozturk:
oh sorry, missed the github notification. I realized it was missing when i run the oscar app and it is fixed when i install sqlparse. No idea why, i will check again soon.

PR #2365, opened 2017-05-16
Fixing custom condition choosing

stblhnv:
When custom conditions are available, form demands a range, but no range should be attached to the custom condition. Also no another fields filled should be. Current commit fixes the bug, which appears when you try to save ConditionalOffer instance with custom condition.

### DIFF
--- a/src/oscar/apps/dashboard/offers/forms.py
+++ b/src/oscar/apps/dashboard/offers/forms.py
@@ -86,10 +86,11 @@ class ConditionForm(forms.ModelForm):
                 _("Please either choose a range, type and value OR "
                   "select a custom condition"))
 
-        if not data['custom_condition']:
-            if not data.get('range', None):
+        if data['custom_condition']:
+            if data.get('range', None) or data.get('type', None) or data.get('value', None):
                 raise forms.ValidationError(
-                    _("A range is required"))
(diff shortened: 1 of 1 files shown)


sasha0:
Thanks for the contribution! Could you please add a test?

stblhnv:
What test should I add? As I see, you already have tests for this forms.

sasha0:
Test for the issue your PR aims to resolve - save ConditionalOffer instance with custom condition.

sasha0:
When you select custom condition and save form, it just returns custom condition model instance and does not respect other fields, except `custom_condition`:

```python
def save(self, *args, **kwargs):
    # We don't save a new model if a custom condition has been chosen,
    # we simply return the instance that has been chosen
    if self.cleaned_data['custom_condition']:
        return Condition.objects.get(
            id=self.cleaned_data['custom_condition'])
    return super(ConditionForm, self).save(*args, **kwargs)
```

As for the validation, works fine already:

```python
if not data['custom_condition']:
    if not data.get('range', None):
        raise forms.ValidationError(
            _("A range is required"))
```

It requires `custom_condition` field and if it's not provided - `range` field required. Then custom condition creation handled in the save method, commented above.

PR #2381, opened 2017-05-31
Improve Chinese Simplified translate

mailgyc:
Improve Chinese Simplified translate: “仪表盘” to “控制台”, and "郵遞" in Chinese Simplified should be “邮递”


solarissmoke:
Would you be able to update your PR to resolve conflicts?

.gitignore
```diff
@@ -46,3 +46,4 @@ settings_local.py
 /sandbox/public/
 /sandbox/whoosh_index/
 /sandbox/logs/
+/oscar
```
solarissmoke: Why are you adding this to .gitignore?

mailgyc: oh, sorry, it's a mistake, it's my virtualenv folder name, you can just delete it.

solarissmoke:
Turns out this needs to happen in transifex (https://www.transifex.com/codeinthehole/django-oscar/). Would you be able to update the translations there? They will get pulled in at the next release.

Issue #2389, opened 2017-06-05
Add a datetime field to AbstractProductAttributeValue


boosh:
There's currently a `date` attribute type. It'd be really handy if there was a `datetime` attribute type as well.

solarissmoke:
Fixed by #2393

PR #2393, opened 2017-06-09
Add datetime attribute to ProductAttributeValue as asked for in issue #2389

marcustas:
Add datetime attribute to ProductAttributeValue as asked for in issue #2389

Used widget DateTimePickerInput from oscar.forms.widgets


sasha0:
LGTM.

PR #2396, opened 2017-06-14
Enhanced CountryAdmin functionalty

jsma:
Now shows display order so this can easily be seen at a glance
Allows filtering by shipping country status to narrow results
Allows searching on additional fields so admin users can search by ISO codes rather than only name


sasha0:
Thanks for the contribution!

PR #2400, opened 2017-06-18
Make the PurchaseInfo class overrideable

### LINKED ISSUE #2399: Make PurchaseInfo a proper class that can be overridden in the same way as other classes
I need to add extra methods to the `PurchaseInfo` class, but because it's defined and used directly in `apps/partner/strategy.py` I can't without duplicating that whole file.

It'd be helpful if `apps/partner/strategy.py` was changed to attempt loading that class using `get_class` and falling back to the current default if it's not been overridden somewhere else.

boosh:
Resolves #2399


sasha0:
As per #2399 - you can customize your strategy class, no need to customize `PurchaseInfo`. Thanks.

PR #2410, opened 2017-07-05
Feature/stacked offers

pjstevns:
Hi there,

Motivation for this feature change is:

* Offer a mechanism for optionally support combining multiple vouchers and offers on the same items in the same basket lines *

All this, while retaining as much as possible current behavior where on item on a basket line can be consumed by one conditional offer only.

This is done by adding an 'exclusive' flag on conditional offers, and by keeping track of the number of items on a line have been consumed by a specific offer - or by any offer.

Exclusive offers can not be combined with non-exclusive offers on the same basket line.

Feel free to comment.


src/oscar/apps/offer/abstract_models.py
```diff
@@ -62,6 +62,12 @@ class AbstractConditionalOffer(models.Model):
     offer_type = models.CharField(
         _("Type"), choices=TYPE_CHOICES, default=SITE, max_length=128)
 
+    exclusive = models.BooleanField(
+        _("Exclusive offer"),
+        help_text=_("Exclusive offers can not be combined on the same items"),
```
solarissmoke: `can not` => `cannot`

src/oscar/apps/basket/abstract_models.py
```diff
@@ -741,25 +739,42 @@ def _tax_ratio(self):
...
+        return self.consumer.available(offer) > 0
+
     # ==========
     # Properties
     # ==========
 
     @property
     def has_discount(self):
-        return self.quantity > self.quantity_without_discount
+        return self.consumer.consumed() and True or False
```
solarissmoke: Suggest `bool(self.consumer.consumed())`, feels less obtuse

src/oscar/apps/dashboard/vouchers/forms.py
```diff
@@ -44,6 +44,10 @@ class VoucherForm(forms.Form):
     benefit_value = forms.DecimalField(
         label=_('Discount value'))
 
+    exclusive = forms.BooleanField(
+        required=False,
+        label=_("Exclusive offers can not be combined on the same items"))
```
solarissmoke: `can not` => `cannot`

solarissmoke:
I think this looks good.

A lot of the benefits only really make sense in an exclusive offer (e.g., `MULTIBUY`, `FIXED`, `SHIPPING_ABSOLUTE`, `SHIPPING_FIXED_PRICE`), and could result in weird results if combined with other offers. Question is whether we just leave that to the user, or whether we need some sort of restriction in terms of what kind of offers can be non-exclusive.

pjstevns:
Samir,

Thanks for reviewing my code. I've fixed all your remarks.

What I don't understand is what you mean by 'fairly substantial change to the ConditionalOffer class'. The change there is absolutely minimal (one new attribute), and also the changes to BasketLine have been done explicitly with backward compatibility in mind.

This change only affects benefits that actually consume order lines. So shipping benefits are not impacted at all, as far as I can tell.

Also, there is nothing exclusive perse for MULTIBUY since this change was specifically designed to allow applying multiple vouchers to baskets that already qualify for multibuy (or similar) benefits.

The only thing I havent thought through completely yet, is the basis for the discounts: the gross price or net price? If you apply multiple discounts will they all apply to the line_price_excl_tax/incl_tax? Of to the _incl_discount variants? Something that only has bearing on relative discounts.

For now I've followed the principle of least surpise (and simplest method) by applying all discounts to the gross price incl tax (but excl discounts). That is what customers expect to see, and yes I am aware that for B2B sites this will have to be tweaked.

I also have a strong feeling about such considerations being out-of-scope for this change: this change is about mechanism, not policy - I want to provide a mechanism that offers a basis for more diverse policies, since one of my main headaches with oscar over the years has been the under-engineering of the offer code, forcing us to implementing conditional offers by basically forking the complete applicator logic whenever we needed combinations of offers.

solarissmoke:
> What I don't understand is what you mean by 'fairly substantial change to the ConditionalOffer class'. 

Any project that has overridden certain methods in `ConditionalOffer` (e.g., the `consume` and `discount` methods) will have to be updated to accept the new `offer` argument, etc., - i.e., so just need to highlight this in the documentation, that's all. You're right though - it should be a fairly simple change.

> I also have a strong feeling about such considerations being out-of-scope for this change: this change is about mechanism, not policy - I want to provide a mechanism that offers a basis for more diverse policies.

Yep, this sounds good to me.

pjstevns:
samir,

I agree. I'll ask @mvantellingen to merge this after 1.5 is finalised.

pjstevns:
Samir,

according to Michael, the code for 1.5 is already on a separate branch, so I don't think that there is a holdup over 1.5's release.

Feel free to verify with him of course!

solarissmoke:
That's fine - I just meant that you've currently added a note about this change to `releases/v1.5.rst` which needs to be moved to `releases/1.6.rst` :-)

pjstevns:
@solarissmoke I've moved the notes to the new v1.6.rst file

solarissmoke:
I think this is good to merge. @mvantellingen @sasha0 any objections?

sasha0:
The only thing - I think it's better to squash commits into one. WDYT?

Otherwise - looks good. Thanks @pjstevns for you efforts, nice job!

solarissmoke:
I'll squash when merging.

sasha0:
Alright, let's merge it then.

PR #2414, opened 2017-07-10
Fix Phone number in template does not translate bug.

mailgyc:
The Phone number in shipping_address.html template does not translate.

### DIFF
--- a/src/oscar/views/generic.py
+++ b/src/oscar/views/generic.py
@@ -172,7 +172,7 @@ class PhoneNumberMixin(object):
             number_required = False
             number_help_text = ''
 
-        self.fields['phone_number'] = forms.CharField(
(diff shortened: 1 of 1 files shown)


src/oscar/views/generic.py
```diff
@@ -172,7 +172,7 @@ def __init__(self, *args, **kwargs):
             number_required = False
             number_help_text = ''
 
-        self.fields['phone_number'] = forms.CharField(
+        self.fields['phone_number'] = forms.CharField(label=_('Phone number'),
```
solarissmoke: I think we need to inherit the label from the model field if it already exists - i.e., we need to set the label in the try/catch block above, and only set it again if we didn't find one.

PR #2417, opened 2017-07-13
When the query is empty, send emptystring as query

sunu:
Currently the form sends "None" as the query string.

### DIFF
--- a/src/oscar/templates/oscar/catalogue/browse.html
+++ b/src/oscar/templates/oscar/catalogue/browse.html
@@ -64,7 +64,7 @@
         {% for value in selected_facets %}
             <input type="hidden" name="selected_facets" value="{{ value }}" />
         {% endfor %}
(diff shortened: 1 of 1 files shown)


src/oscar/templates/oscar/catalogue/browse.html
```diff
@@ -64,7 +64,7 @@
         {% for value in selected_facets %}
             <input type="hidden" name="selected_facets" value="{{ value }}" />
         {% endfor %}
-        <input type="hidden" name="q" value="{{ search_form.q.value }}" />
+        <input type="hidden" name="q" value="{% firstof search_form.q.value %}" />
```
solarissmoke: This is a bit of a non-standard use of `firstof`. We should use the `default_if_none` filter instead.

sunu: Yes, `default_if_none` is surely more appropriate here. Updated.

The web ui doesn't let me squash commits. Hope that's ok.

PR #2419, opened 2017-07-14
refactor proxy_map on conditions and benefits

pjstevns:
- Allow plugin of new conditions and benefits
- Allow usage of custom versions of default conditions and benefits

The current implementation makes it very hard to:
- add a new benefit or condition to the list of proxies
- modify the behavior of a default benefit or condition 

This change seeks to fix both issues

### DIFF
--- a/tests/integration/offer/test_benefit.py
+++ b/tests/integration/offer/test_benefit.py
@@ -28,17 +28,17 @@ class TestBenefitProxyModels(object):
         This test became necessary because the complex name/description logic
         broke with the python_2_unicode_compatible decorator.
         """
-        for type, __ in Benefit.TYPE_CHOICES:
-            benefit = Benefit(type=type, range=range)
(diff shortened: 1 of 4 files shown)


tests/integration/offer/test_benefit.py
```diff
@@ -1,25 +1,46 @@
...
             benefit = Benefit(type=type, range=range)
-            self.assertTrue(all([
+            assert all([
                 benefit.name,
                 benefit.description,
-                six.text_type(benefit)]))
+                six.text_type(benefit)])
+
+    def test_proxy(self, range):
+        for type, __ in Benefit.TYPE_CHOICES:
```
sasha0: The only thing - can we not use reserved names for Python builtins? E.g. `type` -> `type_choice`.

Thanks.

pjstevns:
@sasha0 requested cleanup done now

PR #2422, opened 2017-07-18
Cast PhoneNumber field to unicode before storing in session. Fixes #2421

### LINKED ISSUE #2421: PhoneNumberField breaks customer address registration
PhoneNumber cause TypeError on json encoding for saving the billing address in session
Thats the same error as #1067 but for billing addresses.

tobiase:
JSON serializers cannot handle arbitrary python data types (of which the PhoneNumber is one).  To work around this, we cast the phone number to unicode before storing it in the session.  This works ok as the
PhoneNumber field is reconstituted correctly when the string is deserialized.

### DIFF
--- a/src/oscar/apps/checkout/utils.py
+++ b/src/oscar/apps/checkout/utils.py
@@ -165,7 +165,13 @@ class CheckoutSessionData(object):
         """
         Store address fields for a billing address.
         """
-        self._flush_namespace('billing')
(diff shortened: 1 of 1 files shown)


solarissmoke:
Could you please add unit tests for this? It looks like the session logic had no test coverage in the first place, which is why the issue has gone unnoticed for so long!

PR #2441, opened 2017-08-17
Cast PhoneNumber field to unicode before storing in session. Fixes #2421

### LINKED ISSUE #2421: PhoneNumberField breaks customer address registration

tobiase:


solarissmoke:
Looks great, thanks!

PR #2448, opened 2017-08-25
Add management of attribute options in dashboard

jwayodi:
Add the management of attribute options in the dashboard, via:
- Widget wrapper that displays change, add, and delete buttons next to a
  product attribute's "option group" field
- Forms for creating, updating, and deleting attribute option groups
- Page listing attribute option groups, with their management options
- Management of attribute option groups done in pop-up, and regular
  windows


src/oscar/static/oscar/js/oscar/RelatedObjectLookups.js
```diff
@@ -0,0 +1,128 @@
+// Handles related-objects functionality: Add Another links.
+
+var oscar = (function(o, $) {
```
solarissmoke: This will clobber the global variable of the same name that is defined in dashboard.js. Probably better to inject into that variable.

tests/functional/catalogue/test_catalogue.py
```diff
@@ -1,13 +1,29 @@
+import json
```
solarissmoke: Think these tests should go in `functional/dashboard/test_catalogue/py` instead of `functional/catalogue`

src/oscar/apps/dashboard/catalogue/widgets.py
```diff
@@ -14,3 +22,103 @@ class ProductSelectMultiple(MultipleRemoteSelect):
     # AjaxSelect(data_url=...) for overridability and backwards compatibility
     lookup_url = reverse_lazy('dashboard:catalogue-product-lookup')
     is_multiple = True
+
+
+class RelatedFieldWidgetWrapper(Widget):
```
craigloftus: You're missing test coverage for the widget.

src/oscar/apps/dashboard/catalogue/views.py
```diff
@@ -747,3 +759,238 @@ def get_context_data(self, *args, **kwargs):
...
+    def get_context_data(self, **kwargs):
+        ctx = super(AttributeOptionGroupCreateUpdateView, self).get_context_data(**kwargs)
+
+        if "attribute_option_formset" not in ctx:
+            ctx["attribute_option_formset"] = self.attribute_option_formset(
+                instance=self.object)
+
+        ctx["title"] = self.get_title()
+
+        if RelatedFieldWidgetWrapper.TO_FIELD_VAR in self.request.GET or RelatedFieldWidgetWrapper.TO_FIELD_VAR in self.request.POST:
```
solarissmoke: This logic - for handling the popups etc, is generic I assume. If so, can we move it into a resuable mixin that can, in future, be used for other related object views? Same for the other views below that are using the related object wrapper.

src/oscar/apps/dashboard/catalogue/app.py
```diff
@@ -104,6 +113,18 @@ def get_urls(self):
             url(r'^product-type/(?P<pk>\d+)/delete/$',
                 self.product_class_delete_view.as_view(),
                 name='catalogue-class-delete'),
+            url(r'^attribute-option-group/create/$',
+                self.attribute_option_group_create_view.as_view(),
+                name='catalogue-attribute-option-group-create'),
+            url(r'^attribute-option-group/$',
+                self.attribute_option_group_list_view.as_view(),
+                name='catalogue-attribute-option-group-list'),
+            url(r'^attribute-option-group/(?P<pk>\w+)/update/$',
```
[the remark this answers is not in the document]
jwayodi: The wrapper widget uses an initial placeholder update (and delete) URL (stored in the `data-href-template` attribute [1], and looking like `/dashboard/catalogue/attribute-option-group/__fk__/update/`) [2] [3], which is then later modified with an actual primary key, and set as the real URL by the JavaScript code [4], whenever the selected attribute option group changes (whether manually, or when a new one is created in the pop-up window) [5].

And it remains generic by using `django.core.urlresolvers.reverse`, with the URL pattern name constructed from the app label and model name (in this case, `catalogue` and `AttributeOptionGroup` becoming `reverse('dashboard:catalogue-attribute-option-group-update', args=('__fk__',))`. So the regex should also match alphabetic characters, and the underscore character. Django uses the dot metacharacter (`r'^(.+)/change/$'`) for the ModelAdmin urls [6], which is even more permissive. Should I go ahead and do it like this, instead?

[1] <https://github.com/jwayodi/django-oscar/blob/9a5f6b14d34dbb23a92797b0cfa502aab9f4dd99/src/oscar/templates/oscar/dashboard/catalogue/widgets/related_widget_wrapper.html#L7>
[2] <https://github.com/jwayodi/django-oscar/blob/9a5f6b14d34dbb23a92797b0cfa502aab9f4dd99/src/oscar/apps/dashboard/catalogue/widgets.py#L83>
[3] <https://github.com/jwayodi/django-oscar/blob/9a5f6b14d34dbb23a92797b0cfa502aab9f4dd99/src/oscar/apps/dashboard/catalogue/widgets.py#L60>
[4] <https://github.com/jwayodi/django-oscar/blob/9a5f6b14d34dbb23a92797b0cfa502aab9f4dd99/src/oscar/static/oscar/js/oscar/RelatedObjectLookups.js#L63>
[5] <https://github.com/jwayodi/django-oscar/blob/9a5f6b14d34dbb23a92797b0cfa502aab9f4dd99/src/oscar/static/oscar/js/oscar/RelatedObjectLookups.js#L121>
[6] <https://github.com/django/django/blob/stable/1.11.x/django/contrib/admin/options.py#L546>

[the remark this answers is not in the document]
craigloftus: Okay. What you have is fine in that case, though perhaps a one line comment to indicate why you're using that regex would be worthwhile.

solarissmoke:
This looks good to me. Let's see if anybody else has any feedback before we merge.

PR #2452, opened 2017-09-05
Don't mask invalid URL names in dashboard nav access function.

solarissmoke:
If a `url_name` is invalid, then we should allow the exception to propagate instead of silently discarding that menu item. Currently the exception is silently discarded which makes it very difficult to debug an invalid configuration of `OSCAR_DASHBOARD_NAVIGATION`.

### DIFF
--- a/docs/source/releases/v1.6.rst
+++ b/docs/source/releases/v1.6.rst
@@ -45,7 +45,7 @@ Backwards incompatible changes in Oscar 1.6
  - ``oscar.apps.offer.models.ConditionalOffer`` now has a new flag
    ``exclusive`` to denote that the offer involved can not be combined on the
    same item on the same basket line with another offer.
-   This flag is used by ``oscar.apps.basket.utils.LineOfferConsumer``, a facade 
+   This flag is used by ``oscar.apps.basket.utils.LineOfferConsumer``, a facade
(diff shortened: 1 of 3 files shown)


sasha0:
@solarissmoke maybe we could control this via setting, like `OSCAR_PROPAGATE_INVALID_URLNAME_EXCEPTION`? So that if someone has invalid records in the settings, it won't raise exception on existing instance, considering this interception exists for a long time. WDYT?

solarissmoke:
Hmm. I was just looking through the forums and this has caused quite a few people headaches because invalid menu configurations are silently ignored instead of raising an exception - it is not easy to debug this. 

So even if we did add a setting I think it might be better to default it to raise exceptions rather than silence them? You're right that it could potentially break some sites however. I am not sure what the best approach is.

sasha0:
I think it's better not to change current behavior and default not to raising an exception for the and version and probably deprecate it. Afterwards, just raise exception without a setting in the future versions.

solarissmoke:
@sasha0 I've updated the PR - rather than adding a new setting I've just changed it to log the exception. This way the behaviour is unchanged, but admins will be notified that there is a problem.

sasha0:
Let's maybe document this to release notes with a message that in the future versions exception will be propagated, not intercepted.

solarissmoke:
Added a comment to the release notes.

PR #2453, opened 2017-09-05
Add a little margin for avoid ugly effect.

SalahAdDin:


solarissmoke:
Closing as this is not a change we can make in core - you need to add such styling to your own projects.

SalahAdDin:
@solarissmoke I will explain this bug soon :D

SalahAdDin:
@solarissmoke By default we have a problem with margin inside tab panels:

faltamargen2
This is the wish list, you can see there no margin here.

problemamargen
And here you can see our custom view, in the dashboard panel.

These missing margin exist but in a few views in the dasboard.

For this reason i added it in this PR.

Thank you.

solarissmoke:
Thanks for the explanation - I see what you mean now.

However we still cannot change this in the bootstrap core less as per your PR - those are packaged as stock bootstrap rather than Oscar's modified version and we should not be modifying them. 

For the dashboard we can add some styling to the dashboard CSS to improve this. For the front-end, I think this has to be something that a project would style themselves (same as with any other Bootstrap-based template).

See also https://github.com/twbs/bootstrap/issues/13683

PR #2454, opened 2017-09-07
Add Django master to travis build.

solarissmoke:
This is to help identify issues that will arise in the next release of Django. Test failure is allowed.


solarissmoke:
This isn't working yet - tox is still pulling in Django 1.11, possibly because some of the requirements depend on it - and I can't work out how to force it.

PR #2468, opened 2017-10-03
replaced lightbox with bootstrap modal for dashboard catalogue list i…

mikedebock:
The image zoom functionality in oscar catalogue dashboard list didn't work anymore because lightbox isn't used anymore. I replaced it with bootstrap modal popup.


sasha0:
Thanks for the contribution!

solarissmoke:
@sasha0 I think there are some issues with this approach:

1. Using the UPC as an ID will fail whenever a product has no UPC - and UPC is not required. Should be using product ID instead.

2. It would be better to generate just one modal and manipulate it in JS rather than generating a separate modal for each product.

sasha0:
Right. Let's address these issues in the separate PR.

PR #2476, opened 2017-10-17
Order items in ProductAlertListView by date created, descending.

solarissmoke:
Currently they are ordered naturally, which means you see the oldest alerts first. This is not ideal for merchants who are generally going to be interested in the recently created alerts.


solarissmoke:
Have added a test to cover the `status` filter that I modified.

PR #2482, opened 2017-10-26
Fix order of namespace arguments in Application.urls

craigloftus:
Here the name `Application.name` is the name of the application, where as the `app_name` kwarg is specifying an optional instance namespace. Unsurprisingly given the confusing naming the order returned by `Application.urls` was wrong... it should be application name followed by instance namespace. See: https://docs.djangoproject.com/en/1.11/_modules/django/conf/urls/#url

It would probably be worth dropping this approach of passing a 3-tuple to `url` in the long run as this is undocumented behaviour... and renaming the `app_name` kwarg whilst doing so... but that is a PR for another day.

### DIFF
--- a/src/oscar/core/application.py
+++ b/src/oscar/core/application.py
@@ -12,7 +12,7 @@ class Application(object):
     This is subclassed by each app to provide a customisable container for an
     app's views and permissions.
     """
-    #: Namespace name
+    #: Application namespace name
     name = None
 
     login_url = None
@@ -32,7 +32,11 @@ class Application(object):
     default_permissions = None
 
     def __init__(self, app_name=None, **kwargs):
-        self.app_name = app_name
+        """
+        kwargs:
+            app_name: optionally specify the instance namespace
+        """
+        self.app_name = app_name or self.name
         # Set all kwargs as object attributes
         for key, value in kwargs.items():
             setattr(self, key, value)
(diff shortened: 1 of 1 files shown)


src/oscar/core/application.py
```diff
@@ -12,7 +12,7 @@ class Application(object):
     This is subclassed by each app to provide a customisable container for an
     app's views and permissions.
     """
-    #: Namespace name
+    #: Application namespace name
```
solarissmoke: I think you mean this comment to be "Application name" rather than "Application namespace name", as the latter confuses things even more ;-).

solarissmoke:
Hmm, looking at this some more, why do you say that `app_name` is intended to be the namespace instead of the app name? The comment on the existing code for `Application.name` states that `.name` is the namespace, as opposed to the app name. What is the problem with that?

craigloftus:
> why do you say that app_name is intended to be the namespace instead of the app name?

@solarissmoke Because it is optional (defaults to `None`), and not passed when [apps are instantiated](https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/catalogue/app.py#L50) and providing a 'real' app name is not optional, whereas providing an instance namespace is.

Passing `None` as an app name in the 3-tuple works right now for reason I don't fully grok, but definitely blows up in Django 2 because the implementation behind `url` has changed.

PR #2496, opened 2017-11-26
Add compatibility with Django 2.0 and drop support below Django 1.11

craigloftus:


solarissmoke:
@craigloftus there are a [bunch of compatibility shims](https://github.com/django-oscar/django-oscar/blob/master/src/oscar/core/compat.py#L74-L99) in `oscar.core.compat` that also need to be removed, and the calling code updated.

tests/functional/dashboard/test_catalogue.py
```diff
@@ -225,6 +225,8 @@ def test_can_create_a_child_product(self):
             'dashboard:catalogue-product-create-child',
             kwargs={'parent_pk': parent_product.pk})
         form = self.get(url).form
+        form['title'] = u'{} - variant 1'.format(parent_product.title)
+        form['upc'] = u'{}-1'.format(parent_product.upc)
```
solarissmoke: The fix I made here is wrong - there is an underlying issue with the `NullCharField` that we need to fix. I will look into it.

solarissmoke:
I managed to completely screw up the rebase. Here is a fresh PR which does things properly: #2508 .

PR #2497, opened 2017-11-27
migrate from analytics.js to gtag.js

msomierick:


src/oscar/templates/oscar/partials/google_analytics_transaction.html
```diff
@@ -1,27 +1,18 @@
...
 
-    {% for line in order.lines.all %}
-    ga('ecommerce:addItem', {
-      'id': '{{ order.number|escapejs }}',
-      'name': '{{ line.title|escapejs }}',
-      'sku': '{{ line.partner_sku|escapejs }}',
-      'category': '{{ line.category|default:"Uncategorised"|escapejs }}',
-      'price': '{{ line.unit_price_incl_tax|unlocalize|escapejs }}',
-      'quantity': '{{ line.quantity|escapejs }}'
+    gtag('event', 'add_to_cart', {
```
solarissmoke: The user is not adding an item to cart at this stage - this template is rendered on the thank you page after an order is placed. So you need to fire a `purchase` event here. See https://developers.google.com/analytics/devguides/collection/gtagjs/events#recommended-events for what parameters that should contain.

src/oscar/templatetags/google_analytics_tags.py
```diff
@@ -0,0 +1,22 @@
+from django import template
+from django.templatetags.l10n import unlocalize
+
+
+register = template.Library()
+
+
+@register.filter()
+def make_gtag_item_list(order):
```
solarissmoke: You will need to add tests for any new template tags. It should be possible to do this directly in the template though, without the need for a tag?

src/oscar/templates/oscar/partials/google_analytics_transaction.html
```diff
@@ -2,26 +2,27 @@
...
     {% for line in order.lines.all %}
-    ga('ecommerce:addItem', {
-      'id': '{{ order.number|escapejs }}',
-      'name': '{{ line.title|escapejs }}',
-      'sku': '{{ line.partner_sku|escapejs }}',
-      'category': '{{ line.category|default:"Uncategorised"|escapejs }}',
-      'price': '{{ line.unit_price_incl_tax|unlocalize|escapejs }}',
-      'quantity': '{{ line.quantity|escapejs }}'
-    });
+      gtag('event', 'add_to_cart', {
```
solarissmoke: You're still doing `add_to_cart`? The item information need to be passed to the `purchase` event `items` key. There should be no add_to_cart event here at all.

msomierick: Oops, I had not seen the previous comment.

sasha0:
@msomierick is PR now ready for the review?

sasha0:
@solarissmoke we shall add a note to the release docs.

Issue #1500, opened 2014-09-29
PhoneNumberMixin forces field name and can't be used with two fields


maikhoepfel:
The current implementation of the `PhoneNumberMixin` makes it impossible to customise the field name. It also is impossible to use it with more than one phone number field. 

We should investigate if we can instead specify field name(s) as an attribute, or convert it into a custom form field.

sasha0:
Fixed by PR #2499.

PR #2499, opened 2017-11-30
Add ability to add many phone number fields

samitnuk:
With this change phone numbers (any qty) can be added as `ModelForm`
property `phone_numbers_fields` - dict where "keys" are names of
phone numbers fields and "values" are attributes of fields.

* `PhoneNumberMixin` slightly reworked and moved from `oscar.views.generic` to `oscar.forms.mixins`
* Couple tests added (based on previous tests)

Refs #1500

### DIFF
--- a/tests/integration/checkout/test_forms.py
+++ b/tests/integration/checkout/test_forms.py
@@ -7,6 +7,23 @@ from oscar.apps.checkout.forms import ShippingAddressForm
 from oscar.test.factories import CountryFactory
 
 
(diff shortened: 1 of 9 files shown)


tests/integration/checkout/test_forms.py
```diff
@@ -1,4 +1,3 @@
-# coding=utf-8
```
solarissmoke: This is required on Python 2.

PR #2504, opened 2017-12-07
 Add validation if proxy benefit and condition classes have name property defined.

### LINKED ISSUE #2163: Not able to apply custom benefit
I followed the following [http://django-oscar.readthedocs.io/en/releases-1.2/howto/how_to_create_a_custom_benefit.html](tutorial) I did the same and tried to apply the benefit using ```create_benefit``` present in the ```oscar.apps.offer.custom``` after appling I get a recursion error and the function also warns of recursion error if not applied ```__str__ ``` but even after that I am facing the same problem. Kindly help ASAP

I am using Oscar 1.3

sasha0:
Fixes #2163

### DIFF
--- a/src/oscar/apps/offer/abstract_models.py
+++ b/src/oscar/apps/offer/abstract_models.py
@@ -27,6 +27,57 @@ BrowsableRangeManager = get_class('offer.managers', 'BrowsableRangeManager')
 
 
 @python_2_unicode_compatible
+class BaseOfferMixin(models.Model):
+    class Meta:
+        abstract = True
+
+    def proxy(self):
+        """
+        Return the proxy model
+        """
+        klassmap = self.proxy_map
+        # Short-circuit logic if current class is already a proxy class.
+        if self.__class__ in klassmap.values():
+            return self
+
+        field_dict = dict(self.__dict__)
+        for field in list(field_dict.keys()):
+            if field.startswith('_'):
+                del field_dict[field]
+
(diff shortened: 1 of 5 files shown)


src/oscar/apps/offer/abstract_models.py
```diff
@@ -26,6 +26,60 @@
...
+
+        if self.proxy_class:
+            klass = utils.load_proxy(self.proxy_class)
+            # Short-circuit again.
+            if self.__class__ == klass:
+                return self
+            return klass(**field_dict)
+        if self.type in klassmap:
+            return klassmap[self.type](**field_dict)
+        raise RuntimeError("Unrecognised condition type (%s)" % self.type)
```
solarissmoke: The wording of this exception needs to change, since it no longer applies only to `Condition`s.

src/oscar/apps/offer/abstract_models.py
```diff
@@ -26,6 +26,60 @@
...
+    def name(self):
+        """
+        A plaintext description of the benefit/condition. Every proxy class
+        has to implement it.
+
+        This is used in the dropdowns within the offer dashboard.
+        """
+        proxy_instance = self.proxy()
+        if self.proxy_class and self == proxy_instance:
+            raise AssertionError(_('Name property is not defined on proxy class.'))
```
solarissmoke: We don't usually translate developer-facing exceptions?

PR #2506, opened 2017-12-12
Invoice generation

### LINKED ISSUE #2314: Invoice generation
Original idea was suggested by @jowenn during Oscar sprint on the DjangoCon Europe 2017.

Initially, we can generate simple HTML invoice and integrate into checkout flow.

Feedback and basic invoice template is very welcomed.

- [x] Implement LegalEntity model, which would be configured on per-site basis.
- [x] Implement Invoice model.
- [x] Implement invoice template based on the guideline (https://github.com/wbond/vat_moss-python/blob/master/overview.md#invoices).
- [x] Allow to automatically generate invoice for the order after the checkout (could be on/off by the setting) or manually in the dashboard or admin.

samitnuk:
Fixes #2314


src/oscar/apps/order/utils.py
```diff
@@ -258,3 +260,22 @@ def record_voucher_usage(self, order, voucher, user):
...
+
+    Some platforms may have couple `LegalEntity`s (with couple
+    `LegalEntityAddress`s). In this case needed instances should be
+    selected based on order (ordered products).
+
+    """
+    legal_entity = LegalEntity.objects.first()
+    if legal_entity and legal_entity.has_addresses:
+
+        invoice_number = int(Invoice.get_last_invoice_number()) + 1
```
solarissmoke: This is not safe from race conditions. if the `number` is based on `id` then it should probably be a property rather than a separate field?

src/oscar/apps/order/abstract_models.py
```diff
@@ -1135,3 +1137,74 @@ def description(self):
         if self.voucher_code:
             return self.voucher_code
         return self.offer_name or u""
+
+
+@python_2_unicode_compatible
+class AbstractInvoice(models.Model):
```
solarissmoke: What information is stored in `Invoice` that we can't just store in the `Order` itself? i.e., what is the rationale for having a separate model?

samitnuk:
@solarissmoke 

Many thank you for your comments! I will review and rework this PR respectively.

src/oscar/apps/order/abstract_models.py
sasha0: Idea was to keep all information, necessary to render invoice in the separate model in case if order removed.

sasha0:
@tobiase any comments?

sasha0:
> Saving invoices to the media directory is problematic and we should not support that. Storing potentially sensitive information in a public directory (media) is unsafe. If all we are doing is generating HTML then this should just be a regular view with appropriate permissions. If projects want to generate downloadable files, then they would need to do that in a secure way, so that they can control who is able to download them. I don't think Oscar should provide that logic - i.e., all it should do is generate the Invoice object, and provide a basic view for it.

Actually, if directory contents should not be downloadable, it can be restricted on the webserver. Also, admin users might often need to check invoice for administrative/financial reasons, which usually happened on our projects. Finally, where would you suggest to store documents, out of the media directory?

solarissmoke:
> Actually, if directory contents should not be downloadable, it can be restricted on the webserver. 

Django assumes that the media directory is public by default. I think we would be setting projects up for failure if we contravene that assumption by saving potentially sensitive information to it. Everything else in the media directory is public (images etc). Invoices on the other hand should only be accessible to the customer that they are for, and administrators?

>Also, admin users might often need to check invoice for administrative/financial reasons, which usually happened on our projects. 

That is fine - but surely they should be using permission based views to do that?

>Finally, where would you suggest to store documents, out of the media directory?

Documents that need permissions should not be in the media directory because it is public. They should be served through permission-based views. If that includes PDFs etc then you would use `X-Accel-Redirect` etc to ensure that the web server can serve them securely.

SalahAdDin:
Instead it you can provide a `pdf` generation file with `weasyprint` example in the documentation.

tobiase:
I render invoices on demand and don't store them.

sasha0:
> Invoices on the other hand should only be accessible to the customer that they are for, and administrators?

Yes.

> That is fine - but they should be using permission based views to do that?

Yes.

> Documents that need permissions should not be in the media directory because it is public. They should be served through permission-based views. If that includes PDFs etc then you would use X-Accel-Redirect etc to ensure that the web server can serve them securely.

Yes, I'm now convinced that we should store invoices outside of media directory.

sasha0:
As for PDF generation - we won't be implementing that in the Oscar core and allow developers integrate this on their own.

SalahAdDin:
@sasha0 The logic, i think, is store invoices information and each one can generate his own `PDF` files as they want.

sasha0:
> * I think the structure of LegalEntity and LegalEntityAddress needs tweaking. Things like website and email should be part of LegalEntity in my opinion, and not part of LegalEntityAddress. The address should literally be just an address, and not used to store other information related to the entity.

Ok, we left only phone and fax number.

> * Settings like OSCAR_INVOICE_TERMS_AND_CONDITIONS are too specific and meet a very narrow use case. I think we should just provide a default template and expect projects to override it when they need to add custom content to the template. Most projects will have to do this anyway. Same for OSCAR_INVOICE_ADDITIONAL_DATA.

Agreed. Done.

src/oscar/apps/partner/abstract_models.py
```diff
@@ -304,3 +308,53 @@ class Meta:
...
+
+    Used in Invoices.
+    """
+    legal_entity = models.ForeignKey(
+        'partner.LegalEntity',
+        on_delete=models.CASCADE,
+        related_name='addresses',
+        verbose_name=_('Legal Entity'))
+
+    phone_number = PhoneNumberField(_('Phone number'), null=True, blank=True)
```
tobiase: phone_number and fax_number shouldn't bei null=True because the PhoneNumberField is based upon CharField.
See https://github.com/stefanfoulis/django-phonenumber-field/blob/master/README.rst

tobiase: phone_number and fax_number shouldn't bei null=True because the PhoneNumberField is based upon CharField.
See https://github.com/stefanfoulis/django-phonenumber-field/blob/master/README.rst

sasha0: Good catch, thanks. Fixed.

src/oscar/apps/order/abstract_models.py
```diff
@@ -1136,3 +1138,72 @@ def description(self):
...
+        """
+        Return rendered from html template invoice document.
+        """
+        template_name = 'order/invoice.html'
+        template_context = {
+            'invoice': self,
+            'order': self.order,
+            'legal_entity': self.legal_entity,
+            'legal_entity_address': self.legal_entity.addresses.first(),
+            'user_address': self.order.user.addresses.first()
```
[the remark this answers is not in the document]
SalahAdDin: Yes, you are right.

sasha0:
> Invoices on the other hand should only be accessible to the customer that they are for, and administrators?

For this purposes it was implemented custom storage - `oscar.core.DocumentsStorage`. Could be re-used for other private documents too.

> I think the invoice number deserves its own class, which should be easily overridable.
That would make the use of alphanumerical invoice numbers trivial.

Right. We already move it to separate class.

> I think the Invoice should store even more information than it does now. It should store everything that is needed to recreate the exact same invoice at any given time.
Invoices should be kind of immutable. The current implementation references addresses. If the customer updates an address all historical invoices using that address are wrong after the address change.
In my opinion the data needed to create a invoice can be stored denormalized, for example in a JSON-field.

It's Postgres-specific, which does not make it universal. It could be an overhead to store _all_ related order data by default, so I think if it's the goal - it should be addressed on the project basis.

I'm almost certain that this could be rather receipt generation, than invoice, per my latest comments - https://github.com/django-oscar/django-oscar/issues/2314#issuecomment-367278936. 
Any feedback is much appreciated.
Thanks.

tobiase:
> It's Postgres-specific, which does not make it universal. It could be an overhead to store all related order data by default, so I think if it's the goal - it should be addressed on the project basis.

You're right, the about the JSON field, but we could store the data in a text field for other databases.
I think the overhead it justifyable, because it provides comprehensibility.

src/oscar/core/storages.py
```diff
@@ -0,0 +1,35 @@
...
+
+        elif setting == 'FILE_UPLOAD_DIRECTORY_PERMISSIONS':
+            self.__dict__.pop('directory_permissions_mode', None)
+
+    @cached_property
+    def base_location(self):
+        return self._value_or_setting(self._location, settings.OSCAR_DOCUMENTS_ROOT)
+
+    def url(self, name):
+        raise ValueError("This file is not accessible via a URL.")
```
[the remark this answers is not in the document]
sasha0: Out of the box - it won't yet.

SalahAdDin:
We are still waiting for this :D

solarissmoke:
As per our discussions @sasha0, this is now going to be a standalone app, so closing this PR.

PR #2508, opened 2017-12-14
Drop support below Django 1.11

solarissmoke:


sasha0:
@solarissmoke @craigloftus `django-haystack` does not support Django 2.0 yet.

sasha0:
Also `django-tables2` needs to be upgraded to `>=1.17.0`, since 1.1.16 supports Django 1.8.

solarissmoke:
Oh dear... I'd forgotten about haystack. It is going to be a long time before that is compatible with Django 2...

sasha0:
Django 2.0 support for Haystack already WIP - https://github.com/django-haystack/django-haystack/pull/1557.
Otherwise, we'll have to integrate own search handler to the core and drop Haystack as dependency.

Tushant:
In the code it says 

```
Once Django 1.6 support is dropped, move AbstractBillingAddress and AbstractShippingAddress to the order app, and move PartnerAddress to the partner app. ``` but they are left there yet. Why is that so?

**Comment by solarissmoke** [comment]
I've modified this PR to drop support for EOL version of Django as per our discussion @sasha0. It will be trivial to add Django 2 support once the Haystack dependency is updated.

File: src/oscar/apps/address/abstract_models.py [hunk 1]
```diff
@@ -568,11 +567,6 @@ def validate_unique(self, exclude=None):
                 '__all__': [_("This address is already in your address"
                               " book")]})
 
-    @property
-    @deprecated
-    def num_orders(self):
-        return self.num_orders_as_shipping_address
```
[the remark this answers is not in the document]
solarissmoke: OK. Removed.

src/oscar/core/application.py
```diff
@@ -70,26 +70,11 @@ def post_process_urls(self, urlpatterns):
...
-            # In Django 1.8 and 1.9 we could distinguish the RegexURLPattern
-            # by simply checking if the pattern has a `_callback` attribute.
-            # In 1.10 this attribute is now only available as `callback`.
-            # Since the `callback` attribute is also available on patterns we
-            # should not modify (`RegexURLResolver`) we just do a isinstance()
-            # check here.
-            if DJANGO_VERSION < (1, 10):
-                if not hasattr(pattern, '_callback'):
-                    continue
+            if isinstance(pattern, URLPattern):
```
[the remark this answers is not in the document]
solarissmoke: I have updated the comment and added unit tests for the whole class.

src/oscar/apps/customer/utils.py
```diff
@@ -135,7 +134,7 @@ def get_password_reset_url(user, token_generator=default_token_generator):
     """
     kwargs = {
         'token': token_generator.make_token(user),
-        'uidb64': urlsafe_base64_encode(force_bytes(user.id)),
+        'uidb64': urlsafe_base64_encode(force_bytes(user.id)).decode(),
```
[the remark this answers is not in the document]
solarissmoke: Done.

docs/source/releases/v1.6.rst
```diff
@@ -32,36 +32,46 @@ What's new in Oscar 1.6?
...
-   locales, only current.
+   default Django URL resolvers behaviour.
+
  - Product attribute value save now handled by separate methods, similar to
    validation which allows to override save logic for file, multi-option or
    generic attribute values.
+
+ - Product attribute value save is now handled by separate methods, to make it
+     easier to override this behaviour.
+
```
sasha0: I think now there's duplicate notes about product attribute value save changes.

docs/source/releases/v1.6.rst
```diff
@@ -32,36 +32,46 @@ What's new in Oscar 1.6?
...
  - Fixed the page create/update views in the dashboard to correctly validate
    URLs. Invalid characters in URLs will raise a validation error, as will
    URLs longer than 100 characters.
- - Default shipping method on basket summary page uses default user shipping
-   address in order to calculate indicative shipping cost. Default shipping
-   address definition can be overridden in
-   ``oscar.apps.basket.views.BasketView.get_default_shipping_address`` method.
+
+ - The default shipping method ( ``oscar.apps.basket.views.BasketView.get_default_shipping_address``)
```
sasha0: Shall we move `oscar.apps.basket.views.BasketView.get_default_shipping_address` in order to keep line length limit up to 79 chars?

docs/source/releases/v1.6.rst
```diff
@@ -96,11 +123,31 @@ Backwards incompatible changes in Oscar 1.6
...
+ - jQuery UI was removed from Oscar's static files. Projects that require it
+   should install it at the project level.
+
+ - Compatibility for Internet Explorer version 9 an lower was dropped from Oscar's
+   templates and styles.
 
 Dependency changes
 ------------------
 
+- Updated jQuery to version 3.3.1.
```
sasha0: Should verb "upgraded" be more appropriate here and below?

src/oscar/apps/customer/forms.py
```diff
@@ -79,21 +79,11 @@ def get_reset_url(self, site, request, user, use_https):
 
 
 class SetPasswordForm(auth_forms.SetPasswordForm):
-
-    def clean_new_password2(self):
-        new_password2 = super(SetPasswordForm, self).clean_new_password2()
-        # For backward compatibility with Django 1.8
-        validate_password(new_password2, self.user)
-        return new_password2
+    pass
```
sasha0: Why do we need those forms at all without no customization? We can just use raw Django auth forms in the views.

solarissmoke: For backward compatibility - projects might be importing these classes. I will add a comment marking them as deprecated.

docs/source/releases/v1.6.rst
solarissmoke: Done.

docs/source/releases/v1.6.rst
solarissmoke: Fixed.

src/oscar/apps/customer/forms.py
solarissmoke: I have added a deprecation decorator to both classes. We can remove them in the next release.

PR #2516, opened 2017-12-15
Issues/1976/product options

### LINKED ISSUE #1976: List display for product with options is broken in sandbox
image

For a shop I'm doing, I have products with options. In [product.html](https://github.com/django-oscar/django-oscar/blob/master/src/oscar/templates/oscar/catalogue/partials/product.html), the `AddToBasketForm` only gets included when the product is not a parent product. But we should also exclude it if the product has options. Or maybe the `basket_form` template tag shouldn't return a form if the product has options and `simple` was specified. 

We also need to be careful to not incur too many new database queries; we want to avoid querying for options for each product.

sasha0:
Fixes #1976

### DIFF
--- a/docs/source/releases/v1.6.rst
+++ b/docs/source/releases/v1.6.rst
@@ -78,6 +78,14 @@ Minor changes
    Corresponding data migration will update hashes automatically. Take it into
    account that it will take time to run if you have large user base.
 
+ - Compact basket form is no longer displayed for the products with options
+   on product list pages since it caused form elements to impose inside the
+   product box (see :issues:`1976`).
+
+- Added ``has_options`` property to the
+  ``oscar.apps.catalogue.models.Product`` model to indicated whether the
+  instance or its product class has product options.
+
  - ``oscar.core.validators.validate_password`` was removed.
    Use ``django.contrib.auth.password_validation.validate_password`` instead.
 
--- a/src/oscar/apps/catalogue/abstract_models.py
+++ b/src/oscar/apps/catalogue/abstract_models.py
@@ -495,6 +495,16 @@ class AbstractProduct(models.Model):
         pclass_options = self.get_product_class().options.all()
         return set(pclass_options) or set(self.product_options.all())
 
+    @cached_property
+    def has_options(self):
+        # Extracting annotated value with number of product class options
+        # from product list queryset.
+        num_product_class_options = getattr(self, 'num_product_class_options', None)
+        num_product_options = getattr(self, 'num_product_options', None)
+        if num_product_class_options is not None and num_product_options is not None:
+            return num_product_class_options > 0 or num_product_options > 0
+        return self.get_product_class().options.exists() or self.product_options.exists()
+
     @property
     def is_shipping_required(self):
         return self.get_product_class().requires_shipping
--- a/tests/integration/catalogue/test_options.py
+++ b/tests/integration/catalogue/test_options.py
@@ -1,0 +1,36 @@
+from django.test import TestCase
+
+from oscar.test import factories
+from oscar.core.loading import get_model
+
+
+Product = get_model('catalogue', 'Product')
+
+
+class ProductOptionTests(TestCase):
+    def setUp(self):
+        self.product_class = factories.ProductClassFactory()
+        self.product = factories.create_product(product_class=self.product_class)
+        self.option = factories.OptionFactory()
+
+    def test_product_has_options_per_product_class(self):
+        self.product_class.options.add(self.option)
+        self.assertTrue(self.product.has_options)
+
+    def test_product_has_options_per_product(self):
+        self.product.product_options.add(self.option)
+        self.assertTrue(self.product.has_options)
+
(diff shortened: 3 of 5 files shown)


src/oscar/apps/catalogue/abstract_models.py
```diff
@@ -81,6 +82,17 @@ def __str__(self):
...
 
+    def save(self, *args, **kwargs):
+        is_product_class_has_options = self.options.count() > 0 if self.pk else False
+        super(AbstractProductClass, self).save(*args, **kwargs)
+
+        # Invalidating `has_option` field value and also value of related
+        # products of the class.
+        if is_product_class_has_options != self.has_options:
+            self.has_options = is_product_class_has_options
+            super(AbstractProductClass, self).save(*args, **kwargs)
```
craigloftus: @sasha0 Do we need to call `super().save()` twice? (here and in `Product.save`)

solarissmoke: @craigloftus they are two separate things being dealt with. `Product` and `ProductClass` both have separate m2m relationships with `catalogue.Option`. Saving the `ProductClass` will not automatically trigger updates on associated `Product`s.

tests/integration/catalogue/test_options.py
```diff
@@ -0,0 +1,23 @@
...
+from oscar.test import factories
+
+
+class TestProductOptions(TestCase):
+
+    def setUp(self):
+        self.product = factories.create_product()
+        self.option = factories.OptionFactory()
+
+    def test_product_has_options(self):
```
solarissmoke: Need additional tests to check that `has_options` is set to false whenever all options are removed.

src/oscar/apps/catalogue/abstract_models.py
```diff
@@ -451,9 +464,14 @@ def _clean_parent(self):
                 _("A parent product can't have stockrecords."))
 
     def save(self, *args, **kwargs):
+        is_product_has_options = len(self.options) > 0 if self.pk else False
```
solarissmoke: I think this will cause `has_options` to be incorrect when creating a new product whose product class has options.

specialunderwear:
Can't you di this with annotate instead of storing it in the db. It is not really data ...

solarissmoke:
I think @specialunderwear has a point here... we might be able to do this just with a property method (slightly expensive, but not too crazy). I am not too comfortable with doing stuff on `save()` and don't and don't think it is robust enough anyway (e.g., when a product is initially created the value is wrong).

specialunderwear:
That is not what I mean. I mean this: https://docs.djangoproject.com/en/2.0/topics/db/aggregation/#joins-and-aggregates did you know you can even filter on the annotated values in a query? I think you can add a num_options annotation, which you could then use for your has_options property. Maybe you can also write a
Has_options annotation directly. No extra queries in python (in the database yes it performs a subquery) and somewhat more elegant because no weird fields in the database that are not data.

sasha0:
In this case we will need to override objects loading in the search handler - https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/search/search_handlers.py#L139-L154

sasha0:
Alright, reworked using annotations.

solarissmoke:
The `num_options` annotation counts options on the product class. However if there are options set on the product itself (`Product.product_options`)  then the incorrect result is returned (`has_options` will be `False`). I don't think this is the intended behaviour?

sasha0:
Now `has_options` takes into account both product and product class options.

src/oscar/apps/catalogue/abstract_models.py
```diff
@@ -495,6 +495,16 @@ def options(self):
         pclass_options = self.get_product_class().options.all()
         return set(pclass_options) or set(self.product_options.all())
 
+    @property
+    def has_options(self):
```
solarissmoke: Please add unit tests for this logic, and also to verify that the queryset gets correctly annotated.

sasha0: Done.

docs/source/releases/v1.6.rst
```diff
@@ -74,6 +74,9 @@ Minor changes
    platform-independent and generates the same for Python 2 and 3.
    Corresponding data migration will update hashes automatically. Take it into
    account that it will take time to run if you have large user base.
+ - Compact basket form is no longer displayed for the products with options
```
solarissmoke: This looks good to me. The release note doesn't mention the new `has_options` property though, which I think is important to include somewhere in the documentation?

src/oscar/apps/catalogue/abstract_models.py
```diff
@@ -495,6 +495,16 @@ def options(self):
         pclass_options = self.get_product_class().options.all()
         return set(pclass_options) or set(self.product_options.all())
 
+    @property
```
solarissmoke: Suggest using `cached_property` here are there are queries inside that we don't want to repeat.

PR #2518, opened 2017-12-15
Order line partner permissions.

### LINKED ISSUE #2324: Improve documentation about dashboard partner permissions
In the dashboard order list orders are filtered with respect to the partner permissions (for non-staff users), so that partners see orders with stock records they own - https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/dashboard/orders/views.py#L123-L124

Despite that, in the order detail page partner sees **all** order lines, which belongs to the other partners too. In the multi-vendor projects even though partners treated as suppliers, it does not make much sense for me to display all lines and not only relevant ones.

Thanks @mightym for the observation.

@solarissmoke @mvantellingen WDYT?

sasha0:
Fixes #2324.

### DIFF
--- a/docs/source/ref/apps/dashboard.rst
+++ b/docs/source/ref/apps/dashboard.rst
@@ -31,7 +31,12 @@ If a non-staff user with the ``partner.dashboard_access`` permission is in
   record's partner's ``users``.
 * Delete and list products. Limited to products the user is allowed to update.
 * Managing orders. Similar to products, a user get access if one of an order's
-  lines is associated with a matching partner.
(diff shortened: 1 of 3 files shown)


solarissmoke:
@sasha0 having merged this I'm now thinking that it is more complicated than this. If you filter the lines, what about the order totals etc?

sasha0:
In fact, no. Supplier has nothing to do with payments and how much customer paid for the whole order.
In case of multi-vendor marketplace order placement and payments has to be handled in a different way, so this change should cover generic case.

PR #2520, opened 2017-12-17
added fix for wagtail category injection

rosscdh:
in the case of wagtail where category is injected

```
class Category(OscarWagtailAbstractCategory):
    def get_absolute_url(self):
        return self.get_url()

    def serve(self, request, *args, **kwargs):
        cls = get_class('catalogue.views', 'ProductCategoryView')

        kwargs.update({'page': self.page_ptr})

        response = cls.as_view()(request, category=self, **kwargs)

        return response
```

this is a little more optimal

### DIFF
--- a/src/oscar/apps/catalogue/views.py
+++ b/src/oscar/apps/catalogue/views.py
@@ -166,7 +166,13 @@ class ProductCategoryView(TemplateView):
         return super(ProductCategoryView, self).get(request, *args, **kwargs)
 
     def get_category(self):
-        if 'pk' in self.kwargs:
+        if 'category' in self.kwargs:
+            # in the case of integration with wagtail (cms)
+            # the category is injected into the system viw the as_view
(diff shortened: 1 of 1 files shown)


sasha0:
Could you please describe provide more details about the issue this PR aims to fix? Thanks.

rosscdh:
Of course sorry. 

Using wagtail (cms) and integrating with oscar using the (https://github.com/LabD/django-oscar-wagtail integration), a category object is injected into the context by the view and thus the base oscar view does not require a lookup on pk or other (see file diff).

solarissmoke:
I am not sure that we should be building in special cases for specific CMSes here. It also looks to me like something you can easily override at a project level. On that basis I think it would be better not to apply this in core.

rosscdh:
+1  absolutely fine with that :) and i tend to agree, i only suggested it as category should aways be injected if we have it.. to spare a lookup. all good :)

PR #2523, opened 2017-12-18
Added hook to allow implementations of strategies that depend on productoptions

specialunderwear:
Currently when I need to implement a price strategy that has price
components that are dependent on product options, a rather large part of the
basket code needs to be copy pasted, the whole of ``add_product``. The hook
added here allows me to only override the hook.

Some use cases are: Pens with engravings, the price of engraving depends on
the number of letters. But a lot more things open up when prices can be
assigned based on product configuration.

### DIFF
--- a/src/oscar/apps/basket/abstract_models.py
+++ b/src/oscar/apps/basket/abstract_models.py
@@ -165,6 +165,10 @@ class AbstractBasket(models.Model):
         self.lines.all().delete()
         self._lines = None
 
+    def get_stock_info(self, product, options):
+        "Hook for implementing strategies that depend on product options"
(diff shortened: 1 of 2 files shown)


src/oscar/apps/basket/abstract_models.py
```diff
@@ -165,6 +165,10 @@ def flush(self):
         self.lines.all().delete()
         self._lines = None
 
+    def get_stock_info(self, product, options):
+        "Hook for implementing strategies that depend on product options"
```
sasha0: Let's properly decorate docstring with three double quotes.

PR #2528, opened 2017-12-22
Fix the issue with the quote character in a category name

aleksandrpanteleymonov:
If a category has a double quote character (") in the name it's impossible to see any products in this category or it's ascenders because the query to solr brokes.


sasha0:
Doesn't it need proper escaping for all special characters in this case?

aleksandrpanteleymonov:
@sasha0 you are 100% right. I changed it to the function that espaces all the special characters. thanks for pointing out.

sasha0:
Np. Thanks for the contribution.

PR #2529, opened 2017-12-24
Distinguish active address fields definition and address hash generation.

### LINKED ISSUE #1537: Hashing for addresses can break with customised models
This is copy & paste from an [thread](https://groups.google.com/d/msgid/django-oscar/9882624b-c10a-410f-9fe3-b2f9e401b62d%40googlegroups.com?utm_medium=email&utm_source=footer) on the mailing list:
1. A customer places an order when the old (unmodified) hash code is in place. Their ShippingAddress and UserAddress get saved with a hash computed by Oscar, which is identical for both models.
2. We add a field on ShippingAddress and overload `active_address_fields` to take this field into account. As a side-effect, this changes how the hash is computed for ShippingAddress but not for UserAddress.
3. When the customer places an order with the same address as before, the system crashes with an IntegrityError, since ... Django looks up whether they already have an address in the db by computing the ShippingAddress' hash, which is different (and probably doesn't match anything). Django doesn't find a match, so it creates a UserAddress, saves that, and crashes, since the hash for the newly created UserAddress matches the one for the UserAddress that's already in the db.

Thinking about this in some more detail, it feels wrong to me that "active_address_fields" is used for two unrelated purposes (computing hashes and formatting addresses for display). It feels like Oscar should compute hashes using an internal mechanism that isn't easily corruptible by over-eager tweakers ;-)

---

Any fix for this should try to be backwards-compatible with existing deployments.

sasha0:
Fixes #1537


src/oscar/apps/address/abstract_models.py
```diff
@@ -332,25 +334,41 @@ def name(self):
...
+        return field_values
+
     def generate_hash(self):
         """
         Returns a hash of the address summary
         """
-        # We use an upper-case version of the summary
-        return zlib.crc32(self.summary.strip().upper().encode('UTF8'))
+        field_values = self.get_address_field_values(self.hash_fields)
+        return zlib.crc32(field_values.upper().encode('UTF8'))
```
solarissmoke: Looks like you need to put this through `join_fields()` before trying to hash it.

src/oscar/apps/address/abstract_models.py
```diff
@@ -351,6 +344,36 @@ def join_fields(self, fields, separator=u", "):
...
+                pass
+        return field_values
+
+    def generate_hash(self):
+        """
+        Returns a hash of the address, based on standard set of fields, listed
+        out in `hash_fields` property.
+        """
+        field_values = self.get_address_field_values(self.hash_fields)
+        return zlib.crc32(', '.join(field_values).upper().encode('UTF8')) & 0xffffffff
```
solarissmoke: What is the `0xffffffff` for here? Also, is the hash that this generates backward compatible (i.e., will it be the same hash that the previous method generated)?

src/oscar/apps/address/abstract_models.py
```diff
@@ -351,6 +344,36 @@ def join_fields(self, fields, separator=u", "):
...
+            except exceptions.ObjectDoesNotExist:
+                pass
+        return field_values
+
+    def generate_hash(self):
+        """
+        Returns a hash of the address, based on standard set of fields, listed
+        out in `hash_fields` property.
+        """
+        field_values = self.get_address_field_values(self.hash_fields)
```
solarissmoke: Need a test that checks that a hash is correctly generated when `hash_fields` is overridden (but `base_fields` is not).

tests/integration/address/test_models.py
```diff
@@ -169,6 +179,12 @@ def test_populated_shipping_address_has_same_summary_user_address(self):
         a.populate_alternative_model(sa)
         self.assertEqual(sa.summary, a.summary)
 
+    def test_summary_value(self):
```
solarissmoke: This seems to be testing essentially the same thing as the test below it?

src/oscar/apps/address/abstract_models.py
sasha0: > What is the 0xffffffff for here?

This expression allows to generate the same CRC value for both Python 2 and Python 3 version, since by default they have different range of values.

> Also, is the hash that this generates backward compatible (i.e., will it be the same hash that the previous method generated)?

It appears, that for Python 2 - no, since previously we did not enforce it to return same value for all Python versions.

tests/integration/address/test_models.py
sasha0: Indeed.

src/oscar/apps/address/abstract_models.py
solarissmoke: I think it would be problematic to break backward compatibility since the hash is stored in the DB and used for comparisons. We either need to ensure that it stays the same, or add a migration to update all existing hashes (which could be very slow if there are a large number)?

sasha0: I think firstly we need to define if we want hash to be generate the same for all versions. I believe that it's necessary and in the future ubiquitous upgrade to Python 3 is inevitable so sooner or later it has to be done.

solarissmoke: I think the hash has to be consistent given what we use it for. Maybe we need a migration that converts the old Python 2 hashes into the new ones.

tests/integration/address/test_models.py
sasha0: > I think the hash has to be consistent given what we use it for. Maybe we need a migration that converts the old Python 2 hashes into the new ones.

Maybe. In any case, user is free to skip data migration - it does not modify database schema obviously.

sasha0: > This seems to be testing essentially the same thing as the test below it?

Yes, but tests validates summary generation from different set of fields, so I think we should keep them both.

src/oscar/apps/address/abstract_models.py
sasha0: Ok done.

PR #2532, opened 2017-12-26
Use default user shipping address to calculate indicative shipping cost on basket summary page.

### LINKED ISSUE #827: Extend basket view to use default shipping address
The basket summary view attempts to look up a default shipping method so an indicative order total can be displayed.  The checkout allows the method to be changed later on though.

At the basket stage, we don't know the user's chosen shipping address so we don't pass it.  However, Oscar does have a concept of default shipping address.  We should extend the basket view to pass the user's default shipping address (if one is known).

sasha0:
Fixes #827

### DIFF
--- a/src/oscar/apps/basket/views.py
+++ b/src/oscar/apps/basket/views.py
@@ -54,10 +54,15 @@ class BasketView(ModelFormSetView):
             basket=self.request.basket, user=self.request.user,
             request=self.request)
 
+    def get_default_shipping_address(self):
+        if user_is_authenticated(self.request.user):
+            return self.request.user.addresses.filter(is_default_for_shipping=True).first()
+        return
+
     def get_default_shipping_method(self, basket):
         return Repository().get_default_shipping_method(
             basket=self.request.basket, user=self.request.user,
-            request=self.request)
+            request=self.request, shipping_addr=self.get_default_shipping_address())
 
     def get_basket_warnings(self, basket):
         """
--- a/docs/source/releases/v1.6.rst
+++ b/docs/source/releases/v1.6.rst
@@ -50,6 +50,10 @@ Minor changes
  - Fixed the page create/update views in the dashboard to correctly validate
    URLs. Invalid characters in URLs will raise a validation error, as will
    URLs longer than 100 characters.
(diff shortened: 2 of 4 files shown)


src/oscar/apps/basket/views.py
```diff
@@ -54,10 +54,16 @@ def get_shipping_methods(self, basket):
             basket=self.request.basket, user=self.request.user,
             request=self.request)
 
+    def get_default_shipping_address(self):
+        if self.request.user.is_authenticated():
+            shipping_addresses = self.request.user.addresses.filter(is_default_for_shipping=True)
```
solarissmoke: Can do this all in one query: `user.addresses.filter(is_default_for_shipping=True).first()`, which will be `None` if none exists.

solarissmoke:
The `get_default_shipping_address` method needs a unit test?

sasha0:
Quite simple to live without it.

solarissmoke:
I disagree - I think it's important even it if looks simple. Several bugs have come up recently after adding support for Django 1.11 and 2.0 because of untested code where upstream behaviour has changed. If we had proper test coverage then these would have been picked up earlier - a good example is the `AdvancedWidget` which had no tests for what seemed like a simple behaviour.

sasha0:
Alright then, agreed.

src/oscar/apps/basket/views.py
```diff
@@ -54,10 +54,15 @@ def get_shipping_methods(self, basket):
             basket=self.request.basket, user=self.request.user,
             request=self.request)
 
+    def get_default_shipping_address(self):
+        if self.request.user.is_authenticated():
```
solarissmoke: Should be using `core.compat.is_user_authenticated()` here - calling the method is deprecated.

solarissmoke:
Will you add that unit test?

sasha0:
Sure. Requested review again by mistake.

sasha0:
@solarissmoke ok I've added tests.

solarissmoke:
@sasha0 tests seem to be failing?

sasha0:
@solarissmoke all fixed and ready for the review.

Issue #1968, opened 2016-01-11
Bucket list of model fields we should probably delete


maikhoepfel:
This ticket is becoming the bucket list for model fields that I'd like to see removed. In Oscar, it's always harder to change existing fields than add new ones, so IMHO anything that provides no or little value should be removed. Feel free to suggest more fields.
- `StockRecord.price_retail` and `StockRecord.cost_price`. Not used, and make it harder to understand how pricing works, because there's already quite a few fields and methods with "price" in it.
- `order.Line.est_dispatch_date`: It is displayed when set, but not set anywhere in default Oscar. While I get it's purpose, I haven't actually seen it used in a project. I can be convinced to not delete this. 
- `order.Line.line_price_incl_tax`: The docstring already suggests those fields are redundant. This should be investigated with a bit of care, but in the spirit of normalisation, it sounds like they should be removed.

mafrosis:
FWIW I agree on `est_dispatch_date`. I spent some time investigating the point of the field, before realising it's not used. At this point it's just noise.

Nathre2501:
"est_dispatch_date" could help you if you have set delivery dates per product. I'm using this field to let the user choose a date to receive the package.

sasha0:
For now I don't see motivation to make clean up soon. We could probably deprecate fields until Oscar 2.0. 
Since I don't have problem having these fields for a while, I'd like to see opinions of other developers until we proceed. @django-oscar/api-team @django-oscar/developers @craigloftus please share your thoughts. Thanks.

inopia:
I've been using `StockRecord.price_retail` for manufacturer's suggested retail price (MSRP) and `StockRecord.cost_price` for the price the product actually costs from the supplier. These are useful for when creating sales and calculating profit margins. The fields could have better names but I feel they are useful for what they do. I'd just add them back locally if they were removed.

pwalsh:
I strongly agree with the sentiment expressed by @maikhoepfel in the original post. Fields that are not used, and which are not explicitly hooks for customisation, should be removed, and sooner rather than later.

I'd advocate for immediate deprecation, and removal in Oscar 2.

crgwbr:
We don't use `StockRecord.cost_price`, but we do use `StockRecord.price_retail` to display strike-through MSRP pricing to the customer. But, as long as we can add them back in in our own apps, if they're not used in core, it makes sense to drop them.

solarissmoke:
[...]

PR #2534, opened 2017-12-27
Deprecate unused StockRecord and order.Line model fields

solarissmoke:
- `StockRecord.price_retail`
- `StockRecord.cost_price`
- `order.Line.est_dispatch_date`
- `order.Line.line_price_incl_tax`
- `order.Line.line_price_excl_tax`

These fields are not used in Oscar core. Refs #1968.


solarissmoke:
There are a couple more fields that need to be dropped. I am also thinking of removing these fields from the corresponding dashboard forms etc to make it really obvious that they are deprecated. Any thoughts on that?

sasha0:
I think we'd remove them from the dashboard in Oscar 2.0.

solarissmoke:
In that case this is just a documentation patch and is ready for review.

PR #2552, opened 2018-01-08
Update modelling_your_catalogue.rst

fredmontet:
Small correction to create a codeblock instead of an emphasis


sasha0:
Thanks for the contribution!

PR #2553, opened 2018-01-09
Reduce forbidden monkey passwords

Azd325:


solarissmoke:
Thanks for this. However, this whole validator will be removed in the next release as Django has it's own common password validator now that you should use instead - see #2508.

Azd325:
Thanks for the info 👍

PR #2561, opened 2018-01-11
Fix URL character validation in dashboard PageUpdateForm.

### LINKED ISSUE #2559: dashboard flatpage creation tool allows user to create pages with invalid urls
You can create a flatpage with a url containing spaces and invalid characters when using the Dashboard page creation interface. This causes the page to not resolve correctly.

solarissmoke:
Also fix the `max_length` on the url field. The underlying model field has
a `max_length` of 100, so it is invalid to have a larger value here.

DRY out dashboard page update/create views and refactor tests (perform tests of form logic directly rather than indirectly through view code).

Fixes #2559.


sasha0:
Please update release notes.

PR #2562, opened 2018-01-12
refactor ReviewListView in dashboard

pjstevns:
- split out the queryset filters into separate methods for easy overloading
  in subclasses, i.e. have the keyword search in product titles as well.
- cleanup the get_date_from_to_queryset method a little to make it more dry.

### DIFF
--- a/tests/integration/dashboard/test_reviews.py
+++ b/tests/integration/dashboard/test_reviews.py
@@ -1,14 +1,49 @@
+import datetime
+
 import pytest
 from django.utils import timezone
 
 from oscar.apps.dashboard.reviews import views
+from oscar.core.loading import get_model
+from oscar.test.factories.catalogue import ProductReviewFactory
+from oscar.test.factories.customer import UserFactory
 
 
 now = timezone.now()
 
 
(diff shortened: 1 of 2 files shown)


tests/integration/dashboard/test_reviews.py
```diff
@@ -0,0 +1,74 @@
...
+
+@pytest.fixture
+def reviewlistview(rf):
+    view = views.ReviewListView()
+    view.request = rf.get('/')
+    return view
+
+
+@pytest.mark.django_db
+class TestReviewListView(object):
```
solarissmoke: I think we should also be testing whether the queryset/filters that are generated by `get_queryset()` and submethods is actually correct? It ends up being quite a complex query and currently we don't appear to have any tests to check that it is correct.

sasha0: Indeed.

sasha0: @pjstevns would you add suggested test?

pjstevns: @sasha0 I've expanded the test-coverage

PR #2567, opened 2018-01-16
Make allocation consumption/cancel and are_stock_allocations_available() skip non-stock-tracked products.

khink:
Please see this as a basis for discussion. I may have overlooked something in how Oscar works that already scratches my itch.

When using consume_stock_allocations() on all lines, even those with product that don't track stock, currently Oscar will error in is_allocation_consumption_possible() because num_allocated and num_in_stock aren't set.

In my opinion, it makes sense to facilitate this, so when customizing Oscar we can do::

        self.consume_stock_allocations(
            order,
            order.lines.all(),
            [line.quantity for line in order.lines.all()],
        )

Instead of something like::

        lines = order.lines.filter(product__product_class__track_stock==True)
        self.consume_stock_allocations(
            order,
            lines,
            [line.quantity for line in lines],
        )

I might even argue that the most common use case would be to "consume the order allocations" entirely, so `lines` and `quantities` might default to all lines and their respective quantities, so one can do:

    self.consume_stock_allocations(order)

### DIFF
--- a/src/oscar/apps/partner/abstract_models.py
+++ b/src/oscar/apps/partner/abstract_models.py
@@ -243,6 +243,8 @@ class AbstractStockRecord(models.Model):
     consume_allocation.alters_data = True
 
     def cancel_allocation(self, quantity):
+        if not self.product.product_class.track_stock:
+            return
         # We ignore requests that request a cancellation of more than the
         # amount already allocated.
         self.num_allocated -= min(self.num_allocated, quantity)
--- a/src/oscar/apps/partner/abstract_models.py
+++ b/src/oscar/apps/partner/abstract_models.py
@@ -2,6 +2,7 @@ from django.db import models, router
 from django.db.models import F, Value, signals
 from django.db.models.functions import Coalesce
 from django.utils.encoding import python_2_unicode_compatible
+from django.utils.functional import cached_property
 from django.utils.timezone import now
 from django.utils.translation import ugettext_lazy as _
 from django.utils.translation import pgettext_lazy
@@ -180,7 +181,7 @@ class AbstractStockRecord(models.Model):
             return self.num_in_stock
         return self.num_in_stock - self.num_allocated
 
-    @property
+    @cached_property
     def can_track_allocations(self):
         """Return True if the Product is set for stock tracking."""
         return self.product.product_class.track_stock
--- a/src/oscar/apps/partner/abstract_models.py
+++ b/src/oscar/apps/partner/abstract_models.py
@@ -190,6 +190,9 @@ class AbstractStockRecord(models.Model):
         product is actually shipped, then we 'consume' the allocation.
 
         """
+        # Doesn't make sense to allocate if stock tracking is off.
+        if not self.product.product_class.track_stock:
+            return
         # Send the pre-save signal
         signals.pre_save.send(
             sender=self.__class__,
--- a/src/oscar/apps/partner/abstract_models.py
+++ b/src/oscar/apps/partner/abstract_models.py
@@ -232,6 +232,8 @@ class AbstractStockRecord(models.Model):
         This is used when an item is shipped.  We remove the original
         allocation and adjust the number in stock accordingly
         """
+        if not self.product.product_class.track_stock:
+            return
         if not self.is_allocation_consumption_possible(quantity):
             raise InvalidStockAdjustment(
                 _('Invalid stock consumption request'))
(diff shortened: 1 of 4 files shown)


sasha0:
I think it does make sense, but I think in the first place it needs at least couple of tests, which would allow to determine if change could break anything.

solarissmoke:
To be consistent you would also need to apply the same logic wherever stock is allocated in the first place (and also where allocations are cancelled, etc). I suggest that it would make more sense to perform this check inside the relevant methods in the `StockRecord` class.

khink:
Thanks, i'll work on it and re-submit with tests. 

Regarding that, i couldn't get tests to run, i've asked my question in #2568.

sasha0:
No need to re-submit, you can just update remote branch.

khink:
Failure in ThankYouViewTestCase.test_analytics_event_triggered_only_on_first_view - will check

khink:
By the way, while checking coverage i noticed that `apps.order.processing.EventHandler.are_stock_allocations_available()` isn't tested. Looking further, it's not used in Oscar's code. Is this some sort of API that we expose? In that case, we should document it. Or else, can we just throw it out?

src/oscar/apps/partner/abstract_models.py
```diff
@@ -232,6 +232,8 @@ def consume_allocation(self, quantity):
         This is used when an item is shipped.  We remove the original
         allocation and adjust the number in stock accordingly
         """
+        if not self.product.product_class.track_stock:
```
solarissmoke: Need to apply the same logic in `StockRecord.allocate()`. It doesn't make sense to allocate stock if it isn't being tracked.

solarissmoke:
`are_stock_allocations_available` is a helper method, [documented here](https://github.com/django-oscar/django-oscar/blob/6ce56ccbc033341a72838b0cb06935d6ba498c75/docs/source/howto/how_to_set_up_order_processing.rst#event-handling). Would be great if you can add test coverage!

khink:
I think i'm done.

The Travis failure was due to pyflakes/pep8, which i've fixed in the last commit.

solarissmoke:
This looks good. I am going to test it out a bit more before we merge. Also let's see if anyone else has any comments.

sasha0:
Also needs a note in the release docs and mention in the documentation about how track allocation works now.

khink:
The documentation at http://django-oscar.readthedocs.io/en/releases-1.5/howto/how_to_set_up_order_processing.html will be updated with the methods' docstrings.

khink:
Is there anything else that i can do? Who should add the note to the release docs?

solarissmoke:
No, I think it's good. I will test merge it this weekend if that's okay.

We can do the release notes separately as there are a number of other changes that need to be logged there.

src/oscar/apps/partner/abstract_models.py
```diff
@@ -241,6 +246,8 @@ def consume_allocation(self, quantity):
     consume_allocation.alters_data = True
 
     def cancel_allocation(self, quantity):
+        if not self.product.product_class.track_stock:
```
solarissmoke: This logic is used in 4 places. I think we should move it into a separate property method on the stock record model called `should_manage_allocations()`. That makes it easier for a project to override the behaviour rather than changing it in 4 different places.

solarissmoke: @khink sorry, noticed one more issue regarding overridability. Would be great if you can update the PR to address this, and then we're good to merge.

khink: Not sure what you mean: I should think the `cancel/consume _stock_allocations` methods are easy enough to override if the need arises, so what should be made overridable? What scenario(s) do you have in mind?

solarissmoke: You are doing the same check [here](https://github.com/django-oscar/django-oscar/pull/2567/files#diff-0880376e680127dddc85b4b4f30f06edR196), in the order processing code. If you needed to override cancel/consume you would also have to override `order.processing.EventHandler` which I don't think is reasonable to expect someone to have to do. If we provide a wrapper method, then you can just override that and have `are_stock_allocations_available` work correctly.

solarissmoke: I've just seen your commit - it's all good now. Thanks!

src/oscar/apps/partner/abstract_models.py
```diff
@@ -180,6 +180,11 @@ def net_stock_level(self):
             return self.num_in_stock
         return self.num_in_stock - self.num_allocated
 
+    @property
```
solarissmoke: I think [`cached_property`](https://docs.djangoproject.com/en/2.0/ref/utils/#django.utils.functional.cached_property) would be good here to ensure we don't query this join repeatedly?

PR #2569, opened 2018-01-17
Document test setup.

khink:
Document how to setup a virtualenv for running test, for dummies.

### DIFF
--- a/docs/source/internals/contributing/running-tests.rst
+++ b/docs/source/internals/contributing/running-tests.rst
@@ -5,11 +5,17 @@ Test suite
 Running tests
 -------------
 
-Oscar uses pytest_ to run the tests, which can be invoked using::
+Oscar uses pytest_ to run the tests.
+.. _pytest: http://pytest.org/latest/
+
+First we create a virtualenv and install the required dependencies::
+
+    $ mkvirtualenv oscar-tests
+    $ pip install -e .[test]
+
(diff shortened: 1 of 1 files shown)


docs/source/internals/contributing/running-tests.rst
```diff
@@ -7,6 +7,8 @@ Running tests
 
 Oscar uses pytest_ to run the tests, which can be invoked using::
 
+    $ mkvirtualenv oscar-tests
+    $ pip install -e .[test]
```
sasha0: I think we should add a sentence, saying that next commands allows to create virtual environment and install dependencies in order to run tests.

khink:
There seems to be more, i get test failures: https://gist.github.com/khink/0973018b149d3a985393d9d43a5ae291

khink:
Looks like it does work with python2.7.  I used python3.4 (3.4.7 actually), will try with 3.6 and update.

khink:
I see a "bad magic number" which suggests i should throw away `.pyc` files. That fixed it, indeed.

I'm ready to start writing tests now, and i hope this addition will help others getting started more quickly.

sasha0:
Good good.

Thanks for the contribution!

khink:
Thanks for your speedy response, and your help!

sasha0:
Np!

PR #2572, opened 2018-01-22
Use a python2 virtualenv for building docs.

khink:
While updating documentation in #2571, i noticed that our `make docs` requires sphinx installed, but installing the docs requirements in a python3 virtualenv fails. 

To facilitate simple docs-generation, let's have the Makefile create the proper (python2) virtualenv. 

This has the added advantage of not interfering with the virtualenv you use for testing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,6 +28,7 @@ __pycache__
 /docs/build/*
 *.pdf
 TODO
+venv-docs
 
 # Test files
 .coverage
--- a/Makefile
+++ b/Makefile
@@ -34,10 +34,10 @@ docs:
 	cd docs && make html
 
 test:
-	py.test 
+	py.test
 
 retest:
-	py.test --lf 
+	py.test --lf
 
 coverage:
 	py.test --cov=oscar --cov-report=term-missing
@@ -70,7 +70,7 @@ clean:
 	rm -rf nosetests.xml coverage.xml htmlcov *.egg-info *.pdf dist violations.txt
 
 todo:
-	# Look for areas of the code that need updating when some event has taken place (like 
+	# Look for areas of the code that need updating when some event has taken place (like
 	# Oscar dropping support for a Django version)
 	-grep -rnH TODO *.txt
 	-grep -rnH TODO src/oscar/apps/
--- a/Makefile
+++ b/Makefile
@@ -30,8 +30,12 @@ sandbox: install build_sandbox
 sandbox_image:
 	docker build -t django-oscar-sandbox:latest .
 
-docs:
-	cd docs && make html
+venv-docs:
+	virtualenv --python=$(shell which python3.5) venv-docs
+	venv-docs/bin/pip install -r docs/requirements.txt
+
+docs: venv-docs
+	make -C docs html SPHINXBUILD=$(PWD)/venv-docs/bin/sphinx-build
 
 test:
 	py.test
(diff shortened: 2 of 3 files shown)


sasha0:
Why would we need Python 2 env, if Sphinx does support Python 3?

khink:
Then we wouldn't. But i noticed installing the docs requirements in a python3 virtualenv fails::

```kees@Pinarello ~/Projects/django-oscar (master) $ make docs
cd docs && make html
sphinx-build -a -b html -d build/doctrees   source build/html
make[1]: sphinx-build: No such file or directory
make[1]: *** [html] Error 1
make: *** [docs] Error 2
kees@Pinarello ~/Projects/django-oscar (master) $ workon oscar-tests
(oscar-tests) kees@Pinarello ~/Projects/django-oscar (master) $ python
Python 3.6.4 (default, Dec 21 2017, 20:33:21)
[GCC 4.2.1 Compatible Apple LLVM 9.0.0 (clang-900.0.38)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> ^D
(oscar-tests) kees@Pinarello ~/Projects/django-oscar (master) $ pip install -r docs/requirements.txt
Obtaining file:///Users/kees/Projects/django-oscar (from -r docs/requirements.txt (line 1))
  django-oscar 1.6.dev0 does not provide the extra 'tests'
...
Collecting transifex-client==0.13.0 (from -r docs/../requirements.txt (line 30))
  Could not find a version that satisfies the requirement transifex-client==0.13.0 (from -r docs/../requirements.txt (line 30)) (from versions: 0.1, 0.2, 0.2.1, 0.3, 0.4, 0.4.1, 0.5, 0.5.1, 0.5.2, 0.6, 0.6.1, 0.7, 0.7.2, 0.7.3, 0.8, 0.9, 0.9.1, 0.10, 0.11b0, 0.11b3, 0.11, 0.11.1b0, 0.12b0, 0.12.1, 0.12.2, 0.12.3, 0.12.4, 0.12.5)
No matching distribution found for transifex-client==0.13.0 (from -r docs/../requirements.txt (line 30))
(oscar-tests) kees@Pinarello ~/Projects/django-oscar (master) $ make docs
cd docs && make html
sphinx-build -a -b html -d build/doctrees   source build/html
Running Sphinx v1.6.6

Configuration error:
There is a syntax error in your configuration file: invalid syntax (environ.py, line 114)
Did you change the syntax from 2.x to 3.x?
make[1]: *** [html] Error 1
make: *** [docs] Error 2```

khink:
Maybe there is a way of creating docs with python3, but how?

solarissmoke:
The issue you're having is that `transifex-client` fails to install. This is because it [doesn't yet support Python 3.6](https://github.com/transifex/transifex-client/issues/208).

I'd be inclined to wait until this is fixed upstream rather than dropping to Python 2. (It works fine on Python 3.5 by the way).

sasha0:
I think we should probably move it from the basic requirements, since it is necessary only for pulling/updating translations on the Transifex and used only by core team.

khink:
Python 3.5 does work, thanks! 

I've updated the PR so now `make docs` will work out of the box. I've added a note about the Python < 3.6 requirement (and that the Makefile uses python3.5).

khink:
Thanks for the merge!

PR #2576, opened 2018-01-26
Upgrade jQuery to version 3.x, along with several related JS dependencies

solarissmoke:
This PR is quite large, as there are several interdependencies in our JS stack which all have to be updated together. The key changes are as follows:

1. Switch to using NPM and Gulp to manage JS dependencies. All dependencies are tracked in package.json, and we use Gulp to copy the relevant files into our static directory.

2. Upgrade jQuery to v3.3.1
   Upgrade Bootstrap to v3.3.7
   Upgrade bootstrap-datetime-picker to v2.4.4
   Upgrade jquery-mousewheel to v3.1.13
   Upgrade jquery-sortable to v0.9.13
   Upgrade jquery.inputmask to v3.2.7 (can't get version 3.3.x to play nicely with the datetime picker - will need to work on that later)
   Upgrade select2 to v4.0.6-rc.1
   Use select2-bootstrap-theme v0.1.0-beta.10

3. Update `oscar.forms.widgets.RemoteSelect` to work with v4 for select2.

This could do with a bit of testing to make sure nothing has broken...


solarissmoke:
Not sure whether a PR of this size is a good idea... maybe I will break this down into smaller steps and we can apply them one at a time.

PR #2577, opened 2018-01-26
Upgrade Bootstrap to v3.3.7

solarissmoke:
Also use npm to manage the dependency, and gulp to copy the files into our static directories.


SalahAdDin:
What's about `bootstrap` 4?

Issue #2554, opened 2018-01-10
Django oscar got failed migration via using mysql as backend


saurabh-mishrr:
I am trying to use django with mysql and while I was trying to migrate it throws an error ie.

"mysql.connector.errors.DatabaseError: 1553 (HY000): Cannot drop index 'catalogue_productimage_product_id_display_order_2df78171_uniq': needed in a foreign key constraint"

solarissmoke:
Please can you provide a full stack trace that shows which migration the error originates from.

our-startup:
Operations to perform:
  Apply all migrations: address, admin, analytics, auth, basket, catalogue, contenttypes, customer, flatpages, offer, order, partner, payment, promotions, reviews, sessions, shipping, sites, thumbnail, voucher, wishlists
Running migrations:
  Applying contenttypes.0001_initial... OK
  Applying auth.0001_initial... OK
  Applying address.0001_initial... OK
  Applying address.0002_auto_20150927_1547... OK
  Applying address.0003_auto_20150927_1551... OK
  Applying address.0004_auto_20170226_1122... OK
  Applying admin.0001_initial... OK
  Applying admin.0002_logentry_remove_auto_add... OK
  Applying catalogue.0001_initial... OK
  Applying analytics.0001_initial... OK
  Applying analytics.0002_auto_20140827_1705... OK
  Applying contenttypes.0002_remove_content_type_name... OK
  Applying auth.0002_alter_permission_name_max_length... OK
  Applying auth.0003_alter_user_email_max_length... OK
  Applying auth.0004_alter_user_username_opts... OK
  Applying auth.0005_alter_user_last_login_null... OK
  Applying auth.0006_require_contenttypes_0002... OK
  Applying auth.0007_alter_validators_add_error_messages... OK
  Applying auth.0008_alter_user_username_max_length... OK
  Applying sites.0001_initial... OK
  Applying partner.0001_initial... OK
  Applying customer.0001_initial... OK
  Applying basket.0001_initial... OK
  Applying basket.0002_auto_20140827_1705... OK
  Applying order.0001_initial... OK
  Applying offer.0001_initial... OK
  Applying voucher.0001_initial... OK
  Applying basket.0003_basket_vouchers... OK
  Applying basket.0004_auto_20141007_2032... OK
  Applying basket.0005_auto_20150604_1450... OK
  Applying basket.0006_auto_20160111_1108... OK
  Applying basket.0007_slugfield_noop... OK
[...]

PR #2587, opened 2018-02-01
Fix Catalogue Migrations

kooba:
This PR will turn off `db_constraint` on `productimage.product` foreign key  before removing unique constraint and put it back on after unique constraint is removed. 

This has been reported on https://github.com/django-oscar/django-oscar/issues/2554 and seems to be related to https://bugs.mysql.com/bug.php?id=37910

It seems to affect MySQL only but the change should not negatively affect other engines.


solarissmoke:
Thanks for figuring out what the issue is here. I think this patch looks good, but it highlights the fact that we don't run tests on MySQL. I am going to try and add MySQL to the test matrix before we merge this so that we can verify it fixes the issue canonically.

sasha0:
@solarissmoke for what Python/Django version do you want to add MySQL?

solarissmoke:
I was just going to add it for the whole matrix. We should be testing every version that we support with MySQL.

sasha0:
@kooba have you encountered also unicode/concurrency issues with MySQL? We tried to add MySQL to our testing matrix and found out them - https://github.com/django-oscar/django-oscar/pull/2630.
Thanks.

kooba:
@sasha0 I have not encountered these issues myself yet, but my project is still in infancy stage. I'll try to look at https://github.com/django-oscar/django-oscar/pull/2630 to see if I can help.

craigloftus:
We don't support mysql currently, so closing for now.

PR #2588, opened 2018-02-01
Add make target for test, Makefile-created virtualenv can be python3.5 / python3.6.

khink:
It should now be much easier to run the test (`make test` does everything). You can also still do it like before. The documentation is updated accordingly.

We reuse the docs virtualenv, so we rename that from venv-docs to venv.

Existing documentation was changed to also create a virtualenv (instead of using mkvirtualenv). This ensures you're in the same situation as when using the make command.

### DIFF
--- a/docs/source/internals/contributing/running-tests.rst
+++ b/docs/source/internals/contributing/running-tests.rst
@@ -8,9 +8,19 @@ Running tests
 Oscar uses pytest_ to run the tests.
 .. _pytest: http://pytest.org/latest/
 
+The fast way is::
+
+    $ make test
+
+This will create a virtualenv in `venv`, install the test dependencies and run py.test.
+
+Details
+~~~~~~~
+
 First we create a virtualenv and install the required dependencies::
 
-    $ mkvirtualenv oscar-tests
+    $ virtualenv venv
+    $ source venv/bin/activate
     $ pip install -e .[test]
 
 Then we invoke pytest using ::
--- a/docs/source/internals/contributing/running-tests.rst
+++ b/docs/source/internals/contributing/running-tests.rst
@@ -2,6 +2,14 @@
 Test suite
 ==========
 
+Testing requirements
+--------------------
+
+You'll need:
+
+- A running SQL server (PostgreSQL, or SQLite with `--sqlite` params)
+- python3.5 or python3.6
+
 Running tests
 -------------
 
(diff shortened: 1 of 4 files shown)


sasha0:
I still think we should drop transifex client from the requirements at the first place.

khink:
@sasha0 I currently don't know enough about it to judge that. But if it's easy to drop, then let's go for it.

I would think this should be a different PR, to also contain the motivation. 

But it makes sense to not merge this until that issue is resolved. We are about to recommend virtualenv setup that is not python3.6 compatible, which i think is bad.

By the way, i think python3.6 support in transifex-client is not far away, see https://github.com/transifex/transifex-client/pull/213.

khink:
Looks like the next transifex-client will support Python3.6: https://github.com/transifex/transifex-client/pull/214

khink:
https://pypi.python.org/pypi/transifex-client/0.13.1 was released, so now we can use python3.5 and python3.6 in the Makefile. I updated the documentation accordingly.

solarissmoke:
This looks good to me. Dropping transifex, if we do it, should be handled separately. Currently the travis builds require it so it's not trivial to just drop it from requirements.

sasha0:
We can then move it to the test requirements from general ones.

khink:
I don't think it is a test dependency at the moment.  It's required by `requirements.txt` but that isn't necessary for running the tests (`docs/requirements.txt` requires it, though).

`requirements.txt` is installed by `make install` which does more than running tests (http://django-oscar.readthedocs.io/en/releases-1.5/internals/contributing/development-environment.html). But i can't find any Make command that uses it.

From the looks of things, all that happens is that `venv/bin/tx` is created, but i can't see any place in the docs where its usage is explained. So i'm not sure if it's a requirement at all.

docs/source/internals/contributing/running-tests.rst
```diff
@@ -2,15 +2,33 @@
 Test suite
 ==========
 
+Testing requirements
+--------------------
+
+You'll need:
+
+- A running PostgreSQL server
```
sasha0: SQLite also works, through `--sqlite` params. Although couple of tests would not pass.

khink:
Do we want to handle dropping transifex-client in this PR? I think not, and then this can be merged, right?

solarissmoke:
Agreed - let's address transifex and any other issues separately.

PR #2589, opened 2018-02-01
Upgrade select2 to v4.0, and refactor RemoteSelect widget to work with it

solarissmoke:
Select2 now expects a regular select (instead of a hidden input) to be rendered on the field that you want to use it with. This patch refactors our `RemoteSelect` widget to do that.


solarissmoke:
This fails on Django 1.10 because several methods in the `Select` widget have a different signature. Since Django 1.10 is end of life I don't know whether it is worth trying to fix that - I'd rather just drop 1.10 from the travis matrix. Any objections?

solarissmoke:
This is ready for review. Please test that the various dashboard forms all work correctly - especially the remote select used for upsell products etc.

sasha0:
Please rebase remote branch.

PR #2317, opened 2017-04-22
Use mailgun transaction email template for the customer emails.

sasha0:
Fixes #1454

PR #2590, opened 2018-02-01
Transform Emails with Premailer

alexphelps:
This is a followup on #2317 to implement `premailer` to inline the css for email client compatibility as recommended by [mailgun](http://blog.mailgun.com/transactional-html-email-templates/).

### DIFF
--- a/docs/source/releases/v2.0.rst
+++ b/docs/source/releases/v2.0.rst
@@ -83,3 +83,5 @@ Dependency changes
 
 Deprecated features
 ~~~~~~~~~~~~~~~~~~~
+
+- ``offer.Range.contains()`` is deprecated. Use ``contains_product()`` instead.
--- a/src/oscar/apps/offer/abstract_models.py
+++ b/src/oscar/apps/offer/abstract_models.py
@@ -16,6 +16,7 @@ from django.utils.timezone import get_current_timezone, now
 from django.utils.translation import gettext_lazy as _
 
 from oscar.core.compat import AUTH_USER_MODEL
+from oscar.core.decorators import deprecated
 from oscar.core.loading import get_class, get_classes, get_model
 from oscar.models import fields
 from oscar.templatetags.currency_filters import currency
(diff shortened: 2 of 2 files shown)


alexphelps:
@sasha0 Please review and let me know if you think this needs anything else before being considered for merging.

sasha0:
I investigated a way to introduce `premailer` within the PR #2317 and faced the issue with `base_url` - https://github.com/django-oscar/django-oscar/pull/2317#issuecomment-347895085. `premailer` determines local static path as absolute - due to the leading slash and tries to load it as is.  Providing `base_url` explicitly didn't look as a good idea - premailer then would try to download the file, rather then just process it.

alexphelps:
@sasha0 the way that I've added premailer doesnt require a base_url due to the CSS being inline in a `<style></style>` block and not being in an external file which follows the basic example [here](https://github.com/peterbe/premailer).

I'm still not quite understanding the external CSS file or why one would implement an email with one  given that external CSS is not supported by email. One option with django templates is to create a template partial with the email css inside of a `<style></style>` block that can be included in the template with would be processed by premailer when the template is compiled. 

@craigloftus I agree on adding the option to disable it.

sasha0:
> I'm still not quite understanding the external CSS file or why one would implement an email with one given that external CSS is not supported by email. 

External CSS can be used only in conjunction with Premailer - e.g. you don't need to place inline styles in the `<style></style>` tags - Premailer would fetch and process all for you. Although due to the complications, I mentioned in the comment above https://github.com/django-oscar/django-oscar/pull/2590#issuecomment-364454356, it's not easily possible to achieve now.

I took a look at the code again and agree with @craigloftus. So, please rebase remote branch, resolve conflicts, adjust PR due to the feedback above and mention about this change in the release notes.

Thanks.

SalahAdDin:
@alexphelps did you?

alexphelps:
@SalahAdDin Sorry, havent had time to get back to this. We are using premailer internally on our project now but haven't time to get back to creating a stronger solution to be able to add it to oscar. 

The solution we are using does not use an external file as mentioned, trying to use an external file will cause issues.

Our email base.html template pulls in the css as a jinja template include like this
```
<style>
  {% include "customer/emails/partials/styles.css" %}
</style>
```

Now the CSS is inline in the head, then premailer can transform the html to inline all the styles as so. 
```
from premailer import transform
messages['html'] = transform(messages['html'])
```

Issue #783, opened 2013-08-07
It's hard to add extra context variables into emails


codeinthehole:
Example: I want to have a setting for a contact phone number that I can use in all email templates.  

One potential solution would be to use template context processors but these only work if you use a `RequestContext` instance.  The `get_messages` method within uses a normal `Context` instance which is correct as the request won't always be available when messages are sent from offline processes.

Not sure on the best approach.  Maybe another settings of context variables for email templates.

Habu-Kagumba:
Was a solution ever found for this? @codeinthehole

sasha0:
Should definitely be part of #319.

solarissmoke:
This will be addressed in #2591.

kevinrenskers:
I'm sorry to be commenting on such an old issue, but how exactly has this been addressed? I see the code where the email is sent, with the extra context: https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/checkout/mixins.py#L267-L278. But when I create an email template in the database with the code `ORDER_PLACED` and I want to use new context values.. how do I actually extend the context?

Issue #319, opened 2012-09-27
Review Oscar's communications - and possibly consolidate them into a single framework


codeinthehole:
Oscar currently sends emails via the customer dispatcher and also has a notifications framework.  There are also a few places where emails are sent directly from views.

These components should be reviewed and possibly unified into a single communications framework.

sasha0:
Perhaps we can move it to the separate application and see if it is flexible and customizable enough? WDYT @solarissmoke @mvantellingen ?

solarissmoke:
Agreed - there is a case for refactoring the entire communications workflow. I think a good first step would be to make sure that all existing communications are sent through the Dispatcher?

sasha0:
Yes, that would be a good step to start with.

sasha0:
Having done with @samitnuk large refactoring on one of the projects, in order to have single and consistent backend for all kinds of notifications, now I find original idea even more useful.

solarissmoke:
Fixed in #3260.

PR #2591, opened 2018-02-01
Refactor communications

sasha0:
Refs #319 #783 

Summary:
* Created separate app - `communication`.
* `Dispatcher` moved to `communication` app's utils.
* `Dispatcher` improved with methods to send each separate email.
* Added ability to send email with attachments.
* All event codes (for each separate email) now kept in `Dispatcher`.
* Sending notifications moved to `communication` app as well.
* Added ability to chose if sent email should be saved to DB (through `OSCAR_SAVE_SENT_EMAILS_TO_DB`, `True` by default).
* All relevant models moved from `customer` app to `communication` app (with data migrations) - `CommunicationEventType`, `Email` and `Notification`.
* Moved classes and models added to `MOVED_CLASSES` and `MOVED_MODELS`.
* Tests adapted to check new functionality.
* Added tests to check the sending of each separate email.
* Deleted `TestOrderPlacementMixin` - Now we use `Dispatcher` to send emails and code (event code) does not used when we get message context for email that informs about a placed order.
* All emails' templates (`commtype_*`) moved from `customer/emails` to `communication/emails`.
* Templates from `customer/email/*.html` and `customer/notification/*.html` moved to `communication/email/*.html` and `communication/notification/*.html`.


craigloftus:
@samitnuk @sasha0 To help guide the review would you be able to provide a summary of the changes in the PR, what the intention in the PR is and whether there is ambition for a follow up PR to refactor further?

samitnuk:
Summary:
* Created separate app - `communication`.
* `Dispatcher` moved to `communication` app.
* `Dispatcher` improved with methods to send each separate email.
* All event codes (for each separate email) now kept in `Dispatcher`.
* Sending notifications and products alerts moved to `communications` app as well.
* Added ability to chose if sent email should be saved to DB (through `OSCAR_SAVE_SENT_EMAILS_TO_DB`, `True` by default).
* All relevant models moved to `communication` app (with data migrations).
* Moved classes and models added to `MOVED_CLASSES` and `MOVED_MODELS`.

* Tests adapted to check new functionality.
* Added tests to check the sending of each separate email.

* Deleted `TestOrderPlacementMixin` - Now we use `Dispatcher` to send emails and `code` (event code) does not used when we get message context for email that informs about a placed order.

sasha0:
@craigloftus @solarissmoke we added summary of changes for the PR.

solarissmoke:
My thoughts after a first pass are as follows:

- I think the `communication` app should be a **generic app used to send communications** (primarily email, but that can be extended to support any other type of communication). It should provide a `Dispatcher` that can be used by any app that wants to send messages somewhere. It should store `CommunicationEventType`, `Email` and `Notification` but nothing else.

  It should not have any dependencies on any other Oscar app - I think this is very important. This will ensure that projects can customise the Dispatcher behaviour easily, and use it for any other types of communication - e.g., communication to partners, or other users that are not customers. This for me is the main reason to move it out of the `customer` app.

  It should be responsible for sending messages, and for locating and rendering the right templates for messages. It should not however be responsible for creating messages themselves, or for performing any logic to generate the information content of messages.

- I don't think we should be moving some of the models/classes that have currently been moved. e.g., `ProductAlert` and `ProductAlertForm` belong in the `customer` app and should stay there. Otherwise we create too much interdependency between apps which is always problematic for projects that want to override.

- Similarly stuff like `send_order_placed_email_for_user` or `send_password_reset_email_for_user` should not live in the `Dispatcher`. That logic needs to stay in the respective apps (`checkout` and `customer`). Otherwise, if a project wanted to change the logic for sending order confirmation emails, it would have to override *both* the `checkout` and `communication` apps - instead of just one. 

I'd suggest we start with just moving the `Dispatcher`, `Email`, and `Notification` classes to this new app, and making sure that they are completely generic and have no dependencies on any other app. Then we can refactor how all the other apps are using the Dispatcher, and fix the places that aren't using it (e.g., product alerts).

sasha0:
> I don't think we should be moving some of the models/classes that have currently been moved. e.g., `ProductAlert` and `ProductAlertForm` belong in the `customer` app and should stay there. Otherwise we create too much interdependency between apps which is always problematic for projects that want to override.

@solarissmoke so basically frontend and dashboard views should stay where they are?

sasha0:
> I'd suggest we start with just moving the `Dispatcher`, `Email`, and `Notification` classes to this new app, and making sure that they are completely generic and have no dependencies on any other app. Then we can refactor how all the other apps are using the Dispatcher, and fix the places that aren't using it (e.g., product alerts).

Ok, I'm leaving only those mentioned models in the new application. What do you think about `CustomerDispatcher` and `OrderDispatcher` for sending notifications, such as `send_order_placed_email_for_user` or `send_password_reset_email_for_user`?

solarissmoke:
I've had another look at this - still feel that there is confusion about what this `communication` app should be doing.

In my view the communication app should be a generic app *used to send communications only* (primarily email, but that can be extended to support any other type of communication). It should provide a Dispatcher that can be used by any app that wants to send messages somewhere. It should store CommunicationEventType, Email and Notification but nothing else. *It should have no dependency on any other apps*.

Things like `send_registration_email_for_user`, `send_order_placed_email_for_user`, `notify_user_about_product_alert` do not belong here - this defeats the whole purpose of having a dispatcher that is a generic tool for sending messages.

The only methods that the Dispatcher should expose to other apps are `dispatch_anonymous_messages`, `dispatch_user_messages` - everything else should be handled by the app that is generating the message.

This means that the communications app will actually be quite simple - it just provides a Dispatcher that is easy for projects to extend.

src/oscar/apps/checkout/mixins.py
```diff
@@ -268,44 +265,15 @@ def send_signal(self, request, response, order):
...
-            dispatcher = Dispatcher(logger)
-            dispatcher.dispatch_order_messages(order, messages,
-                                               event_type, **kwargs)
-        else:
-            logger.warning("Order #%s - no %s communication event type",
-                           order.number, code)
+    def send_order_placed_email(self, order):
+        extra_context = self.get_message_context(order)
+        dispatcher = Dispatcher(logger=logger)
+        dispatcher.send_order_placed_email_for_user(order, extra_context)
```
solarissmoke: As noted in my comment on the main thread, I think it is a mistake to move this logic to the Dispatcher. The order app itself needs to be responsible for generating the order placed email.

samitnuk: What do you think about the idea suggested by @sasha0 (https://github.com/django-oscar/django-oscar/pull/2591#issuecomment-468375839)?

> What do you think about `CustomerDispatcher` and `OrderDispatcher` for sending notifications, such as `send_order_placed_email_for_user` or `send_password_reset_email_for_user`?

Probably this is the optimal approach to use "special dispatcher" for concrete emails sending - each method that can be easily customised.

solarissmoke: I think this makes sense - would these classes subclass the core Dispatcher or just use it as an object?

solarissmoke:
@sasha0 @samitnuk I want to suggest that we move this to the 2.1 milestone, as I feel like it's going to be quite a big change, and we may want to provide some backward compatibility (i.e., add a new dispatcher that fits our new model, but leave the old one there and deprecate it) - or at the very least we need thorough documentation on how to migrate from the old one to the new one.

If this sounds OK to you then I can look at preparing a 2.1 beta release tomorrow?

sasha0:
> @sasha0 @samitnuk I want to suggest that we move this to the 2.1 milestone, as I feel like it's going to be quite a big change, and we may want to provide some backward compatibility (i.e., add a new dispatcher that fits our new model, but leave the old one there and deprecate it) - or at the very least we need thorough documentation on how to migrate from the old one to the new one.

Yes, it will be fine. I'd ask that we prioritize its review over other our PRs and also define a timeline for the 2.1.

> If this sounds OK to you then I can look at preparing a 2.1 beta release tomorrow?

Yes, we can start to prepare rc1 tomorrow. PS keep in mind that we had bunch of deprecated fields, that to be removed in 2.0.

src/oscar/apps/checkout/mixins.py
samitnuk: I think these classes can use it just as an object.

samitnuk: E.g.
```python
from oscar.core.loading import get_class


Dispatcher = get_class('communication.utils', 'Dispatcher')


class CustomerDispatcher:

    # Event codes
    REGISTRATION_EVENT_CODE = 'REGISTRATION'

    dispatcher = Dispatcher()

    def send_registration_email_for_user(self, user, extra_context):
        messages = self.dispatcher.get_messages(
            self.REGISTRATION_EVENT_CODE, extra_context)
        self.dispatcher.dispatch_user_messages(user, messages)
```

solarissmoke: :+1:

sasha0:
@solarissmoke ready for the review.

vdboor:
Curious, why does this have a 2.1 release notes file included?
Wouldn't it be more useful to have a single "large breaking release" at 2.0 with these things included?

src/oscar/apps/customer/utils.py
```diff
@@ -1,131 +1,136 @@
...
-            CommunicationEvent._default_manager.create(order=order, event_type=event_type)
+    def send_email_changed_email_for_user(self, user, extra_context):
+        messages = self.dispatcher.get_messages(
+            self.EMAIL_CHANGED_EVENT_CODE, extra_context)
+        self.dispatcher.dispatch_user_messages(user, messages)
 
-    def create_customer_email(self, user, messages, email):
-        """
-        Create Email instance in database for logging purposes.
+    def send_product_alert_email_for_user(self, product):  # noqa: C901 too complex
```
craigloftus: This method is copied over from `customer.alerts.utils` but the original remains? I wonder whether for consistency, we should have a `CustomerAlertDispatcher` (or `AlertDispatcher`) for this method to live in?

src/oscar/templates/oscar/communication/emails/commtype_product_alert_confirmation_body.html
```diff
@@ -0,0 +1,19 @@
+{% extends "oscar/communication/emails/base.html" %}
+{% load i18n %}
+
+{% block body %}
+    {% blocktrans with title=alert.product.get_title|safe domain=site.domain confirm_path=alert.get_confirm_url cancel_path=alert.get_cancel_url %}Hello,
+        <p>
+            Somebody (hopefully you) has requested an email alert when '{{ title }}' is back in stock.
+            Please click the following link to confirm: http://{{ domain }}{{ confirm_path }}
```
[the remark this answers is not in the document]
samitnuk: Probably the most simple solution for the moment is to add `BASE_URL` or `DEFAULT_DOMAIN` to `get_base_context` method (https://github.com/django-oscar/django-oscar/pull/2591/files#diff-b3a3dfc0e4b07c771eac76d1fbc6cf2dR161) and update all email templates with it.

E.g.
```python
def get_base_context(self):
    site = Site.objects.get_current()
    return {
        'BASE_URL': 'http://{}'.format(site.domain),
        'site': site,
    }
```

@craigloftus @sasha0 WDYT?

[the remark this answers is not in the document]
craigloftus: @samitnuk I'd suggest as an alternative a simple template tag, so each template doesn't have to reimplement the logic for constructing absolute urls. And I would have that template use site from the template context and a setting like `URL_SCHEMA='https'`.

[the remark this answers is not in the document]
craigloftus: @samitnuk Coincidentally I came across https://github.com/fusionbox/django-absoluteuri being used in a project today, which uses the approach I describe, might make sense to just depend on that?

src/oscar/apps/communication/abstract_models.py
```diff
@@ -0,0 +1,222 @@
...
+        # Ensure the email subject doesn't contain any newlines
+        messages['subject'] = messages['subject'].replace("\n", "")
+        messages['subject'] = messages['subject'].replace("\r", "")
+
+        return messages
+
+    def __str__(self):
+        return self.name
+
+    def is_order_related(self):
```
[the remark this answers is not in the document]
samitnuk: `is_order_related` is used in a couple of places to get a preview of email (on `dashboard/comms/` page):
https://github.com/django-oscar/django-oscar/blob/6fa2834df028f0cade9158e05ffcecacc8678026/src/oscar/apps/dashboard/communications/forms.py#L56-L67

https://github.com/django-oscar/django-oscar/blob/6fa2834df028f0cade9158e05ffcecacc8678026/src/oscar/templates/oscar/dashboard/comms/detail.html#L63-L66

https://github.com/django-oscar/django-oscar/blob/6fa2834df028f0cade9158e05ffcecacc8678026/src/oscar/templates/oscar/dashboard/comms/detail.html#L76-L79

So probably we can keep it for the moment. Or should we re-work it somehow?

src/oscar/templates/oscar/communication/emails/commtype_product_alert_confirmation_body.html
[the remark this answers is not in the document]
samitnuk: Thanks for this! I will add a simple template tag as you offered previously. Probably this is a better option than an additional dependency just for this small feature.

src/oscar/apps/communication/abstract_models.py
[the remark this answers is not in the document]
craigloftus: Okay, I guess keep them for now, and we can rework in a separate PR.

samitnuk:
@solarissmoke @craigloftus thank you for your comments.

I updated PR based on your recommendations. Please review.

sasha0:
@solarissmoke @craigloftus please take a loot when you get chance - I believe we're close to finish PR, so I'd very much appreciate your review/thoughts on it.

src/oscar/apps/communication/abstract_models.py
```diff
@@ -0,0 +1,217 @@
...
+                regex=r'^[a-zA-Z_][0-9a-zA-Z_]*$',
+                message=_(
+                    "Code can only contain the letters a-z, A-Z, digits, "
+                    "and underscores, and can't start with a digit."))],
+        help_text=_("Code used for looking up this event programmatically"))
+
+    #: Name is the friendly description of an event for use in the admin
+    name = models.CharField(
+        _('Name'), max_length=255,
+        help_text=_("This is just used for organisational purposes"))
```
solarissmoke: Suggest we drop the help text here - don't think it's necessary and `organisational` is confusing as it suggests it is related to an Organisation (as opposed to organising things).

src/oscar/apps/communication/abstract_models.py
```diff
@@ -0,0 +1,217 @@
...
+
+    #: Code used for looking up this event programmatically.
+    # e.g. PASSWORD_RESET. AutoSlugField uppercases the code for us because
+    # it's a useful convention that's been enforced in previous Oscar versions
+    code = AutoSlugField(
+        _('Code'), max_length=128, unique=True, populate_from='name',
+        separator=six.u("_"), uppercase=True, editable=True,
+        validators=[
+            RegexValidator(
+                regex=r'^[a-zA-Z_][0-9a-zA-Z_]*$',
```
solarissmoke: In the code that fetches email templates, we normalise the code with `code = self.code.lower()`. This means that if someone has defined two identical codes with different case, things will break. This seems like a bug even with the current code.

Since this is a breaking change I would suggest we enforce uppercase codes (I'd personally prefer lower, but historically Oscar has generated uppercase).

src/oscar/apps/communication/abstract_models.py
```diff
@@ -0,0 +1,217 @@
...
+        return self.category == self.ORDER_RELATED
+
+    def is_user_related(self):
+        return self.category == self.USER_RELATED
+
+
+class AbstractNotification(models.Model):
+    recipient = models.ForeignKey(
+        AUTH_USER_MODEL,
+        db_index=True,
```
solarissmoke: `db_index` is unnecessary on foreign key fields - they are indexed always.

src/oscar/apps/communication/utils.py
```diff
@@ -0,0 +1,175 @@
...
+            self.notify_user(user, subject, **kwargs)
+
+    # Internal
+
+    def create_email(self, user, messages, email):
+        """
+        Create ``Email`` instance in database for logging purposes.
+        """
+        if email and user.is_authenticated:
+            return Email._default_manager.create(
```
solarissmoke: Why are we using `_default_manager` here? To facilitate easy overriding I think we should be using `Email.objects`.

src/oscar/apps/communication/utils.py
```diff
@@ -0,0 +1,175 @@
...
+        Send message to the registered user / customer and collect data in database.
+        """
+        if not user.email:
+            self.logger.warning("Unable to send email messages as user #%d has"
+                                " no email address", user.id)
+            return None
+
+        email = self.send_email_messages(user.email, messages, attachments=attachments)
+
+        if getattr(settings, 'OSCAR_SAVE_SENT_EMAILS_TO_DB', True):
```
solarissmoke: I'd prefer we explicitly define settings in `defaults.py` rather than auto-loading defaults at the places we use them. This way things are less likely to break e.g., if we read the same setting elsewhere.

src/oscar/apps/communication/utils.py
```diff
@@ -0,0 +1,175 @@
...
+            self.create_email(user, messages, email)
+
+        return email
+
+    def send_email_messages(self, recipient_email, messages, from_email=None, attachments=None):
+        """
+        Send email to recipient, HTML attachment optional.
+        """
+        if hasattr(settings, 'OSCAR_FROM_EMAIL'):
+            from_email = settings.OSCAR_FROM_EMAIL
```
solarissmoke: `OSCAR_FROM_EMAIL` exists in `defaults.py` so we don't need this check.

src/oscar/apps/customer/abstract_models.py
```diff
@@ -1,20 +1,15 @@
...
 from django.utils.crypto import get_random_string
 from django.utils.translation import gettext_lazy as _
 
 from oscar.core.compat import AUTH_USER_MODEL
 from oscar.core.loading import get_class
-from oscar.models.fields import AutoSlugField
 
 
-CommunicationTypeManager = get_class('customer.managers', 'CommunicationTypeManager')
+CommunicationTypeManager = get_class('communication.managers', 'CommunicationTypeManager')
```
solarissmoke: Unused import.

src/oscar/apps/customer/alerts/utils.py
```diff
@@ -1,134 +1,127 @@
...
+            mail_connection=mail_connection,
+        )
+
+    def send_alerts(self):
+        """
+        Check all products with active product alerts for
+        availability and send out email alerts when a product is
+        available to buy.
+        """
+        products = Product.objects.filter(productalert__status=ProductAlert.ACTIVE).distinct()
```
solarissmoke: We need to filter on `browsable()` products here.

src/oscar/core/loading.py
```diff
@@ -224,6 +235,15 @@ def get_model(app_label, model_name):
...
     """
+    oscar_moved_model = MOVED_MODELS.get(app_label, None)
+    if oscar_moved_model:
+        if model_name.lower() in oscar_moved_model[1]:
+            original_app_label = app_label
+            app_label = oscar_moved_model[0]
+            warnings.warn(
+                'Model %s has recently moved from %s to the application %s, '
+                'please update your imports.' % (model_name, original_app_label, app_label),
+                DeprecationWarning, stacklevel=2)
```
solarissmoke: We should use `RemovedInOscar22Warning` here (need to define this in `oscar.utils.deprecation`).

tests/functional/customer/test_alert.py
```diff
@@ -27,21 +30,19 @@ def test_can_create_a_stock_alert(self):
         form.submit()
 
         alerts = ProductAlert.objects.filter(user=self.user)
-        self.assertEqual(1, len(alerts))
+        assert len(alerts) == 1
```
solarissmoke: I'd suggest we avoid unnecessary changes like this, or make them in a separate PR - it makes the diff on this PR much bigger, and harder to see what the functional changes are. From what I can see we're just changing to pytest-style assertions here, but the actual logic is unchanged.

src/oscar/apps/customer/migrations/0006_auto_20190430_1736.py
```diff
@@ -0,0 +1,40 @@
...
+        ),
+        migrations.RemoveField(
+            model_name='notification',
+            name='recipient',
+        ),
+        migrations.RemoveField(
+            model_name='notification',
+            name='sender',
+        ),
+        migrations.DeleteModel(
```
solarissmoke: Deleting data is something we need to be careful about. If we're going to do this I think we need to document very clearly that these models are being deleted. Given that many sites may want to migrate data, I actually wonder whether we need to stage this, so that we:

1. Add new models and allow projects to migrate data into them.
2. Deprecate the old models in Oscar 2.1
3. Delete the old models in Oscar 2.2.

Doing everything in one go will make it very difficult for projects to do data migrations.

src/oscar/apps/communication/abstract_models.py
```diff
@@ -0,0 +1,222 @@
...
+from oscar.apps.communication.managers import CommunicationTypeManager
+from oscar.core.compat import AUTH_USER_MODEL
+from oscar.models.fields import AutoSlugField
+
+
+class AbstractEmail(models.Model):
+    """
+    This is a record of all emails sent to a customer.
+    Normally, we only record order-related emails.
+    """
```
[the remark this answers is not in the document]
samitnuk: Updated, but GitHub did not mark this comment as `Outdated`.

src/oscar/apps/communication/abstract_models.py
```diff
@@ -0,0 +1,222 @@
...
+    ORDER_RELATED = 'Order related'
+    USER_RELATED = 'User related'
+    CATEGORY_CHOICES = (
+        (ORDER_RELATED, _('Order related')),
+        (USER_RELATED, _('User related'))
+    )
+
+    category = models.CharField(
+        _('Category'), max_length=255, default=ORDER_RELATED,
+        choices=CATEGORY_CHOICES)
```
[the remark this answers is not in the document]
samitnuk: I think we will remove this field when we drop the `is_order_related` and `is_user_related` methods.

docs/source/ref/templatetags.rst
```diff
@@ -191,3 +191,30 @@ This renders something like:
...
+
+.. code-block:: html+django
+
+    {% absolutize_url site.domain reset_url as absolute_reset_url %}
+    {% blocktrans with url=absolute_reset_url %}
+        You can reset your password here - {{ url }}
+    {% endblocktrans %}
+
+The schema for absolute url can be set through ``OSCAR_URL_SCHEMA``
+setting (``http`` by default).
```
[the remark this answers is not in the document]
samitnuk: Added to `defaults.py`.

tests/functional/customer/test_alert.py
samitnuk: Yes, you are right - the actual logic is unchanged.

Will do changes like this in separate PR in the future. Should I remove them from this PR, or we can keep them?

src/oscar/apps/customer/migrations/0006_auto_20190430_1736.py
samitnuk: In general, we do not delete data here, we just mark these models as *deleted* (with `state_operations` no DB changes will be made).

The same we are doing in the `communications/migrations/0001_initial.py` (https://github.com/django-oscar/django-oscar/pull/2591/files#diff-c64e3a7f51c8fb93665d85a22e77896dR21) - no new DB tables for the `communication` app will be created - we only mark them as created, but the tables from the `customer` app will be used.

Please correct me if I missed something.

solarissmoke: I'd missed that these were state operations - thanks for clarifying!

Please ping me once the rest of the PR is ready for another review.

samitnuk:
@solarissmoke I think this PR is ready for another review 😉

docs/source/releases/v2.1.rst
```diff
@@ -15,6 +15,8 @@ Welcome to Oscar 2.1.
 Compatibility
 ~~~~~~~~~~~~~
 
+Oscar 2.1 is compatible with Django 1.11, Django 2.1 and Django 2.2
```
solarissmoke: Please omit this for now - we'll add in when we cut a release, so that we can make sure it's up to date (Django 2.1 is now end of life and we've dropped support for it).

docs/source/releases/v2.1.rst
```diff
@@ -64,4 +106,8 @@ Dependency changes
 .. _deprecated_features_in_2.1:
 
 Deprecated features
-~~~~~~~~~~~~~~~~~~
+~~~~~~~~~~~~~~~~~~~
+
+- Field ``category`` removed from ``Notification`` model.
```
solarissmoke: Think this needs to do under a heading "Backwards incompatible changes in Oscar 2.1", after the what's new section, as it's not a deprecation but a breaking change.

src/oscar/apps/checkout/mixins.py
```diff
@@ -268,44 +263,15 @@ def send_signal(self, request, response, order):
     def get_success_url(self):
         return reverse('checkout:thank-you')
 
-    def send_confirmation_message(self, order, code, **kwargs):
```
solarissmoke: Think we need to document that this method has been removed and replaced with `send_order_placed_email` - could you please add this to the release notes?

src/oscar/apps/communication/abstract_models.py
```diff
@@ -0,0 +1,213 @@
...
+        null=True)
+    email = models.EmailField(_('Email Address'), null=True, blank=True)
+    subject = models.TextField(_('Subject'), max_length=255)
+    body_text = models.TextField(_("Body Text"))
+    body_html = models.TextField(_("Body HTML"), blank=True)
+    date_sent = models.DateTimeField(_("Date Sent"), auto_now_add=True)
+
+    class Meta:
+        abstract = True
+        app_label = 'communication'
```
solarissmoke: Is it necessary to set `app_label` and `db_table` explicitly? These look the same as the defaults that Django would produce.

src/oscar/apps/communication/abstract_models.py
```diff
@@ -0,0 +1,213 @@
...
+
+    # File templates
+    email_subject_template_file = 'oscar/communication/emails/commtype_%s_subject.txt'
+    email_body_template_file = 'oscar/communication/emails/commtype_%s_body.txt'
+    email_body_html_template_file = 'oscar/communication/emails/commtype_%s_body.html'
+    sms_template_file = 'oscar/communication/sms/commtype_%s_body.txt'
+
+    class Meta:
+        abstract = True
+        app_label = 'communication'
```
solarissmoke: As above - please check if we need to specify this explicitly.

src/oscar/apps/customer/alerts/utils.py
```diff
@@ -1,134 +1,126 @@
...
-    }
-
-    code = 'PRODUCT_ALERT_CONFIRMATION'
-    messages = CommunicationEventType.objects.get_and_render(code, ctx)
-
-    if messages and messages['body']:
-        Dispatcher().dispatch_direct_messages(alert.email, messages)
-
-
-def send_product_alerts(product):   # noqa C901 too complex
```
solarissmoke: Can we make this backwards compatible by leaving it here for now, adding a deprecation warning and then calling `AlertsDispatcher().send_product_alert_email_for_user(instance.product)`?

src/oscar/apps/customer/notifications/services.py
```diff
@@ -1,18 +0,0 @@
-from oscar.core.loading import get_model
-
-Notification = get_model('customer', 'Notification')
-
-
-def notify_user(user, subject, **kwargs):
```
solarissmoke: Can we deprecate these instead of removing, and change them to call the new dispatcher methods?

docs/source/releases/v2.1.rst
```diff
@@ -31,6 +33,46 @@ What's new in Oscar 2.1?
...
+  moved to ``communication/email/*.html`` and ``communication/notification/*.html``.
+
+- Models ``CommunicationEventType``, ``Email`` and ``Notification`` added to
+  ``MOVED_MODELS`` to show warning when they will be imported from ``customer`` app.
+  E.g. ``Email = get_model('customer', 'Email')``.
+
+- Added ``absolute_url`` template tag - to get absolute URLs in templates based on
+  provided path and domain. The schema for absolute url can be set through
+  ``OSCAR_URL_SCHEMA`` setting, ``http`` by default.
+
```
solarissmoke: I've made a couple of comments in the code, but would be grateful if you can add to these notes:

1. All methods that have been removed, and what they have been replaced with.
2. All context processors that have been removed, and what they have been replaced with.

samitnuk: Added to this file.

src/oscar/apps/checkout/mixins.py
samitnuk: Added to `docs/source/releases/v2.1.rst`

src/oscar/apps/communication/abstract_models.py
samitnuk: I removed `db_table`s. `app_labels` kept because of
https://github.com/django-oscar/django-oscar/blob/1b76597f8e6d11cd047c26b6235da556afd482ac/tests/integration/core/test_loading.py#L132

src/oscar/apps/customer/alerts/utils.py
samitnuk: Added for all methods in `src/oscar/apps/customer/alerts/utils.py`:
* `send_alerts`
* `send_alert_confirmation`
* `send_product_alerts`

src/oscar/apps/customer/notifications/services.py
samitnuk: Deprecation added for all methods in `src/oscar/apps/customer/notifications/services.py`:
* `notify_user`
* `notify_users`

samitnuk:
@solarissmoke Many thanks! 

This PR updated based on your comments. Please review when you will have a chance.

solarissmoke:
Thanks @samitnuk. I think this is ready will rebase and merge in #3260.

PR #2592, opened 2018-02-02
Drop Django 1.10 from test matrix.

solarissmoke:
Django 1.10 reached end of life in December 2017.


sasha0:
As discussed, we shall drop support of Django 1.8 and 1.9 as well.

solarissmoke:
In that case there is a lot more we can drop in terms of compatibility code - I will update over the weekend.

solarissmoke:
Closing in favour of #2508.

PR #2595, opened 2018-02-02
Make sure all major imports are dynamic.

sasha0:


src/oscar/apps/catalogue/reviews/abstract_models.py
```diff
@@ -6,11 +6,13 @@
...
 from django.utils.translation import pgettext_lazy
 
-from oscar.apps.catalogue.reviews.managers import ProductReviewQuerySet
-from oscar.apps.catalogue.reviews.utils import get_default_review_status
 from oscar.core import validators
+from oscar.core.loading import get_class
 from oscar.core.compat import AUTH_USER_MODEL, user_is_authenticated
 
+ProductReviewQuerySet = get_class('catalogue.reviews.managers', 'ProductReviewQuerySet')
+get_default_review_status = get_class('catalogue.reviews.utils', 'get_default_review_status')
```
solarissmoke: This isn't actually a class - I don't think we should make loading of methods dynamic. If someone needs to override this default they should override the model field that uses it I think?

src/oscar/apps/checkout/forms.py
```diff
@@ -2,14 +2,14 @@
...
-from oscar.apps.customer.utils import normalise_email
 from oscar.core.compat import get_user_model
-from oscar.core.loading import get_model
+from oscar.core.loading import get_class, get_model
 from oscar.forms.mixins import PhoneNumberMixin
 
 User = get_user_model()
+AbstractAddressForm = get_class('address.forms', 'AbstractAddressForm')
 Country = get_model('address', 'Country')
+normalise_email = get_class('customer.utils', 'normalise_email')
```
solarissmoke: As above I don't think we should make methods overridable like this - better to override the field that uses it.

sasha0:
> A lot of these changes are making core methods (rather than classes) dynamically loaded. I don't know why but this makes me very uneasy. Oscar's class loading is somewhat magical already, and this will make it even more so. I think we need to be moving towards more explicit overriding where possible (e.g., when we switch to use Django's appconf), rather than dynamic overriding.

Currently most of the imports are already dynamic, so I don't see bigger impact to make it completely consistent and load _all_ application classes dynamically, which I believe is quite appropriate until we do refactoring and switch to appconf.

sasha0:
> I don't think get_class should be used for anything that isn't actually a class.

Ok, maybe loading functions is an overhead (except offer utils). Updated PR.

solarissmoke:
This looks OK to me, with the caveat that I think we need to move to a more explicit overriding scheme in the future.

src/oscar/apps/offer/abstract_models.py
```diff
@@ -16,14 +16,15 @@
...
-from oscar.core.loading import get_class, get_model
+from oscar.core.loading import get_class, get_classes, get_model
 from oscar.models import fields
 from oscar.templatetags.currency_filters import currency
 
-BrowsableRangeManager = get_class('offer.managers', 'BrowsableRangeManager')
+ActiveOfferManager, BrowsableRangeManager \
+    = get_classes('offer.managers', ['ActiveOfferManager', 'BrowsableRangeManager'])
+ZERO_DISCOUNT = get_class('offer.results', 'ZERO_DISCOUNT')
+load_proxy, unit_price = get_classes('offer.utils', ['load_proxy', 'unit_price'])
```
craigloftus: I have no problem with allowing the overriding of these functions in principle, but I think we need to be adding `get_functions` to do it?

craigloftus: Just the name. It would make it explicit what type of object is going to be returned. We could consider `get_object` and `get_objects` as alternatives for use with functions and classes.

sasha0: if that's only the name - I don't see the reason to rename it simply because it's used everywhere through the application to load classes. Also, previously part of the signals where loaded by this method before I refactored them to load directly.

I agree about the necessity to refactor app loading and switch to Django Appconf, but for now this PR will allow to finalize the original idea to make all major imports dynamic for consistency and maximum customizability.

PR #2601, opened 2018-02-08
Set maximum line quantity on basket page.

### LINKED ISSUE #1412: Set maximum quantity on basket page
Extracted from #1405, we should try and set the `max` attribute on the basket line form. I think this will require a new method on the availability policy class to fetch the maximum permitted quantity.

sasha0:
Fixes #1412

### DIFF
--- a/src/oscar/apps/basket/abstract_models.py
+++ b/src/oscar/apps/basket/abstract_models.py
@@ -135,21 +135,29 @@ class AbstractBasket(models.Model):
                 .order_by(self._meta.pk.name))
         return self._lines
 
+    def max_allowed_quantity(self):
+        """
+        Returns maximum product quantity, that can be added to the basket
+        with the respect to basket quantity threshold.
+        """
+        basket_threshold = settings.OSCAR_MAX_BASKET_QUANTITY_THRESHOLD
+        if basket_threshold:
+            total_basket_quantity = self.num_items
+            max_allowed = basket_threshold - total_basket_quantity
+            return max_allowed, basket_threshold
+
     def is_quantity_allowed(self, qty):
         """
         Test whether the passed quantity of items can be added to the basket
         """
         # We enforce a max threshold to prevent a DOS attack via the offers
         # system.
-        basket_threshold = settings.OSCAR_MAX_BASKET_QUANTITY_THRESHOLD
-        if basket_threshold:
-            total_basket_quantity = self.num_items
-            max_allowed = basket_threshold - total_basket_quantity
-            if qty > max_allowed:
-                return False, _(
-                    "Due to technical limitations we are not able "
-                    "to ship more than %(threshold)d items in one order.") \
-                    % {'threshold': basket_threshold}
+        max_allowed, basket_threshold = self.max_allowed_quantity()
+        if max_allowed is not None and qty > max_allowed:
+            return False, _(
+                "Due to technical limitations we are not able "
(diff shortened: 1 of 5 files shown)


src/oscar/apps/basket/abstract_models.py
```diff
@@ -135,21 +135,26 @@ def all_lines(self):
                 .order_by(self._meta.pk.name))
         return self._lines
 
+    def max_allowed_quantity(self):
+        basket_threshold = settings.OSCAR_MAX_BASKET_QUANTITY_THRESHOLD
```
solarissmoke: Please add a docstring explaining what this method is doing.

src/oscar/apps/basket/forms.py
```diff
@@ -19,6 +19,10 @@ def __init__(self, strategy, *args, **kwargs):
         super(BasketLineForm, self).__init__(*args, **kwargs)
         self.instance.strategy = strategy
 
+        num_available = getattr(self.instance.purchase_info.availability, 'num_available', None)
+        max_allowed_quantity = min(num_available, self.instance.basket.max_allowed_quantity()[0])
```
solarissmoke: This will give an incorrect result (`None`) if `num_available` is `None`. Is there a reason you're using `getattr` above? The availability class is required to set `num_available` always.

sasha0: The reason is that num_available required only by StockRequired availability class, others don't require this argument.

solarissmoke: But the result of `min(None, 6)` will be `None`, meaning `max_allowed_quantity` will be set to `None`, which I don't think is what you want?

src/oscar/apps/basket/forms.py
```diff
@@ -19,6 +19,16 @@ def __init__(self, strategy, *args, **kwargs):
         super(BasketLineForm, self).__init__(*args, **kwargs)
         self.instance.strategy = strategy
 
+        max_allowed_quantity = None
+        num_available = getattr(self.instance.purchase_info.availability, 'num_available', None)
+        basket_max_allowed_quantity = self.instance.basket.max_allowed_quantity()[0]
+        if all([num_available, basket_max_allowed_quantity]):
+            max_allowed_quantity = min(num_available, basket_max_allowed_quantity)
+        elif any([num_available, basket_max_allowed_quantity]):
```
solarissmoke: This can be simplified to just `else:` - the `any` check is redundant.

src/oscar/apps/basket/abstract_models.py
```diff
@@ -135,21 +135,29 @@ def all_lines(self):
                 .order_by(self._meta.pk.name))
         return self._lines
 
-    def is_quantity_allowed(self, qty):
+    def max_allowed_quantity(self):
         """
-        Test whether the passed quantity of items can be added to the basket
+        Returns maximum product quantity, that can we added to the basket
```
solarissmoke: Typo - "that can we added" -> "that can be added".

src/oscar/apps/basket/forms.py
sasha0: In `else` condition they both could be `None`, so I believe `any` check is necessary.

solarissmoke: No - because you are doing a second `if` check below. It will not add an attr if both are `None`.

PR #2603, opened 2018-02-09
Update date/time picker widgets to use template-based rendering.

solarissmoke:
This allows us to reuse a single template for all three widgets (date, time, datetime).


sasha0:
The only pending thing - release notes update.

solarissmoke:
I've added to the release notes. Also removed a duplicated line that must have appeared during merge.

PR #2606, opened 2018-02-09
use elasticsearch for search app and make search app reusable for other models

v1kku:
This is a WIP branch submitted for review and comments on the approach used to implement reusable search. Only elasticsearch 5.X is supported at the moment. Older versions will not be supported (and will not work at all).

Documentation and more tests will be added.


SalahAdDin:
Will you deprecate `sorl` backend support?

solarissmoke:
It's probably worth describing in more detail what this patch is doing:

1. Drop django-haystack completely.
2. Create a new search backend that is powered by Elasticsearch. It does not support Solr or any other search engine (but you could write your own that exposes the same basic API). Using a single search engine allows us to use its features (e.g., faceting and aggregations) to the fullest extent, rather than lowest-common-denominator which is what we ended up with when using Haystack.
3. Provide support for indexing of *any* document, not just products.
4. For the catalogue, provide a number of helpful utilities and classes, including:
     a) Built-in support for indexing and faceting of attributes - any attribute for which a facet is defined in `OSCAR_SEARCH['FACETS']` will be automatically indexed and facets generated during search.
     b) Dynamic price range faceting
     c) Category filtering
     d) Auto-suggest querying, filterable by category
     e) Top and related products querying, filterable by category

The filtering and faceting capability that this introduces is significantly more powerful and flexible than what is currently possible with Haystack, and is the main motivation for this change.

SalahAdDin:
I think it's the right way.

sasha0:
Please rebase remote branch and resolve conflicts. Thanks.

src/oscar/defaults.py
```diff
@@ -208,43 +206,48 @@
...
+    ],
+    "INDEX_CONFIG": {
+        "index.requests.cache.enable": True,
+        "number_of_shards": 1,
+        "number_of_replicas": 0
+    },
+    "INDEX_NAME": "oscar",
+    "INDEXED_FIELDS": {},
+    "SHOW_PRICE_RANGE_FACET": False,
+    "PRODUCTS": {
```
craigloftus: I think we need a structure that makes it clearer what the `PRODUCTS`, `RELATED_PRODUCTS`, `TOP_PRODUCTS` 'are'. Perhaps these options would be better defined on the `Search` subclass definitions? We could consider copying Django's `class Meta` pattern?

src/oscar/apps/search/registries.py
```diff
@@ -0,0 +1,24 @@
...
+
+from django.conf import settings
+
+
+class AnalyzerRegistry(object):
+
+    def __init__(self):
+        self.analyzers = {}
+
+    def define_in_index(self, index):
```
craigloftus: What problem is this trying to solve? Is this to allow adding analyzers to existing indexes?

src/oscar/defaults.py
```diff
@@ -208,43 +206,48 @@
...
+        "oscar.apps.search.analyzers.edgengram_analyzer"
+    ],
+    "INDEX_CONFIG": {
+        "index.requests.cache.enable": True,
+        "number_of_shards": 1,
+        "number_of_replicas": 0
+    },
+    "INDEX_NAME": "oscar",
+    "INDEXED_FIELDS": {},
+    "SHOW_PRICE_RANGE_FACET": False,
```
craigloftus: This seems like a search specific setting rather than a high level setting?

src/oscar/apps/search/management/commands/update_oscar_index.py
```diff
@@ -0,0 +1,16 @@
+from django.core.management.base import BaseCommand
+
+from django_elasticsearch_dsl.registries import registry
+
+
+class Command(BaseCommand):
+    help = "Completely rebuilds the search index by removing the old data and then updating."
```
craigloftus: Help text doesn't reflect what the command does.

src/oscar/apps/search/management/commands/update_oscar_index.py
```diff
@@ -0,0 +1,16 @@
...
+
+    def handle(self, **options):
+        for Document in registry.get_documents():
```
craigloftus: I think we DRY these two management commands out.

craigloftus:
Closing in favour of #2734

src/oscar/apps/catalogue/documents.py
```diff
@@ -0,0 +1,226 @@
...
+
+class ProductDocumentMeta(DocTypeMeta):
+
+    def __new__(cls, name, bases, attrs):
+        attrs['product_attributes'] = []
+
+        try:
+            attribute_configs = settings.OSCAR_SEARCH['INDEXED_FIELDS'].get('product', {})
+            indexed_attributes = ProductAttribute.objects.filter(code__in=attribute_configs.keys())
+
```
specialunderwear: It seems that this construct hints at an incompatibility between elasticsearch-dsl and oscar's datascructure.
If we need to make a query to create new types/classes - which are later converted **back** to json (elasticsearch query language) - it would follow it is better to _skip_ the creation of types entirely and revert back to using json directly. Looks like the extra abstraction layer elasticsearch-dsl introduces is more of a burden than a benefit.

PR #2607, opened 2018-02-10
Update ImageInput to use built-in template rendering and add tests for the widget

solarissmoke:


src/oscar/forms/widgets.py
```diff
@@ -20,36 +19,25 @@ class ImageInput(FileInput):
...
-            'image_url': image_url,
-            'image_id': "%s-image" % final_attrs['id'],
-        })
+        ctx['image_url'] = ''
+        if value and not isinstance(value, InMemoryUploadedFile):
+            # can't display images that aren't stored - pass empty string to context
+            ctx['image_url'] = value
+
+        image_id = ctx['widget']['attrs'].get('id')
+        ctx['image_id'] = "%s-image" % image_id if image_id else None
```
sasha0: How would it formatted in case if `image_id` is `None`?

solarissmoke: There should never be a case when the ID isn't set, especially with the way we use this input, so I've updated the code to always expect an ID.

PR #2609, opened 2018-02-10
Remove unnecessary webob dependency from oscar.test utils.

### LINKED ISSUE #2547: webob in use, but not referenced in pip dependencies
Hello,
I update django-oscar 1.4 to 1.5.1. webob is now used in /oscar/test/utils.py:6. But webob is not added to python dependencies, so it's not added to requirements with pip-compile.

solarissmoke:
The logic we were performing in `extract_cookie_value` is unnecessary - where we need to unquote cookie values we can just use `http.cookies._unquote` as Django itself does.

Fixes #2547.

### DIFF
--- a/src/oscar/test/utils.py
+++ b/src/oscar/test/utils.py
@@ -3,8 +3,6 @@ from django.contrib.messages.storage.fallback import FallbackStorage
 from django.contrib.sessions.backends.db import SessionStore
 from django.core.signing import Signer
 from django.test import RequestFactory as BaseRequestFactory
-from webob.compat import bytes_
-from webob.cookies import _unquote
 
 from oscar.core.loading import get_class, get_model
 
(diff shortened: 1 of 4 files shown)


src/oscar/test/utils.py
```diff
@@ -3,11 +3,15 @@
...
-from webob.compat import bytes_
-from webob.cookies import _unquote
 
 from oscar.core.loading import get_class, get_model
 
+try:
+    # Python 3
+    from http.cookies import _unquote
+except ImportError:
+    from Cookie import _unquote
```
sasha0: Does that make sense to move to compat?

src/oscar/test/utils.py
```diff
@@ -30,11 +34,5 @@ def request(self, user=None, basket=None, **request):
...
-def extract_cookie_value(response_cookies, cookie_name):
-    """
-    Making sure cookie unescaped from double quotes when extracting from
-    test response using Webob approach of cookie parsing.
-    """
-    oscar_open_basket_cookie = bytes_(response_cookies[cookie_name])
-    oscar_open_basket_cookie = _unquote(oscar_open_basket_cookie)
-    return oscar_open_basket_cookie.decode('utf-8')
+def unquote_cookie(cookie_value):
+    return _unquote(cookie_value)
```
sasha0: Docstring would be still useful.

src/oscar/test/utils.py
solarissmoke: Yep, have moved to compat and added a docstring.

PR #2614, opened 2018-02-12
Add test for category URL caching.

sasha0:


setup.py
```diff
@@ -67,6 +67,7 @@
     'pytest-django==3.1.2',
     'pytest-xdist==1.20.0',
     'tox==2.8.2',
+    'python-memcached==1.59'
```
solarissmoke: We should not be adding this to the setup.py dependencies. Projects should be able to choose which cache backend to use.

tests/integration/catalogue/test_category.py
```diff
@@ -44,6 +45,13 @@ def test_unicode_slug(self):
             self.assertEqual(root_category.slug, u'vins-français')
             self.assertEqual(child_category.slug, u'château-dyquem')
 
+    def test_url_caching(self):
```
solarissmoke: Do we need to have memcached running to test this? Should be possible and more specific just to mock the `cache.set()` call and check that the right data is being saved to the cache.

setup.py
sasha0: This is similar to Postgresql client - it's by default included into the tests requirements, although users may need to choose another one, but we use it to run builds by default since it is frequently used. Memcached is also recommended cache backend and is similarly important part of stack.

tests/integration/catalogue/test_category.py
sasha0: Yes, because default cache backend does not serialize and deserialize data as well as don't write it to the storage. One of of our projects, category slug, deserialized from the cache, has `bytes` type, not `str`, which is intended to verify by this change.

solarissmoke: Can't we use the default (locmem) or file based cache then? 

The reason I don't think adding memcached to the test dependencies is advisable is that memchached requires additional system-level libraries (python-dev, libmemcached-devel etc) which will make life that much harder for anyone who wants to set it up locally.

sasha0:
@solarissmoke ok I've finalized it finally.

PR #2616, opened 2018-02-12
Issues/1611/variant product page

sasha0:
Refs #1611


src/oscar/templates/oscar/catalogue/detail.html
```diff
@@ -28,7 +28,7 @@
             <a href="{{ c.get_absolute_url }}">{{ c.name }}</a>
         </li>
         {% endfor %}
-        <li class="active">{{ product.title }}</li>
+        <li class="active">{{ product.get_title }} {% if product.is_child and product.attribute_summary %}({{ product.attribute_summary }}){% endif %}</li>
```
solarissmoke: As noted in the previous PR, this will become very long for any product that has more than 2-3 attributes defined (which can happen quite easily). I don't think it is a good default. Same for the product title/heading.

I think the attribute summary needs to be displayed separately from the title. The child product title can be edited separately anyway, so projects can control what is displayed there.

src/oscar/templates/oscar/catalogue/detail.html
```diff
@@ -78,9 +78,19 @@ <h1>{{ product.get_title }}</h1>
...
 
-            {% block product_basket_form %}
-                {% include "catalogue/partials/add_to_basket_form.html" %}
-            {% endblock %}
+            {% if not product.is_parent %}
+                {% block product_basket_form %}
+                    {% include "catalogue/partials/add_to_basket_form.html" %}
+                {% endblock %}
+            {% else %}
+                <b>{% trans 'Variants:' %}</b><br>
```
solarissmoke: I don't think "variants" is a word we should be using with the end user. Not sure what to use instead. I also think it depends on how a project is using variants. I suggest we get wider input on whether projects would benefit from this before applying this change.

src/oscar/apps/catalogue/views.py
```diff
@@ -28,8 +28,10 @@ class ProductDetailView(DetailView):
     # Whether to redirect to the URL with the right path
     enforce_paths = True
 
-    # Whether to redirect child products to their parent's URL
-    enforce_parent = True
```
solarissmoke: This will break compatibility - I think it would be better not to change it unless there is a very good justification for the original default no longer being appropriate?

If we do make the change then the [documentation](http://django-oscar.readthedocs.io/en/releases-1.5/topics/modelling_your_catalogue.html#parent-and-child-products) needs updating.

src/oscar/templates/oscar/catalogue/detail.html
sasha0: > I don't think "variants" is a word we should be using with the end user. Not sure what to use instead.

We mostly use term "variant" for child product, so I don't think we have to invent anything here.

> I also think it depends on how a project is using variants.

We can wrap this snippet into the template block, so that it can be easily customized. Also, users have opportunity to change translation of the string to necessary word/phrase.

src/oscar/apps/catalogue/views.py
sasha0: > This will break compatibility - I think it would be better not to change it unless there is a very good justification for the original default no longer being appropriate?

If we don't change value of the property, change on the product detail page (list of variants with absolute URLs to their pages), which replaced variant selection from the dropdown, will not work. I also don't find redirection on variant selection in the dropdown best solution, although it could allow to preserve backwards compatibility.

src/oscar/templates/oscar/catalogue/detail.html
sasha0: > I think the attribute summary needs to be displayed separately from the title. The child product title can be edited separately anyway, so projects can control what is displayed there.

I think we can display attribute summary under the title, so that we can easily identify what variant is it. What do you think?

src/oscar/apps/catalogue/views.py
sasha0: Or we can introduce it within other breaking changes in 2.0. WDYT?

sasha0: Again, we can properly display variant-specific field values only on separate page, thus default property update looks inevitable. Those who decide to update data within the variant selection, could override this property value in the project.

docs/source/releases/v1.6.rst
```diff
@@ -112,6 +113,13 @@ Minor changes
 
  - Fixed ``PhoneNumberMixin``s handling of empty phone number fields in forms.
 
+ - Redirection to parent product detail page was disabled for variant products,
```
solarissmoke: These need to be moved to 2.0 release notes.

src/oscar/templates/oscar/catalogue/detail.html
solarissmoke: Still don't think this is a good idea. Just try creating a product with 6 or 7 attributes, and this will not render well at all.

src/oscar/templates/oscar/catalogue/detail.html
```diff
@@ -66,7 +66,7 @@
             {% if user.is_staff %}
                 <a class="pull-right hidden-xs" href="{% url 'dashboard:catalogue-product' pk=product.id %}"><small><i class="icon-edit"></i> {% trans "Edit this product" %}</small></a>
             {% endif %}
-            <h1>{{ product.get_title }}</h1>
+            <h1>{{ product.get_title }} {% if product.is_child and product.attribute_summary %}({{ product.attribute_summary }}){% endif %}</h1>
```
solarissmoke: As above I don't think attribute summary can be used in titles, because you cannot control its length.

src/oscar/templates/oscar/catalogue/detail.html
solarissmoke: Would prefer to drop the `b` tags and use a `h2` instead, and also wrap this in a block.

docs/source/releases/v2.0.rst
```diff
@@ -132,6 +132,13 @@ Minor changes
 - Added a ``datetime_filters`` tag library that provides a ``timedelta`` template
   filter for rendering time deltas in human readable format.
 
+- Redirection to parent product detail page was disabled for variant products,
```
solarissmoke: I think this needs to move to the "backward incompatible changes" section, rather than "minor changes", because it can break behaviour for existing projects.

sasha0: Indeed.

Issue #1408, opened 2014-07-11
Make Oscar agnostic of thumbnailing solution


maikhoepfel:
Oscar really shouldn't care how thumbnails are generated. It [leads to trouble](https://groups.google.com/d/msgid/django-oscar/673527cf-81f6-4901-9176-279992454ae3%40googlegroups.com?utm_medium=email&utm_source=footer) when integrating e.g. django CMS as they use `easy-thumbnails`. There's also solutions like [thumbor](https://github.com/thumbor/thumbor) and [resrc.it](http://www.resrc.it/) better suited for handling thumbnails when using remote storages like S3.

I'm imagining a backend-agnostic `oscar_thumb` template tag that is easy to customise. 
As far as I'm aware, the only advanced usage of the `thumbnail` tag is `crop=centre` for the gallery thumbs, which surely could be made to work somehow.

Questions remaining:
- [ ] Should we drop Sorl as a required dependency? That means PIL wouldn't get installed by default, which makes the final environment lighter. But it probably doesn't provide for a good out-of-the-box experience.

creimers:
@uKev Can I contact you about your djangoCMS - Oscar integration experience? If so, how?

maikhoepfel:
Please do! m@maikhoepfel.de.

maikhoepfel:
Pinging @JostCrow because he's working on this. \o/

sasha0:
Looks like @JostCrow already did some implementation, but not sure if that was finished. Looking forward to see it!

JostCrow:
Sorry for the late response. It has been a little bit busy.

I am still working on a good solution. I have indeed started on creating a solution. Feel free to give any feedback if you have some.

alfaro28:
Any updates on this issue?, I recently had trouble because my project was already using easy_thumbnails

Bogh:
:+1:

Bogh:
I've looked over @JostCrow implementation and I think it's not a good idea to duplicate the templates. Give that there are quite a few. I was thinking a template tag that returns the thumbnail of the image and wraps both libraries and any future that wants to be used. 

Allowing for a setting that will specify the THUMBNAIL_FUNCTION to be called and create and return the thumb of the image.

I will start working in this direction and I'm open to suggestions.

sephii:
@Bogh any update on this? Have you seen this issue? https://github.com/mariocesar/sorl-thumbnail/issues/480

sephii:
[...]

PR #2621, opened 2018-02-20
Configurable thumbnail backends.

sasha0:
Refs #1408

### DIFF
--- a/src/oscar/__init__.py
+++ b/src/oscar/__init__.py
@@ -71,7 +71,6 @@ INSTALLED_APPS = [
     'widget_tweaks',
     'haystack',
     'treebeard',
-    'sorl.thumbnail',
     'django_tables2',
 ]
 
--- a/src/oscar/apps/catalogue/receivers.py
+++ b/src/oscar/apps/catalogue/receivers.py
@@ -3,34 +3,28 @@
 from django.conf import settings
 
 if settings.OSCAR_DELETE_IMAGE_FILES:
-
-    from oscar.core.loading import get_model
-
     from django.db import models
     from django.db.models.signals import post_delete
 
-    from sorl import thumbnail
-    from sorl.thumbnail.helpers import ThumbnailError
+    from oscar.core.loading import get_model
+    from oscar.core.thumbnails import get_thumbnailer
 
     ProductImage = get_model('catalogue', 'ProductImage')
     Category = get_model('catalogue', 'Category')
 
     def delete_image_files(sender, instance, **kwargs):
         """
-        Deletes the original image, created thumbnails, and any entries
-        in sorl's key-value store.
+        Deletes the original image and created thumbnails.
         """
-        image_fields = (models.ImageField, thumbnail.ImageField)
+        image_fields = (models.ImageField,)
+        thumbnailer = get_thumbnailer()
         for field in instance._meta.fields:
             if isinstance(field, image_fields):
                 # Make Django return ImageFieldFile instead of ImageField
-                fieldfile = getattr(instance, field.name)
-                try:
-                    thumbnail.delete(fieldfile)
-                except ThumbnailError:
-                    pass
+                field_file = getattr(instance, field.name)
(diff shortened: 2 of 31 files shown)


sasha0:
We already outlined abstract thumbnail backend functionality and while we are working on the tests, I'd appreciate feedback on the draft version of the functionality.

samitnuk:
Changelog

#### General
- [x] Thumbnailers moved from `install_requires` to `test_requires` in setup.py.
- [x] In `primary_image` method of `AbstractProduct` returned `MissingProductImage().name` instead of just `MissingProductImage()`. In this case "missed image" will be properly used by both thumbnailers.
- [x] Created `thumbnails` module with thumbnailers related classes in `core`.
- [x] Created `oscar_thumbnail` template tag in `image_tags`.
- [x] Templates modified to use `oscar_thumbnail` tag


#### Sandbox
- [x] `easy_thumbnails` added to installed apps in `sandbox` settings.

#### Tests
- [x] `easy_thumbnails` added to installed apps in test settings.
- [x] Test settings updated with `MEDIA` settings.
- [x] Added `ProductImageFactory` to `catalogue` factories.
- [x] Added tests for thumbnailer classes (creation/deletion).
- [ ] Added test to check thumbnails were deleted after image deletion (on product/category delete).
- [ ] Added test to check thumbnails in templates (in common and as context value).

sasha0:
Please rebase remote branch.

samitnuk:
@sasha0
Rebased.

sasha0:
We finished initial implementation and would appreciate review on this, so that we can finalize and add more tests and documentation @django-oscar/core @django-oscar/developers

vdboor:
From a sorl-thumbnail perspective this looks really good. I can't tell though how the options relate to the easy-thumbnails settings.

src/oscar/core/thumbnails.py
```diff
@@ -0,0 +1,66 @@
...
+        opts['size'] = (width, height)
+        return get_thumbnailer(source).get_thumbnail(opts)
+
+    def delete_thumbnails(self, source):
+        from easy_thumbnails.files import get_thumbnailer
+        get_thumbnailer(source).delete()
+
+
+def get_thumbnailer():
+    path = settings.OSCAR_THUMBNAILER.split('.')
```
solarissmoke: Please use Django's `import_string` utility here: https://docs.djangoproject.com/en/2.2/ref/utils/#django.utils.module_loading.import_string

src/oscar/__init__.py
```diff
@@ -71,7 +71,6 @@ def get_version():
     'widget_tweaks',
     'haystack',
     'treebeard',
-    'sorl.thumbnail',
```
solarissmoke: I think this needs to move into `extras_require` so that you can install will `pip install django-oscar[sorl]`. Same for `easy_thumbnails`.

src/oscar/apps/catalogue/receivers.py
```diff
@@ -3,34 +3,28 @@
...
-                    pass
+                field_file = getattr(instance, field.name)
+                thumbnailer.delete_thumbnails(field_file)
 
-    # connect for all models with ImageFields - add as needed
+    # Connect for all models with ImageFields - add as needed
     models_with_images = [ProductImage, Category]
     for sender in models_with_images:
-        post_delete.connect(delete_image_files, sender=sender)
+        pre_delete.connect(delete_image_files, sender=sender)
```
solarissmoke: What is the reason for changing this to `pre_delete`?

samitnuk: Returned to `post_delete` with adding `save=False` to `delete` method of thumbnailer (`easy-thumbnails`).

src/oscar/core/thumbnails.py
samitnuk: Thanks. Updated.

src/oscar/__init__.py
samitnuk: Thanks. Done.

Issue #2215, opened 2017-01-31
Extract oscar.apps.promotions into a separate package


sephii:
Providing the flatpages features makes it more difficult to integrate Oscar in a CMS since it causes a duplication of features and forces to enable a potentially useless app for Oscar to work (flatpages), as well as to hack around to hide this feature from the dashboard.

I think it would be better to provide this feature as a separate package.

mvantellingen:
Yup agreed. The combination of for example wagtail and oscar is really nice, see https://github.com/LabD/django-oscar-wagtail

mvantellingen:
BTW, you are talking about the promotions app right?

sephii:
Exactly.

sasha0:
We can extract promotion app into `django-oscar-promotion` package, but might require promotion models refactoring, attempted in PR #1036 and #1567.

sasha0:
If we move promotion app out of the Oscar, what would we use for the home page? Mb redirection to the catalogue index?

khink:
Disclaimer: I have worked on only 2 Oscar sites so far. There might be people out there using this feature, it would be good to hear from them. 

Also, @sephii 's could you update this issue's title? Flatpages appear to be a django.contrib thing but you mean Promotions (oscar.apps.promotions).

That said, we use Wagtail (also via django-oscar-wagtail) in the one project where we have CMS stuff, and i think we would do so again the next time around. In conclusion: to me, the proposal to remove flatpages makes sense.

And +1 on a redirect to the catalogue index.

solarissmoke:
I think we are mixing up issues here. The original issue mentions Oscar's dependency on `django.contrib.flatpages`. We allow editing of flatpages in the dashboard, and these are used in a few [checkout](https://github.com/django-oscar/django-oscar/blob/6ce56ccbc033341a72838b0cb06935d6ba498c75/src/oscar/templates/oscar/partials/nav_checkout.html) [templates ](https://github.com/django-oscar/django-oscar/blob/6ce56ccbc033341a72838b0cb06935d6ba498c75/src/oscar/templates/oscar/checkout/layout.html).

As far as I can see there is no long between `oscar.promotions` and `flatpages`.

That being the case there are two distinct questions here:

1. Do we drop the requirement for `django.contrib.flatpages` and remove flatpage editing from the dashboard? There is

2. Do we drop the promotions app?

Both of these would be breaking changes so it would be good to get a sense of how many projects rely on either feature at present.

[...]

PR #2623, opened 2018-02-22
Extract promotions from the Oscar core.

sasha0:
Refs #2215


src/oscar/apps/dashboard/app.py
```diff
@@ -54,6 +52,9 @@ def get_urls(self):
             }, name='logout'),
 
         ]
+        if is_promotions_enabled():
+            promotions_app = get_class('dashboard.promotions.app', 'application')
```
vdboor: How about `self.promotions_app or get_class(..)` for backwards compatibility with possible overrides?

sasha0: I'd probably check this case with the test.

solarissmoke:
This looks OK to me, but a couple of things to consider:

1. It is hard to override the home page view - you have to override the whole `Shop` application and then replace the URL pattern for the home page, which isn't that straightforward. I think a lot of projects will want to replace that view, so we should provide an easier way to do it - not sure what the best approach is but maybe a setting that defines which view class to use.

2. Think we need to have https://github.com/django-oscar/django-oscar-promotions populated before we can merge this?

sasha0:
> It is hard to override the home page view - you have to override the whole Shop application and then replace the URL pattern for the home page, which isn't that straightforward. I think a lot of projects will want to replace that view, so we should provide an easier way to do it - not sure what the best approach is but maybe a setting that defines which view class to use.

How to be with the existing `OSCAR_HOMEPAGE` settings?

> Think we need to have https://github.com/django-oscar/django-oscar-promotions populated before we can merge this?

Indeed.

sasha0:
> It is hard to override the home page view - you have to override the whole Shop application and then replace the URL pattern for the home page, which isn't that straightforward. I think a lot of projects will want to replace that view, so we should provide an easier way to do it - not sure what the best approach is but maybe a setting that defines which view class to use.

After thinking about this for a while, I'm not sure if it not make sense to introduce a new setting, at least for now, since it does not make much difference to have specific `IndexView` or to navigate to the page, intended as a home page.

Anyhow, I didn't have any cases I need to customize `IndexView` on my projects (mostly API or redirection to catalogue or other view through `OSCAR_HOMEPAGE` setting), so I would appreciate any input. If there's no any feedback - I'm inclined just to leave it as is and not introduce new setting.

solarissmoke:
>  I'm inclined just to leave it as is and not introduce new setting.

OK with me.

PR #2630, opened 2018-02-27
Add mysql build to travis matrix

solarissmoke:
This adds a build to our test matrix that runs using MySQL - we need to figure out whether/how to fix the issues that the failing tests are presenting.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,8 @@ matrix:
     - python: 3.6
       env: TOXENV=py36-django111
     - python: 3.6
+      env: TOXENV=py36-django111-mysql
+    - python: 3.6
       env: TOXENV=py36-djangomaster
 
     - python: 3.6
--- a/src/oscar/apps/order/utils.py
+++ b/src/oscar/apps/order/utils.py
@@ -3,6 +3,7 @@ from decimal import Decimal as D
 from django.conf import settings
 from django.contrib.sites.models import Site
 from django.db import transaction
+from django.db.utils import OperationalError
 from django.utils.translation import ugettext_lazy as _
 
 from oscar.apps.order.signals import order_placed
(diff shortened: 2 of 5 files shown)


solarissmoke:
Struggling to get it to actually use mysql - tests are passing because it is somehow still using postgres.

sasha0:
> I've allowed this build to fail because all the concurrency tests fail on MySQL (which presumably means that it would fail in production too). We will need to look at fixing that separately (alternatively we need to state that we don't support MySQL).

I think we need firstly to determine root cause of the failure of concurrency tests, and unicode-related test too. Does not make sense for me to add it to the matrix, then investigate failure and afterwards do any statement (and potentially remove it from the matrix). What do you think?

solarissmoke:
Yeah okay, fair enough. I will try and look into the issues when I have time. There is a separate question though about whether we want to support MySQL at all - these failures (especially the unicode issues) suggest that not many people are using it in the first place.

solarissmoke:
I am not able to find a way around MySQLs deadlock whenever there concurrency. In addition it is somehow possible to use the same voucher twice with MySQL - the integrity checks don't happen as they should. 

We don't use MySQL at all so I am not likely to spend too much more time trying to figure this out. At present I think we need to say that Oscar only really supports Postgres.

ladrua:
We use MySQL, what fairly new to the oscar though, what can I do to help?

solarissmoke:
There are 3 tests that are failing on MySQL (see the travis build) - two related to concurrency and one related to unicode string comparison. Would be great if you can help us figure out what the underlying issues are and how we can resolve them.

For the concurrency issue, MySQL basically throws a "deadlock" exception if you try to make two simultaneous queries that modify the same row. We need to be able to catch that exception and retry - without compromising database integrity.

For the unicode issue it appears that MySQL comparison rates `Château` as equal to `Chateau`...

kooba:
So far my findings:

`AutoSlugFieldTest.test_switch_to_unicode_slug` test on Travis fails because search for existing slug in `AutoSlugField.create_slug` with kwargs `{'slug': 'château-margaux-1960'}` is returning existing `SluggedTestModel` with `slug` `chateau-margaux-1960` so it changes it to be unique `château-margaux-1960-2`. To make utf8 search sensitive in MySQL collations will have to be set to `utf8_bin` (`DATABASES['default']['TEST']['COLLATION'] = 'utf8_bin'` in `tests/settings.py`). 

This unfortunately makes all queries case sensitive and tests like `TestASignedInUser.test_cant_update_their_email_address_if_it_already_exists` will fail because they expect `User.email` uniqueness to be case insensitive. Could user email be lowercased before the save/search?

solarissmoke:
Thanks for that. We could lowercase the email and solve that issue fairly easily.

I think it's worth noting however that both `utf8` and `utf8_bin` comes with another issue which is that they don't support the full range of unicode characters. You're supposed to use `utf8mb4`, but Django itself doesn't support that (see https://code.djangoproject.com/ticket/18392).

kooba:
`utf8_bin` is sufficient to store all language characters and `utf8mb4` is only required to support other characters like emojis, hopefully not required for things like slugs.

solarissmoke:
Problem is that we support arbitrary user input - e.g., reviews - where users can and do post emojis. This actually forced us to stop using mysql...

kooba:
I see now. That's a problem.

sasha0:
I suppose until https://github.com/django/django/pull/8886 is merged and released we won't be able to properly support MySQL. @solarissmoke shall we add a note to the documentation until then?

craigloftus:
We don't support mysql currently, so closing for now.

sniharika211:
> Yeah okay, fair enough. I will try and look into the issues when I have time. There is a separate question though about whether we want to support MySQL at all - these failures (especially the unicode issues) suggest that not many people are using it in the first place.

Hi. Is that fix now. Or we can't use MySQL with DjangoOscar. I got the MYSQL error again. Was eager to know if I am proceeding wrong.

PR #2633, opened 2018-03-02
Merge Oscar and Django app configs

jwayodi:
- Move Oscar app configs (permissions and URLconfs) into Django app config classes
  - Most of this code is copied over as is, with only minor porting to make it work inside the Django app config classes
- Rename the unified app config modules from `config.py` to `apps.py`
- Add support for specifying an application configuration class (introduced in Django 1.7), in addition to an application package, in `INSTALLED_APPS`
  - The `oscar.core.loading.get_class` function now uses the Django app registry to discover the app for a particular app label, whichever way it is specified in `INSTALLED_APPS`
  - Obtain app properties (e.g. labels) from the app config, instead of manually extracting them from the `INSTALLED_APPS` entry
- Use the Django app registry to obtain app configs (`django.apps.apps.get_app_config(app_label)` and `django.apps.AppConfig.create(installed_apps_entry)`), in multiple places
  - e.g. when including URLconfs of a sub app (`oscar.core.loading.get_class(app_package.app, 'application').urls` becomes `django.apps.apps.get_app_config(app_label).urls`)
- Remove the `oscar.get_core_apps` function, and its app overriding
  - To override apps, now just replace the Oscar app entry in `INSTALLED_APPS` (before the Django app registry is initialised), with that of the forked app (with the same app label as that of the overriden Oscar app), and let Django load it into the Django app registry
  - The app config of an overriding app can be obtained from the Django app registry
  - The `oscar.core.loading.get_class` function now uses the Django app registry to discover the app for a particular app label
  - Modify the app-forking code to work without calling the `oscar.get_core_apps` function, and instead use the Django app registry
- Change the API for forking apps, adding explicit specifying of the app config name (`django.apps.AppConfig.name`)
  - The command to fork an Oscar app was previously:
    `python manage.py oscar_fork_app app_label target_path`
    and now becomes:
    `python manage.py oscar_fork_app app_label target_path target_package`


src/oscar/core/customisation.py
```diff
@@ -36,99 +38,100 @@ def subfolders(path):
...
         "      ...\n"
         "  ]\n"
-        "  from oscar import get_core_apps\n"
-        "  INSTALLED_APPS = INSTALLED_APPS + get_core_apps(\n"
-        "      ['%s'])"
-    ) % app_package
-    record = "\n%s\n\n%s" % (
+        "  from oscar import OSCAR_CORE_APPS\n"
+        "  INSTALLED_APPS = INSTALLED_APPS + OSCAR_CORE_APPS\n"
+        "  INSTALLED_APPS[INSTALLED_APPS.index('{}')] = '{}'"
```
craigloftus: I am not sure about  these instructions. I'd be in favour of `Replace the entry 'oscar.apps.partner' with 'fish_shop.apps.partner' in INSTALLED_APPS` And drop the `OSCAR_CORE_APPS` setting in favour of documentation.

SalahAdDin:
Can i help with this?

src/oscar/apps/checkout/apps.py
```diff
@@ -1,24 +1,30 @@
...
-    user_address_delete_view = get_class('checkout.views',
-                                         'UserAddressDeleteView')
-    shipping_method_view = get_class('checkout.views', 'ShippingMethodView')
-    payment_method_view = get_class('checkout.views', 'PaymentMethodView')
-    payment_details_view = get_class('checkout.views', 'PaymentDetailsView')
-    thankyou_view = get_class('checkout.views', 'ThankYouView')
+    namespace = 'checkout'
+
+    def ready(self):
+        self.index_view = get_class('checkout.views', 'IndexView')
```
solarissmoke: What is the reason for having to move these assignments into the `ready()` method?

src/oscar/core/loading.py
```diff
@@ -214,11 +214,12 @@ def _get_installed_apps_entry(app_name):
     This does depend on the order of INSTALLED_APPS and will break if
     e.g. 'dashboard.catalogue' comes before 'catalogue' in INSTALLED_APPS.
     """
-    for installed_app in settings.INSTALLED_APPS:
+    for entry in settings.INSTALLED_APPS:
```
solarissmoke: Can't we use [`get_app_configs()`](https://docs.djangoproject.com/en/2.0/ref/applications/#django.apps.apps.get_app_configs) here instead of re-creating all the configs? Also there might be side effects to initialising configs multiple times.

src/oscar/core/loading.py
```diff
@@ -214,11 +214,12 @@ def _get_installed_apps_entry(app_name):
...
+        app_config = AppConfig.create(entry)
         # match root-level apps ('catalogue') or apps with same name at end
         # ('shop.catalogue'), but don't match 'fancy_catalogue'
-        if installed_app == app_name or installed_app.endswith('.' + app_name):
-            return installed_app
+        if app_config.name == app_name or app_config.name.endswith('.' + app_name):
+            return app_config.name
     return None
```
solarissmoke: It should be possible to make the logic in `_find_installed_apps_entry` much more sane now. If we now require that the forked app uses the same label as the corresponding core app, then we should not have to mess around with sniffing through dotted paths to find a matching app? We should just be able to use [`apps.get_app_config()`](https://docs.djangoproject.com/en/2.0/ref/applications/#django.apps.apps.get_app_config) to fetch the appropriate app.

solarissmoke:
@django-oscar/core I would appreciate review and testing on this, as it is a significant functional change. 

It will break some old behaviour, e.g., the fact that `get_class` worked with any module, even a non-Oscar one, and that it only checked the last portion of a dotted path. If this is going to cause problems for projects then it would be good to get early feedback.

sasha0:
@solarissmoke since we revamp application configuration, load and fork, I propose also to allow forking of third party Oscar apps, since we aim to extract promotions out of the Oscar core. Customization of other pluggable apps is not available at the moment as well. We discussed that potentially the app can provide command to fork it - but I don't think package-specific approach is convenient. WDYT?

solarissmoke:
@sasha0 I would prefer to discuss that as part of a separate issue/PR. This PR is complex enough already without adding more functionality. Also the logic for forking third party apps is more complicated as we found previously.

solarissmoke:
Would be grateful for another review on this. I've tested and am happy that it works well, but it's a bit change so it would be good for another pair of eyes to look at this.

auvipy:
yea definitely, the more the better for review :)

solarissmoke:
@django-oscar/developers @django-oscar/core last call for comments on this PR before we merge :-).

PR #2641, opened 2018-03-10
Add support for Django 2.

solarissmoke:
The blocking dependency (Haystack) now supports Django 2.

### DIFF
--- a/src/oscar/core/application.py
+++ b/src/oscar/core/application.py
@@ -1,8 +1,15 @@
 from django.urls import reverse_lazy
-from django.urls.resolvers import RegexURLPattern as URLPattern
 
 from oscar.core.loading import feature_hidden
 from oscar.views.decorators import permissions_required
+
+
+try:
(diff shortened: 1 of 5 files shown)


src/oscar/core/application.py
```diff
@@ -1,10 +1,17 @@
...
 from oscar.core.loading import feature_hidden
 from oscar.views.decorators import permissions_required
 
 
+try:
+    # Django 2
+    from django.urls import URLPattern
+except ImportError:
+    # Django 1.11
+    from django.urls.resolvers import RegexURLPattern as URLPattern
```
sasha0: I believe we'd move that to `core.compat`.

solarissmoke: I'm not sure why? You'd end up with unused import warnings in `compat.py`.

sasha0: Right. Let's then leave it as is.

PR #2659, opened 2018-03-31
Add a setting to allow overriding Oscar's dynamic class loader.

solarissmoke:
Oscar's dynamic class loader makes some assumptions about how to dynamically load classes, two of which can be problematic:

1. It finds the first installed app whose dotted path ends with the same label as what you are trying to import. This causes problems if you have two apps which share the the same label (e.g., `myproject.search.SomeClass` will be loaded in place of `oscar.apps.search.SomeClass` just because it has `search` in it, even though `myproject.search` is not in any way related to (or forked from) `oscar.apps.search`. We have projects where this is problematic, and it is not possible to simply change the order of apps in `INSTALLED_APPS` as this has other side-effects.

2. If you've forked an app, there is no way to subclass ("double-fork") the classes in that app for use by other apps. Admittedly most projects would never want to do that anyway, but we have cases where for example we've modified the catalogue app and want to be able to reuse that on different projects.

This patch allows projects to replace Oscar's class loader with their own, should they need to customise how classes are dynamically loaded. It is a very simple change (just allowing you to specify your own loader function).


sasha0:
So basically dynamic class loader is going to be dynamic too?
Firstly, wouldn't it even further complicate class loading debugging? Secondly, what custom class loader would you use on your projects? Not yet sure about real-world use-case.

solarissmoke:
> So basically dynamic class loader is going to be dynamic too?

No. It's swappable (i.e., you can use your own if you want). This doesn't make it  any more dynamic than before, and doesn't increase the complexity at all for any project by default.

> Secondly, what custom class loader would you use on your projects? Not yet sure about real-world use-case.

I can give you two use cases based on the points I made above:

1. We have forked the catalogue application and heavily modified it to support some common use cases that we have, and stored this in a separate repo. We use this in several projects. However, because the app has been forked, it's very difficult to further customise this catalogue app at the project level (i.e., we cannot fork it a second time). A custom class loader would solve this issue for us because we could configure it to look in this special app that we want to reuse. This is not an uncommon use case.

2. We use Wagtail on a lot of sites. Wagtail 2 restructured its modules and now comes with a `wagtail.search` module. Try using that with Oscar and it will break, because the default class loader will look inside `wagtail.search` for `oscar.apps.search` modules. One solution is the change the order of installed apps, but this isn't always possible.

src/oscar/core/loading.py
```diff
@@ -58,6 +59,17 @@ def get_class(module_label, classname, module_prefix='oscar.apps'):
 
 
 def get_classes(module_label, classnames, module_prefix='oscar.apps'):
+    # Allow projects to supply their own logic for loading classes.
+    # If none is provided then use Oscar's default_class_loader
+    _custom_class_loader = getattr(settings, 'OSCAR_DYNAMIC_CLASS_LOADER', None)
+    if _custom_class_loader:
+        class_loader = import_string(_custom_class_loader)
```
solarissmoke: I feel like importing this afresh every time this function is called is bad form - but not sure how/where to cache it (bearing in mind that `functools.lru_cache` doesn't exist in Python 2). Any suggestions here would be appreciated.

sasha0:
> This Wagtail example probably points to a more fundamental issue in Oscar, but there are other cases where we hit the same issue (app name conflicts) where a custom class loader would help to avoid them.

Alright, but it does mean we need to modify existing class loader or ship another one which will allow to properly load classes from applications with same package name, like `wagtail.search`.

solarissmoke:
Agreed. I think modifying the existing loader is a much bigger change that needs its own PR. There are also various other issues with the current approach which I think could do with improving - e.g., caching of loader results.

sasha0:
Let's introduce it as a new class loader, not a replacement of the old one, since this potentially could be breaking change.

sasha0:
So, it looks good but needs a mention in release notes.

docs/source/ref/settings.rst
```diff
@@ -480,6 +480,19 @@ Default: ``False``
...
+
+Dynamic importer settings
+=========================
+
+``OSCAR_DYNAMIC_CLASS_LOADER``
+----------------------------------
+
+Default: not set
+
+A dotted Python path to custom callable to dynamically import classes, to replace
```
craigloftus: I'd suggest the following alternative wording:

    Default: ``oscar.core.loading.default_class_loader``

    A dotted path to the callable used to dynamically import classes.

And then specify the value in `defaults.py`

src/oscar/core/loading.py
```diff
@@ -58,6 +59,17 @@ def get_class(module_label, classname, module_prefix='oscar.apps'):
 
 
 def get_classes(module_label, classnames, module_prefix='oscar.apps'):
+    # Allow projects to supply their own logic for loading classes.
+    # If none is provided then use Oscar's default_class_loader
+    _custom_class_loader = getattr(settings, 'OSCAR_DYNAMIC_CLASS_LOADER', None)
```
craigloftus: With a default value specified this function could be a lot simpler, refactor to create `get_class_loader` and then decorate that call with `django.utils.lru_cache` (works with 2 and 3) to cache a reference to the loader.

solarissmoke: Don't think we need to bother with `lru_cache` if we're just returning a settings attribute. It also makes testing easier ;-).

solarissmoke:
@craigloftus updated based on your feedback. @sasha0 updated release notes as well.

PR #2663, opened 2018-04-03
Remove unused OSCAR_SETTINGS setting

solarissmoke:
`OSCAR_SETTINGS` just duplicates all of Oscar's settings into a dictionary, which is then not used anywhere. Looking back at the project history, it seems like it hasn't been used for a few years.


sasha0:
Anyway needs mention in the release docs.

solarissmoke:
I've added a mention to the release notes.

PR #2666, opened 2018-04-09
Update several test dependencies in setup.py.

solarissmoke:


solarissmoke:
The coverage changes appear to be due to a bug in coverage - the `migrations` files that are reporting changes should have been excluded from coverage in the first place.

PR #2668, opened 2018-04-10
Replace insecure md5 hash generation in Order.verification_hash()

### LINKED ISSUE #2667: Anonymous order verification hash uses insecure md5
In `src/oscar/apps/order/abstract_models.py`, the `AbstractOrder` model has a `verification_hash` method, which uses the insecure MD5 hashing algorithm to compute a verification code.

I Am Not A Cryptographer, but my spidey sense is tingling...

- `key = '%s%s' % (self.number, settings.SECRET_KEY)`  <- user already knows the number; hash input is a simple concatenation of number and secret key
- `hash = hashlib.md5(key.encode('utf8'))` <- hash is also known to the user

Wouldn't that make the site-wide secret key recoverable from an anonymous order URL?

solarissmoke:
Fixes #2667. The current approach is insecure as noted there.

**Important note**: this change is obviously not backward compatible (i.e., old anonymous order status URLs will break). I don't know how to prevent that while still patching the security issue. Note that in any case URLs become invalid every time you change the `SECRET_KEY` - i.e., it was easy to break the URLs even before. 

Ideally sites should be changing `SECRET_KEY` after this fix is released - in which case there is no hope of backward compatibility (which is annoying - but I can't see a way around it).

I wonder whether we should be using a different signing key (not `SECRET_KEY`) to avoid this problem in future. Feedback on this is appreciated.

### DIFF
--- a/src/oscar/apps/order/abstract_models.py
+++ b/src/oscar/apps/order/abstract_models.py
@@ -1,11 +1,12 @@
-import hashlib
 from collections import OrderedDict
 from decimal import Decimal as D
 
 from django.conf import settings
+from django.core.signing import BadSignature, Signer
 from django.db import models
 from django.db.models import Sum
 from django.utils import timezone
+from django.utils.crypto import constant_time_compare
 from django.utils.encoding import python_2_unicode_compatible
 from django.utils.timezone import now
 from django.utils.translation import ugettext_lazy as _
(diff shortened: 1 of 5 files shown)


sasha0:
We need either to better state about incompatibility in the docs or provide way to re-generate hashes automatically.

solarissmoke:
@sasha0 I am not sure waht you mean about regenerating hashes? These hashes are not stored anywhere in the database. The only way to preserve backward compatibility is to allow validation of the old (insecurely generated) hashes.

Yes we will need to note the incompatibility - I will prepare the necessary documentation once we agree that this is the right solution to the issue.

solarissmoke:
@craigloftus @sasha0 I realised that there is a way to make this backward compatible, and have added a mechanism to do that. A project must explicitly enable this feature as it means copying the old `SECRET_KEY` into a setting that Oscar can use. This obviously still exposes the old key to exposure, hence the check that makes sure that the old key is not equal to the new one.

Have also added a lengthy note to the release notes for 1.6, which we can then backport into 1.5.

src/oscar/apps/order/abstract_models.py
```diff
@@ -299,9 +302,47 @@ def __str__(self):
...
+        old_verification_key = getattr(settings, 'OSCAR_DEPRECATED_ORDER_VERIFY_KEY', None)
+        if old_verification_key is None:
+            return False
+
+        if old_verification_key == settings.SECRET_KEY:
+            raise ImproperlyConfigured(
+                'OSCAR_DEPRECATED_ORDER_VERIFY_KEY cannot be equal to SECRET_KEY')
+
+        string_to_hash = '%s%s' % (self.number, old_verification_key)
+        order_hash = hashlib.md5(string_to_hash.encode('utf8')).hexdigest()
```
mvantellingen: Can we log an error / warning here? ``logger.warning("Using old ..")``

solarissmoke: Done.

PR #2677, opened 2018-04-13
Add eslint checks to all Oscar JS to the travis build.

### LINKED ISSUE #2674: Add linting of JS assets
We should try to lint JS ~and CSS~ as part of the travis build. 

(Edit: CSS is more tricky because we compile a lot of it - let's deal with that separately).

solarissmoke:
This threw up several issues in our existing JS which have also been fixed - primarily variables being stored in the global context in lots of places.

I have used fairly relaxed rules for the time being (e.g, not enforcing double/single quotes or camel case) which we can review later. Some things (e.g., camel case variable names) we cannot enforce without changing the API exposed by `oscar.ui`, which I don't think we should do.

Fixes #2674.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,0 +1,21 @@
+{
+    "env": {
+        "browser": true
+    },
+    "extends": "eslint:recommended",
+    "rules": {
+        "indent": [
+            "error",
+            4
+        ],
+        "linebreak-style": [
+            "error",
+            "unix"
+        ],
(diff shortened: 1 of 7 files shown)


sasha0:
> Some things (e.g., camel case variable names) we cannot enforce without changing the API exposed by oscar.ui, which I don't think we should do.

Right. We can revise that during architecture revision, which would likely require client code refactoring too.

PR #2679, opened 2018-04-14
Decouple price and availability in base strategies.

### LINKED ISSUE #2664: Regression: price_excl_tax is necessary for availability
I get to upgrade an old Oscar project, and that went really well. Thanks for keeping Oscar so backwards-compatible! I noticed just a small regression for a corner case:

I work on a project where it was essential that Oscar stores the tax-inclusive price and the price without tax is calculated. The reasoning is that the tax-inclusive prices are legally binding, and doing it this way ensures that a bug in the price calculation logic has less of an impact.

This was easily done by storing the price_incl_tax on the stock record, and writing a short pricing policy for it. But two recent changes [here](https://github.com/django-oscar/django-oscar/commit/82d9e7d2647ab50aacd653d21f93f8423f974fbf#diff-38aa6a43fa33c6bad17543acc1e24541) and [here](https://github.com/django-oscar/django-oscar/commit/3e4dda9a89a51323ea50d216bd895101aa014bbf#diff-38aa6a43fa33c6bad17543acc1e24541) make this harder, because they look at a `price_excl_tax` and assume because that is not set, the product is not available to buy. 

I'd like to argue that pricing and availability are concepts that should be kept separate, and the two commits linked above essentially bypass Oscar's pricing logic. 

My workaround was to backport the `StockRequired` policy from Oscar 1.4.

solarissmoke:
Revert the changes in #1913 that required a stock record to have `price_excl_tax` set in order to report the product as available, on the basis that price and availability should be kept separate.

Add separate checks to the basket form/add logic that checks at the time of
adding a product to the basket whether a price exists, and report an error if it doesn't. If a price is not found, then the add to basket form will raise a `ValidationError` - projects are responsible for ensuring that they supply a valid pricing strategy in such cases.

Fixes #2664.

### DIFF
--- a/docs/source/releases/v1.6.rst
+++ b/docs/source/releases/v1.6.rst
@@ -115,6 +115,11 @@ Minor changes
 Backwards incompatible changes in Oscar 1.6
 -------------------------------------------
 
+ - Fixed a regression introduced in Oscar 1.5 (see :issue:`2664`) where
+   ``StockRequired.availability_policy`` was dependent on the product
+   having a price. Price and availability are now decoupled, and it is possible
+   to defer determination of a price until a product is added to the basket.  
+
  - ``oscar.apps.customer.auth_backends.EmailBackend`` now rejects inactive users
    (where ``User.is_active`` is ``False``).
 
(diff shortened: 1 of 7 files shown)


maikhoepfel:
From a brief look over, it looks good to me! Thanks for taking the time to fix, @solarissmoke!

sasha0:
We shall also mention about this in the release docs.

solarissmoke:
Thanks for the review all. I have added a note to the release notes and merged - this will ship in the final 1.6 release.

PR #2682, opened 2018-04-23
Fixes ``is_quality_allowed``, updated docs with correct default setting

maerteijn:
See https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/basket/abstract_models.py#L155

It always expects two results. When ``OSCAR_MAX_BASKET_QUANTITY_THRESHOLD`` is set to something that evaluates to ``False`` this will raise a TypeError exception:
```python
File ".../oscarapi/views/basket.py", line 78, in validate
    allowed, message = basket.is_quantity_allowed(quantity)
File ".../oscar/apps/basket/abstract_models.py", line155, in is_quantity_allowed
    max_allowed, basket_threshold = self.max_allowed_quantity()
TypeError: 'NoneType' object is not iterable
```

In the documentation the default value is set to ``None``. However in https://github.com/django-oscar/django-oscar/blob/master/src/oscar/defaults.py#L17 it's set to 10000. So I updated the documentation./

### DIFF
--- a/docs/source/ref/settings.rst
+++ b/docs/source/ref/settings.rst
@@ -330,7 +330,7 @@ The time to live for the basket cookie in seconds.
 ``OSCAR_MAX_BASKET_QUANTITY_THRESHOLD``
 ---------------------------------------
 
-Default: ``None``
+Default: ``10000``
 
 The maximum number of products that can be added to a basket at once.
 
--- a/src/oscar/apps/basket/abstract_models.py
+++ b/src/oscar/apps/basket/abstract_models.py
@@ -141,10 +141,11 @@ class AbstractBasket(models.Model):
         with the respect to basket quantity threshold.
         """
         basket_threshold = settings.OSCAR_MAX_BASKET_QUANTITY_THRESHOLD
+        max_allowed = 0
         if basket_threshold:
             total_basket_quantity = self.num_items
             max_allowed = basket_threshold - total_basket_quantity
-            return max_allowed, basket_threshold
+        return max_allowed, basket_threshold
 
     def is_quantity_allowed(self, qty):
         """


solarissmoke:
`None` (no limit) is not the same as a default of (10,000) which I know is high, but we are still changing default behaviour, which I don't think we should without justification. 

I think it would be preferable to fix `Basket.max_allowed_quantity` to return a valid result when `OSCAR_MAX_BASKET_QUANTITY_THRESHOLD = None` - simply changing it to return `None, None` would fix this issue.

solarissmoke:
(I've deleted my previous comment, which didn't make a lot of sense).

I think the original intention was that  `OSCAR_MAX_BASKET_QUANTITY_THRESHOLD = None` means there is no limit. The change you have proposed would result in `max_allowed_quantity` returning `0` for `max_allowed` in such a case - i.e., you would never be able to add anything to the basket. I don't think that is what you intended?

I think it would be preferable to fix `Basket.max_allowed_quantity` to simply return `None, None` when  `OSCAR_MAX_BASKET_QUANTITY_THRESHOLD = None`.

maerteijn:
@solarissmoke Yes, you are right. I was a bit confused by
```python
if max_allowed is not None and qty > max_allowed:
```
but in the morning with a fresh cup of coffee it makes more sense 😀

solarissmoke:
Would you mind adding a unit test to prevent a regression of this issue? Otherwise looks good to me.

maerteijn:
Improved the tests as requested.

And suddenly the ``tests/functional/test_basket.py`` was failing. First I thought it had something to do with this pull request, but they released ``django-extra-views==0.11.0`` an hour ago which raises a lot of ``DeprecationWarning`` exceptions.

So I pinned it to ``<0.11.0`` for now. But this should be addressed in another pull request I think

sasha0:
@maerteijn thanks for the patch and @solarissmoke for the review and feedback.

PR #2689, opened 2018-04-26
Update ModelFormSetView subclasses to use new factory_kwargs attribute.

### LINKED ISSUE #2688: Compatibility with django-extra-views 0.11
django-extra-views 0.11 raises the following exception, which we don't catch:

> Setting `BasketView.extra` at the class level is now deprecated. Set `BasketView.factory_kwargs` instead.

solarissmoke:
Support for the old-style attributes was dropped in django-extra-views 0.11 (see [changelog](https://github.com/AndrewIngram/django-extra-views/blob/master/CHANGELOG.rst))

Also pin the version in setup.py as the package does not use semantic versioning (i.e, there are breaking changes between 0.x releases).

Fixes #2688.

### DIFF
--- a/docs/source/releases/v1.6.rst
+++ b/docs/source/releases/v1.6.rst
@@ -118,7 +118,7 @@ Backwards incompatible changes in Oscar 1.6
  - Fixed a regression introduced in Oscar 1.5 (see :issue:`2664`) where
    ``StockRequired.availability_policy`` was dependent on the product
    having a price. Price and availability are now decoupled, and it is possible
-   to defer determination of a price until a product is added to the basket.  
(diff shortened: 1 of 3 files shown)


sasha0:
We shall mention about this in release notes.

solarissmoke:
Updated the release notes. Will backport to 1.5 once this is merged.

veris-neerajdhiman:
@solarissmoke I am still getting this error 
`DeprecationWarning: Setting `StoreAddressInline.extra` at the class level is now deprecated. Set `StoreAddressInline.factory_kwargs` instead.`

apps version 
- django-extra-views==0.11.0
- django-oscar==1.6.4

Issue #2652, opened 2018-03-17
Offer Tracking for Stacked Offers


ad-65:
I was testing out the stacked offer tracking that will be included in v1.6 and came across what may be a problem. It looks like the `LineOfferConsumer` class might be tracking the applied offers incorrectly.

As an example, in the basket app when  `AbstractLine.consume()` runs it should mark a line as consumed using `LineOfferConsumer.consume()` and passing the applicable offer. I think it should be storing the offer that consumed the items regardless of if the offer was exclusive or not. However, if the offer is exclusive it doesn't store the offer that consumed the line items.

This is for a test basket with an offer that should be applied to all three items in the line:

```python
class Line(AbstractLine):

    def consume(self, quantity, offer=None):
        super(Line, self).consume(quantity, offer)
        print "offer------------------------->", offer
        print "exclusive--------------------->", offer.exclusive
        print "has_discount------------------>", self.has_discount
        print "quantity_with_discount-------->", self.quantity_with_discount
        print "has_offer_discount------------>", self.has_offer_discount(offer)
        print "quantity_with_offer_discount-->", self.quantity_with_offer_discount(offer)
```

Output:

```
offer-------------------------> Discount 123456
exclusive---------------------> True
has_discount------------------> True
quantity_with_discount--------> 3
has_offer_discount------------> False
quantity_with_offer_discount--> 0
```

If the offer is set to non-exclusive then the behavior is as expected:

```
offer-------------------------> Discount 123456
exclusive---------------------> False
has_discount------------------> True
quantity_with_discount--------> 3
has_offer_discount------------> True
quantity_with_offer_discount--> 3
```

I don't actually need stacked offers per se, I just need to know what offer was applied to the line as a whole. Maybe this is stored somewhere else when a whole line is consumed by the same offer?

pjstevns:
I'll take a look at this asap.

The LineOfferConsumer interface could do with more extensive tests.

PR #2690, opened 2018-04-26
Add support for restricted offer combinations

pjstevns:
- non-exclusive offers that only allow combination with specific other
  offers


solarissmoke:
If you have time it might be worth fixing https://github.com/django-oscar/django-oscar/issues/2652 while you're at it?

sasha0:
@pjstevns all's set now?

sasha0:
Also we'd anyway add at least a note to the documentation about restricted offer combinations.

pjstevns:
@sasha0 This MR is in quite good shape now.

solarissmoke:
FWIW I think we should hold off merging this until #2788 is fixed - that was introduced when stacked offers were added, and it will get harder to fix later on.

pjstevns:
Fair enough. I'll look into this.

sasha0:
Alright, since fix for #2788 was delivered, it's time to finalize this one.

Looks good, but please rebase remote branch.

setup.cfg
```diff
@@ -7,7 +7,7 @@ filterwarnings =
 
 [flake8]
 exclude = migrations
-ignore = F405,W503,E731
+ignore = F405,W503,E731,D100,D101,D102,D106,D107
```
sasha0: Why was the rule extended?

src/oscar/apps/dashboard/vouchers/apps.py
```diff
@@ -41,5 +41,15 @@ def get_urls(self):
                 name='voucher-delete'),
             url(r'^stats/(?P<pk>\d+)/$', self.stats_view.as_view(),
                 name='voucher-stats'),
+            url(r'^sets$', self.set_list_view.as_view(),
```
solarissmoke: I think it makes sense to expose these in the dashboard menu while we're at it? Wonder if that (and this) is better done in a separate PR though.

solarissmoke:
This looks good to me (one small comment on the change to voucher set views). I want to hold off merging this until we release 2.0, just to avoid having to maintain a separate tree on the 2.0 release branch.

sasha0:
Since 2.0 is already out, I think after rebase this PR is ready to merge.

solarissmoke:
Rebased in #3485 as I'm not able to push to this PR branch.

PR #2693, opened 2018-04-28
Prepare bugfix release 1.5.4

solarissmoke:
1. Pin version of django-extra-views to <0.11
2. Upgrade TinyMCE to v4.7.11.


setup.py
```diff
@@ -26,7 +26,7 @@
     # We use the ModelFormSetView from django-extra-views for the basket
```
craigloftus: I think we should update this comment to reflect the new reason it is pinned.

solarissmoke: Yep, I've updated that comment.

Issue #2685, opened 2018-04-25
setup() should add zip_safe=False parameter in setup.py 


john-shine:
cd django-oscar folder
python setup.py install
it make django-tables2 install as egg file instead of package folder, then when I enter dashboard, it comes Error:
TemplateDoesNotExist: django_tables2/table.html
take me much time to debug.

To avoid this problem, zip_safe=False parameter should add to setup() in setup.py

solarissmoke:
As per my comment in #2695, it looks like this is an issue with dependency packages rather than Oscar. I'm closing for now but feel free to reopen if you can clarify the reason this is necessary and steps to reproduce the problem.

PR #2695, opened 2018-05-07
fix packages installed as egg

john-shine:
fix packages installed as egg

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -93,6 +93,7 @@ setup(
         'docs': docs_requires,
         'test': test_requires,
     },
+    zip_safe=False,
     classifiers=[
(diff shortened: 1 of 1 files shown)


solarissmoke:
@john-shine please can you clarify the issue that this is meant to solve, and the steps to reproduce it? I have tried to follow the steps you gave in #2685 and I'm not able to reproduce an issue. Also the issue in #2685 appears to be with a dependency rather than with Oscar itself?

solarissmoke:
Closing due to lack of feedback - feel free to reopen if you can clarify the reason this is necessary and steps to reproduce the problem.

PR #2698, opened 2018-05-11
Fix description of the `add_product` method of the `Basket` model

Jamim:
The `stock_info` attribute was removed from `add_product` method's signature at 05363f6854f179ee1a96511661302980be50ca6e, but it wasn't removed from the description. So it makes [the docs](https://django-oscar.readthedocs.io/en/latest/ref/apps/basket.html#oscar.apps.basket.abstract_models.AbstractBasket.add) a bit confusing.


solarissmoke:
Thanks!

PR #2700, opened 2018-05-13
Fix validation of empty/missing phone_number field in PhoneNumberMixin.

solarissmoke:
If a `phone_number` field is missing or empty, then `PhoneNumberMixin.clean()` fails with a key error because the field does not exist in `cleaned_data`. This patch fixes the `clean_phone_number_field()` method to only attempt to clean a field if it exists in `cleaned_data`.

Also added more comprehensive tests for the validation logic in the mixin, to ensure that required/optional phone number fields are correctly handled.


sasha0:
Shall we mention this in release notes?

solarissmoke:
Yeah, sorry, I forgot earlier. Have added a note to the release notes.

Issue #2702, opened 2018-05-19
RelatedFieldWidgetWrapper raises KeyError on Django 


kmetaxas:
Affects: Django 1.11.0 but not >=1.11.1 (i never tested <1.10. *may* affect early 2.0.)

Problem: Trying to create or view a ProductClass in th Dashboard results in a KeyError on context['field'].

Bug description
-------------------
`RelatedFieldWidgetWrapper` in apps/dashboard/widgets.py: inherits from Widget but does not call super().\_\_init__.  
This means that some variables that `Widget.get_context()` builds are not present in `RelatedFieldWidgetWrapper.get_context()` (context['widget']).

When rendering ProductClass form in the dashboard, the template calls widget_tweaks's `render_field` with a class attribute added. When adding class attributes, widget tweaks will test the `BoundField` for truthyness.
In Django==1.11.0 , `BoundField` is only has \_\_len__ (which is used for truthyness tests) which eventually leads to accessing context['widget'] (which is not created). In later versions (after [this commit](https://github.com/django/django/commit/844ae40a7cdcb1738ca3a2e8f13bb6b8c2e22efd) ) BoundField got a `__bool__()` method so no access to subwidgets happens.

Solution 1: Change `RelatedFieldWidgetWrapper` to call super().\_\_init__. I don't quite understand it entirely and maybe there is a reason it doesn't, so i'm not submitting a PR
Solution 2: Require >Django==1.11.0

solarissmoke:
This is a bug fixed in Django 1.11.1 (https://code.djangoproject.com/ticket/28058). See also https://github.com/jazzband/django-widget-tweaks/issues/44.

> Solution 2: Require >Django==1.11.0

I think our implicit assumption is that projects will follow Django's recommendation of always using the latest minor version. There have been many bug fixes since Django 1.11.0 was released, several of which would affect an Oscar project. 

On that basis my suggestion here would be simply to update your project to the latest version of Django.

solarissmoke:
Can you explain a use case where a field using this widget would implement subwidgets? 

Our `RelatedFieldWidgetWrapper` is basically a copy of [Django's], which also doesn't implement subwidgets (because it is only used for a very specific purpose). I can't see why we would need it to implement it here.

solarissmoke:
[...]

PR #2706, opened 2018-05-21
fix Bug2702. delegate subwidgets() to wrapped widget.

kmetaxas:
Should fix bug #2702 .
Delegates `subwidgets()` to wrapped `self.widgets.subwidgets()`.
Also added test case that simply makes sure that subwidget's generator can be made into a list (no exceptions, and can come to an end successfully)


solarissmoke:
See discussion in #2702.

PR #2708, opened 2018-05-22
Drop the use of empty and . values for the form action attribute

craigloftus:
[Following Django](https://github.com/django/django/commit/4660ce5a6930e07899ed083801845ee4c44c09df), this commit drops the use of empty action attributes, inline with the HTML5 spec.

### DIFF
--- a/docs/source/releases/v2.0.rst
+++ b/docs/source/releases/v2.0.rst
@@ -29,6 +29,8 @@ Removal of deprecated features
 
 Minor changes
 ~~~~~~~~~~~~~
+ - Dropped ``action=""`` and ``action="."`` attributes, following the lead of Django
+   and as per the HTML5 specification.
 
 Dependency changes
 ------------------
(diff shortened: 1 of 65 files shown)


solarissmoke:
For reference, here is the [relevant section](https://www.w3.org/TR/html5/sec-forms.html#form-submission-algorithm) of the HTML5 specification:

> 7. Let action be the submitter element’s action. 
> 8. If action is the empty string, let action be the document’s URL of the form document.

[and](https://www.w3.org/TR/html5/sec-forms.html#form-control-infrastructure-form-submission):

> The `action` and `formaction` content attributes, if specified, must have a value that is a valid non-empty URL potentially surrounded by spaces.
> 
> The **action** of an element is the value of the element’s `formaction` attribute, if the element is a Submit Button and has such an attribute, or the value of its form owner’s action attribute, if it has one, or else the empty string.

PR #2710, opened 2018-05-23
Use Django's class-based authentication views.

solarissmoke:
The procedural views were deprecated in Django 1.11 and will be removed in Django 2.1.


src/oscar/app.py
```diff
@@ -39,22 +39,26 @@ def get_urls(self):
...
+                login_forbidden(
+                    auth_views.PasswordResetConfirmView.as_view(
+                        form_class=self.set_password_form,
+                        success_url=reverse_lazy('password-reset-complete')
+                    )
+                ),
                 name='password-reset-confirm'),
             url(r'^password-reset/complete/$',
-                login_forbidden(auth_views.password_reset_complete),
+                login_forbidden(auth_views. PasswordResetCompleteView.as_view()),
```
craigloftus: Looks like you've got an extra space here.

solarissmoke: Thanks, fixed.

PR #2711, opened 2018-05-23
 Limit access to dashboard for partners to show only their stats

Jamim:
These changes makes the Dashboard more safe in terms of displaying potentially sensitive store information to non staff (probably partner) users and at the same time gives them an ability to see more relevant stats.

### DIFF
--- a/src/oscar/apps/dashboard/views.py
+++ b/src/oscar/apps/dashboard/views.py
@@ -185,10 +185,6 @@ class IndexView(TemplateView):
             'total_open_stock_alerts': open_alerts.count(),
             'total_closed_stock_alerts': closed_alerts.count(),
 
-            'total_site_offers': self.get_active_site_offers().count(),
-            'total_vouchers': self.get_active_vouchers().count(),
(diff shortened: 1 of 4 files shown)


src/oscar/apps/dashboard/views.py
```diff
@@ -146,13 +134,32 @@ def get_stats(self):
         datetime_24hrs_ago = now() - timedelta(hours=24)
 
         orders = Order.objects.all()
+        alerts = StockAlert.objects.all()
+        baskets = Basket.objects.filter(status=Basket.OPEN)
+        customers = User.objects.filter(orders__isnull=False).distinct()
+        lines = Line.objects.filter()
+
+        user = self.request.user
+        if not user.is_staff:
```
solarissmoke: I think it's worth adding some tests to check that these filters return the expected results.

solarissmoke:
Thanks for this! It looks good to me - my only request would be some additional tests to make sure that the queryset filters that are applied for non-staff users are returning expected results.

sasha0:
I'll add tests in the separate PR.

PR #2714, opened 2018-05-24
Drop support for Python 2

### LINKED ISSUE #2608: Drop support for Python 2
Oscar should drop support for Python 2 in version 1.7. Version 1.6 (which we hope to release soon) will continue to support Python 2 for as long as Django 1.11 (which supports Python 2) is maintained, but will only contain security and bug fixes. 

New features added to version 1.7 will  support Python 3 only.

This will enable us to simplify and rationalise the existing code base. It will also ensure that we are not blocked from upgrading upstream libraries, several of which have also switched to supporting Python 3 only.

solarissmoke:
It's quite a big diff... probably better reviewed one commit at a time. Fixes #2608.


src/oscar/apps/address/abstract_models.py
```diff
@@ -553,7 +549,7 @@ def save(self, *args, **kwargs):
         # Ensure that each user only has one default shipping address
         # and billing address
         self._ensure_defaults_integrity()
-        super(AbstractUserAddress, self).save(*args, **kwargs)
+        super().save(*args, **kwargs)
```
mvantellingen: This one might be wrong

src/oscar/apps/address/abstract_models.py
```diff
@@ -574,7 +570,7 @@ class Meta:
         unique_together = ('user', 'hash')
 
     def validate_unique(self, exclude=None):
-        super(AbstractAddress, self).validate_unique(exclude)
+        super().validate_unique(exclude)
```
mvantellingen: same

PR #2723, opened 2018-05-29
Make the checkout/thank_you.html template more customizable

Jamim:
This PR:
- moves the basket totals to a partial template ([just like it is for the checkout/checkout.html](https://github.com/django-oscar/django-oscar/blob/1.6.1/src/oscar/templates/oscar/checkout/checkout.html#L153))
- wraps the order contents with a block ([like it is in checkout/checkout.html](https://github.com/django-oscar/django-oscar/blob/1.6.1/src/oscar/templates/oscar/checkout/checkout.html#L97), too)


solarissmoke:
Looks good - thanks for this!

PR #2729, opened 2018-05-31
Add a currency middleware that sets the current currency on the request object

v1kku:
I'm breaking down #2606 into smaller PRs to make it easier to review and merge. This is the first PR.


craigloftus:
Discussed with @v1kku and we're going to take a different approach.

PR #2732, opened 2018-06-01
Make OrderPlacementMixin more customizable

Jamim:
This PR changes the signature of the `get_message_context` method to make it useful for various event codes that could be provided to the `send_confirmation_message` method.

### DIFF
--- a/src/oscar/apps/checkout/mixins.py
+++ b/src/oscar/apps/checkout/mixins.py
@@ -270,7 +270,7 @@ class OrderPlacementMixin(CheckoutSessionMixin):
         return reverse('checkout:thank-you')
 
     def send_confirmation_message(self, order, code, **kwargs):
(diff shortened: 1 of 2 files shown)


src/oscar/apps/checkout/mixins.py
```diff
@@ -292,7 +292,7 @@ def send_confirmation_message(self, order, code, **kwargs):
             logger.warning("Order #%s - no %s communication event type",
                            order.number, code)
 
-    def get_message_context(self, order):
+    def get_message_context(self, order, code):
```
solarissmoke: I think `code` needs to default to `None` so as not to break backward compatibility with projects that have overridden this method and are not expecting a `code` argument.

Jamim:
Hello @solarissmoke,
I made few changes on this PR.
Could you please review it again?
Thanks!

PR #2734, opened 2018-06-08
add elasticsearch based search app

v1kku:
Given the size and difficulty to review of #2606 I'm breaking it down into smaller chunks. This is the search app alone and it's tests. The intended plan is to replace the old search app but this is being submitted in a different module first to make it easier to review the code as otherwise the diffs would get in the way, and those aren't needed as this is a completely new app (doesn't need to be compared with the old one).


src/oscar/apps/es_search/forms.py
```diff
@@ -0,0 +1,97 @@
...
+    # focus on the search widget.
+    q = forms.CharField(
+        label=_('Search'),
+        widget=SearchInput({
+            "placeholder": _('Search'),
+            "tabindex": "1",
+            "class": "form-control"
+        }))
+
+    items_per_page = forms.TypedChoiceField(
```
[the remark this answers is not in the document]
v1kku: The search class lets you set the max results, I think this should still be here even without using `OSCAR_PRODUCTS_PER_PAGE`.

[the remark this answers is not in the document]
craigloftus: Okay

src/oscar/apps/es_search/facet.py
```diff
@@ -0,0 +1,185 @@
...
+from purl import URL
+
+from django.utils.translation import gettext_lazy as _
+
+
+class Facet(object):
+
+    facet_name = None
+
+    def __init__(self, name, agg, request_url, user_defined_facets, selected_facets={}):
```
solarissmoke: Please drop `{}` as a default argument, as this is passed by reference and is mutable - i.e., every time a facet is initialised the **same** dictionary will be used. Need to default to `None` and then assign a dictionary in the method itself.

src/oscar/apps/es_search/facet.py
```diff
@@ -0,0 +1,185 @@
...
+
+class FacetsBuilder(object):
+
+    FACET_TYPE_MAPPING = {
+        'range': RangeFacet,
+        'auto_range': RangeFacet,
+        'histogram': HistogramFacet
+    }
+
+    def __init__(self, aggs, request_url, selected_facets={}, user_defined_facets={}):
```
solarissmoke: Drop `{}` as a default argument, as this is passed by reference and is mutable.

src/oscar/apps/es_search/abstract_models.py
```diff
@@ -0,0 +1,17 @@
+from django.db import models
+from django.utils.translation import gettext_lazy as _
+
+
+class AbstractSynonym(models.Model):
+    synonyms = models.CharField(max_length=255,
+                                help_text='Comma-separated list of synonyms, optionally mapped using '
```
solarissmoke: Help text needs translation

src/oscar/apps/es_search/facet.py
```diff
@@ -0,0 +1,185 @@
...
+        from_value = bucket.get('from', '')
+        to_value = bucket.get('to', '')
+        return '{}-{}'.format(from_value, to_value)
+
+    @staticmethod
+    def get_display_name(bucket):
+        from_value = int(bucket.get('from', 0))
+        to_value = int(bucket.get('to', 0))
+        if not from_value:
+            return _('Up to {}'.format(to_value))
```
solarissmoke: The way this is translated is wrong - the translated string will change dynamically based on `to_value`. Should be:

```
_('Up to %(to_value)s') % {'to_value': to_value}
```

Same for all the other translations in this method.

src/oscar/apps/es_search/facet.py
```diff
@@ -0,0 +1,185 @@
...
+
+    def get_display_name(self, bucket):
+        interval = self.user_defined_facets[self.field_name]['params']['interval']
+
+        from_value = int(float(bucket.get('key')))
+
+        to_display_value = (from_value + interval) - 1
+
+        if not from_value:
+            return 'Up to {}'.format(to_display_value)
```
solarissmoke: Missing translations.

src/oscar/apps/es_search/models.py
```diff
@@ -0,0 +1,12 @@
+from oscar.core.loading import is_model_registered
+
+from .abstract_models import AbstractSynonym
+
+__all__ = []
```
solarissmoke: This should not be empty?

v1kku: It should be empty at this point. Synonym is appended on line 12

sasha0:
Hm, why do we generally need additional application `es_search`? I assume we could just keep everything in the `search` application.

solarissmoke:
@sasha0 that is temporary, to allow reviewing of this app without a huge diff on the search app. Once we are ready to merge we will replace `search`.

specialunderwear:
I'd like to test this code, but where has the rest of it gone? I mean the catalogue documents and such?

solarissmoke:
@specialunderwear see https://github.com/django-oscar/django-oscar/pull/2606 .

sandeepimpressads:
Is there any timeline for this to be complete ?

specialunderwear:
I tried to use the elasticsearch search the way it is, picking pieces and bits from #2606 . I got it to work but the complexity when trying to integrate with customizations was daunting at best. As I have mentioned here: https://github.com/django-oscar/django-oscar/pull/2606/files#r259803631 I think this complexity comes from the usage of elasticsearch-dsl, which unfortunately appears to be a bad match with oscar's EAV model. I ended up using the search engine incorparated into our CMS of choice, wagtail. It ships with an implementation of a querycompiler for django's ORM querylanguage straight to elasticsearch json queries. I was able to plug that directly into the catalogue and with some work into the site search as well. I will be working to get some of that code merged back into wagtail on the upcoming wagtail sprint (https://www.wagtail.space) .

sasha0:
> I will be working to get some of that code merged back into wagtail on the upcoming wagtail sprint (https://www.wagtail.space) .

Since Wagtail Space Arnhem 2019 finish, @specialunderwear would you please share if your proposed approach of new search engine worked well for the Wagtail?

specialunderwear:
It works very well, I have finished my own implementation, but I haven't finished my contribution to wagtail core yet. I will need some time to get everything sorted and check if I can release/contribute a solution for oscar.

sasha0:
Could you please leave few notes how do you construct elasticsearch queries and provide few details about the implementation?

Apparently everyone subscribed to the thread interested in transition from Haystack and it would be good to give more context for the comparison with the solution, provided in the PR.

specialunderwear:
Well, because an actual querycompiler was implemented from django orm to elasticsearch, queries do not have to be changed for the browse pages, they just work. As for searching, it works by an extra method on the queryset ``search``. Have a look: http://docs.wagtail.io/en/v2.0/topics/search/index.html

Also very nice to have, if some django query can not be made because for example a field is used in a filter that is not in the index, an error is raised explaining exactly how the missing field can be added to the index.

Some feature I have added and not yet contributed is

1. query for fields that do not correspond to a model field (arbitrary indexed fields)
2. Use es6 suggestions
3. Use es6 completion native type for search bar completions.

sasha0:
Can't wait to see your implementation, at least draft version.

Ready to help out with testing, finalization etc.

specialunderwear:
Ok I will try hurry up a bit.

craigloftus:
I think it is clear now that the implementation in this PR isn't the right approach, so going to close this PR.

specialunderwear:
> Ok I will try hurry up a bit.

I've got this prepared, which can be used right now, but needs removal of a dependency before it could be contributed to oscar https://github.com/specialunderwear/django-oscar-elasticsearch
I could use some help with that!

PR #2738, opened 2018-06-12
Mention support for Django 2.0

alasdairnicol:
The tox config and PyPI page both suggest that Django 2.0 is supported, but it isn't mentioned on the release notes.

### DIFF
--- a/docs/source/releases/v1.6.rst
+++ b/docs/source/releases/v1.6.rst
@@ -19,8 +19,8 @@ Table of contents:
 Compatibility
 -------------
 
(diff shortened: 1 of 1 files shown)


solarissmoke:
Hi @alasdairnicol , this has already been fixed [here](https://github.com/django-oscar/django-oscar/commit/16772a3b9b8a390c5e2a74da05f719649b51671f). The master branch and pypi will be updated when we issue a new minor release, which will be soon as there are a couple of bug fixes to push out. Thanks!

alasdairnicol:
@solarissmoke sorry I missed that commit. Thanks for the quick response.

PR #2742, opened 2018-06-14
Drop backward compatibility for dynamic loading of moved formset classes

solarissmoke:


src/oscar/core/loading.py
```diff
@@ -140,27 +112,7 @@ def default_class_loader(module_label, classnames, module_prefix):
...
-                'please update your imports.' % (', '.join(oscar_moved_classnames),
-                                                 oscar_move_module_label),
-                DeprecationWarning, stacklevel=2)
-            oscar_move_module = _import_module(oscar_move_module_label, classnames)
-        else:
-            oscar_move_module = None
-    else:
-        oscar_move_module = None
-
-    if oscar_module is oscar_move_module is local_module is None:
```
sasha0: I don't think we should remove the logic too - it may be useful in the future if any other classes/functions will move. We can just remove formsets from the `MOVED_ITEMS`. WDYT?

solarissmoke: I don't think we need all this complexity. Backward compatibility can be achieved just by aliasing classes if we ever move them. 

For example, with `apps.basket.forms.BasketLineFormSet` you would just put this into `forms.py`:

```
# Load the new class from the correct location
MovedBasketLineFormSet = get_class('basket.formsets', 'BasketLineFormSet')

# Subclass the new class into the old one, so that the old one can still be loaded.
@deprecated
class BasketLineFormSet(MovedBasketLineFormSet):
    pass
```

Then you just delete this deprecated class when the deprecation period has ended.

solarissmoke: This would also let you import normally from the old (deprecated) path, which you can't currently do - which means the current approach isn't fully backward compatible because any project that has subclassed the original class cannot import from the old path anyway.

PR #2747, opened 2018-06-15
Remove various deprecated utilities, methods and classes.

solarissmoke:
This PR removes:

- `customer.forms.SetPasswordForm`
- `customer.forms.PasswordChangeForm`
- `views.decorators.staff_member_required`
- `UserAddress.num_orders`

### DIFF
--- a/docs/source/releases/v2.0.rst
+++ b/docs/source/releases/v2.0.rst
@@ -43,10 +43,21 @@ Removal of deprecated features
 
   Product alert emails are now sent as Communication Events.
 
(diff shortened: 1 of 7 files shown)


solarissmoke:
I thought about this some more and am now wondering whether we should remove `SetPasswordForm` and `PasswordChangeForm`, even though they are copies of the corresponding Django forms. Removing them means it would not longer be possible to customise them through dynamic loading. Maybe it's better just to leave them there?

craigloftus:
It is easy enough to override the view or application. I think the better argument for keeping them would be consistency.

Issue #2745, opened 2018-06-14
Issue with license of Unidecode dependency


craigloftus:
A dep conflict we were having with Unidecode led me to a [Wagtail issue](https://github.com/wagtail/wagtail/issues/3311) pointing out that Unidecode is GPLv2+ licensed, a viral license, which is incompatible with Oscar's BSD license.

We're using Unidecode to sanely strip unicode chars in our custom [slugify](https://github.com/django-oscar/django-oscar/blob/master/src/oscar/core/utils.py#L38) function.

Without just dropping the `OSCAR_SLUG_ALLOW_UNICODE` setting and allowing unicode slugs everywhere, the only alternative I can see would be to adopt a solution like Wagtail's `cautious_slugifier` ([wagtail/core/utils.py#L64](https://github.com/wagtail/wagtail/blob/f0eef2fc88043ba627b89125876f1cfe53b7fc51/wagtail/core/utils.py#L64))?

solarissmoke:
The [commit](https://github.com/django-oscar/django-oscar/commit/051c9770ebee4ebd5b17eb287ce1459b76deac7f) that introduced this setting was intended to provide backward compatibility for the `allow_unicode` argument added to Django's own slugify function in Django 1.9.

Since we no longer support Django 1.8 I'd say we just replace this whole thing with Django's and pass `OSCAR_SLUG_ALLOW_UNICODE` as the `allow_unicode` argument.

craigloftus:
But the problem with Django's slugify with `allow_unicode=False` is that it just strips the unicode characters, potentially resulting in zero length slugs.

```
slugify('你好')
> ''

slugify(unidecode('你好'))
> 'ni-hao'
```

solarissmoke:
Hmm okay. In that case I agree that a `cautious_slugifier` is the best alternative.

solarissmoke:
Fixed in #2749.

PR #2749, opened 2018-06-15
Drop unidecode and replace with cautious_slugify from Wagtail

craigloftus:
Unidecode's license is GPLv2, which is not compatible with Oscar's license. See #2745 for discussion.

This PR introduces the `cautious_slugify` function from Wagtail as an alternative to produce non-empty slugs from unicode strings. See the tests for examples of the change in output.

I changed the API slightly to drop support for directly passing the `OSCAR_SLUG_FUNCTION` a function and also changed the call signature of the slug function to match the Django slugify function; which will make it easier for downstream projects to decide how to handle unicode.


src/oscar/core/utils.py
```diff
@@ -8,43 +10,89 @@
...
-    Slugify a string (even if it contains non-ASCII chars)
+    Slugify a string
+
+    The OSCAR_SLUG_FUNCTION can be set with a dotted path to the slug
+    function to use, defaults to 'oscar.core.utils.default_slugifier'.
+
+    OSCAR_SLUG_MAP can be set of a dictionary of target:replacement pairs
+
+    OSCAR_SLUG_BLACKLIST can be set to a iterable of words to remove after
+    the slug is generated; though it will not reduce a slug to zero length.
```
solarissmoke: Think we need a test for this behaviour ("it will not reduce a slug to zero length").

docs/source/releases/v2.0.rst
```diff
@@ -40,6 +40,14 @@ Minor changes
 - ``OrderPlacementMixin.get_message_context()`` is now passed a ``code`` argument
   specifying the communication event type code for a message being sent.
 
+- We've dropped the dependency on Unidecode due to license incompatibilities,
+  ``oscar.core.utils.default_slugifier`` now uses ``oscar.core.utils.cautious_slugify``
+  to handle unicode characters in slugs when ``OSCAR_SLUG_ALLOW_UNICODE`` is ``False``.
+
+- ``OSCAR_SLUG_FUNCTION`` previously accepted a function as its value, it now
```
solarissmoke: `It now only accepts...` should be a new sentence.

src/oscar/core/utils.py
```diff
@@ -8,43 +10,89 @@
...
+
+def default_slugifier(value, allow_unicode=False):
     """
-    Oscar's default slugifier function. Uses Django's slugify function.
+    Oscar's default slugifier function. When unicode is allowed
+    it uses Django's slugify function, otherwise it uses cautious_slugify.
     """
-    return django_slugify(value, allow_unicode=settings.OSCAR_SLUG_ALLOW_UNICODE)
+    if allow_unicode:
+        slug = django_slugify(value, allow_unicode=True)
```
solarissmoke: Minor, but you could just return here instead of assigning to `slug`.

src/oscar/core/utils.py
craigloftus: That is what `test_uses_blacklist_all_words` is testing for

solarissmoke: As yes, OK.

PR #2752, opened 2018-06-19
Add compatibility for Django 2.1, and drop support for Python 3.4.

solarissmoke:
Django 2.1 works fine already - the only issue was that more passwords have been added to the CommonPasswordValidator which caused some tests to fail. I've removed that validator from the test settings.

Python 3.4 reaches end of life in March 2019 and is not supported by Django 2.1 and above. Seeing as it will be another few months before Oscar 2 is released, I think we can safely drop support for it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,21 +1,24 @@
 sudo: false
 language: python
 
+addons:
+  postgresql: "9.6"
+
 matrix:
   fast_finish: true
   include:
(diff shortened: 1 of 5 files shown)


solarissmoke:
Do we want/need to support Django 2.1 in Oscar 1.6.x? If so there are some additional fixes that we need to backport (e.g., dropping procedural auth views).

PR #2753, opened 2018-06-19
[1.6.x] Add compatibility for Django 2.1

solarissmoke:
1. Drop use of procedural auth views that are removed in Django 2.1
2. Remove `CommonPasswordValidator` from test settings, as this catches some test passwords.


bgsuello:
is there an ETA for this pr?

PR #2755, opened 2018-06-21
Update urls.py configuration

raratiru:
Django 2.0  uses django.urls.path() function which allows for a simpler, more readable URL routing syntax.

From Django 1.9, application.urls should be passed directly,  without include.

### DIFF
--- a/docs/source/internals/getting_started.rst
+++ b/docs/source/internals/getting_started.rst
@@ -186,6 +186,28 @@ you will also need to include Django's i18n URLs:
         url(r'', include(application.urls)),
     ]
 
+For `Django > 1.9`_ pass ``application.urls`` directly to ``path``. For `Django > 2.0`_ use ``path`` and ``re_path`` instead of ``url``.:
+
+.. _`Django > 2.0`: https://docs.djangoproject.com/en/2.0/releases/2.0/#simplified-url-routing-syntax
+.. _`Django > 1.9`: https://docs.djangoproject.com/en/2.0/releases/1.9/#passing-a-3-tuple-or-an-app-name-to-include
+
+.. code-block:: django
+
+    from django.contrib import admin
+    from django.urls import path, include, re_path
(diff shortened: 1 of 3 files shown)


docs/source/internals/getting_started.rst
```diff
@@ -186,6 +186,28 @@ you will also need to include Django's i18n URLs:
         url(r'', include(application.urls)),
     ]
 
+For `Django > 1.9`_ pass ``application.urls`` directly to ``path``. For `Django > 2.0`_ use ``path`` and ``re_path`` instead of ``url``.:
+
+.. _`Django > 2.0`: https://docs.djangoproject.com/en/2.0/releases/2.0/#simplified-url-routing-syntax
+.. _`Django > 1.9`: https://docs.djangoproject.com/en/2.0/releases/1.9/#passing-a-3-tuple-or-an-app-name-to-include
```
solarissmoke: We don't support Django 1.10 and below, so this comment is not necessary. Actually we need to drop all uses of `include()` with 3-tuples because it is deprecated in Django 1.11.

There are a few other places in the documentation that need to be updated to stop using `include()`: `how_to_change_a_url.rst`, `how_to_disable_an_app_or_feature.rst`.

docs/source/internals/getting_started.rst
```diff
@@ -186,6 +186,28 @@ you will also need to include Django's i18n URLs:
...
+
+.. code-block:: django
+
+    from django.contrib import admin
+    from django.urls import path, include, re_path
+    from oscar.app import application
+
+    urlpatterns = [
+        re_path('^i18n/', include('django.conf.urls.i18n')),
```
solarissmoke: There is no need for `re_path` - if we're simplifying then we can simplify to `path('i18n/', ...)`

raratiru:
Thank you for the comments. I have committed the proposed changes.

Issue #2761, opened 2018-06-28
Input validation error in BenefitForm


PhilipKazmeier:
### Issue Summary

By defining a custom incentive, the input validation of the offer creation allows to input values that, lead to an error. 
After this, the whole offers page of the dashboard is inaccessible, because of an error message.

### Steps to Reproduce

Prerequisite: Existing custom incentive

1. Start creating a new offer in the dashboard
2. Set random name, e.g. "Test"
3. For the incentive just select "Range" with any of the provided options and leave the rest blank
4. Submit the incentive
5. Fill the rest of the procedure with any data
6. Submit the offer
7. Dashboard shows an error, and offers page is no longer accessible without an error

### Technical details
Usage of sandbox version as described in the [docs][sandbox]. No further modifications needed.

```
$ python --version
Python 3.6.3
$ pip show django | grep Version
Version: 2.0.6
$ pip show django-oscar | grep Version
Version: 2.0.dev0
```

The problem in this case, is that [clean][clean-method] method of the BenefitForm, does only check if one of the values is set. By setting the range, this check is fulfilled.

### Fix
I submitted a simple PR #2762 for fixing this input validation error.

[sandbox]: https://django-oscar.readthedocs.io/en/latest/internals/sandbox.html#running-the-sandbox-locally
[clean-method]: https://github.com/django-oscar/django-oscar/blob/06323399b0dfdcbbf63606a3607ca9648a991cca/src/oscar/apps/dashboard/offers/forms.py#L137

solarissmoke:
Fixed in #2762.

PR #2762, opened 2018-06-28
Fix input validation for BenefitForm in dashboard.offers

PhilipKazmeier:
This PR fixes a problem in the input validation of the BenefitForm as described in the related issue #2761.

I chose the same text as for no input at all, to not add new texts, that have to be translated.


solarissmoke:
Thanks for this - the comprehensive tests you've added for the form are much appreciated!

PR #2768, opened 2018-07-09
Fixes wrong fieldname in StockRecordForm, causing the field to not be removed.

specialunderwear:
Fixes wrong fieldname in StockRecord form.


specialunderwear:
ok merge dan

PR #2777, opened 2018-07-29
fix: usr: handle exception if SMTP Recipient is refused by server

pawelmarkowski:
Service fails if customer has a fake email address - missing try except block
`
File "/home/pmarkowski/workspace/mtechnology/shop/venv/lib/python3.6/site-packages/oscar/apps/customer/utils.py", line 123, in send_email_messages
email.send()
File "/home/pmarkowski/workspace/mtechnology/shop/venv/lib/python3.6/site-packages/django/core/mail/message.py", line 294, in send
return self.get_connection(fail_silently).send_messages([self])
File "/home/pmarkowski/workspace/mtechnology/shop/venv/lib/python3.6/site-packages/django/core/mail/backends/smtp.py", line 110, in send_messages
sent = self._send(message)
File "/home/pmarkowski/workspace/mtechnology/shop/venv/lib/python3.6/site-packages/django/core/mail/backends/smtp.py", line 126, in _send
self.connection.sendmail(from_email, recipients, message.as_bytes(linesep='\r\n'))
File "/usr/lib64/python3.6/smtplib.py", line 881, in sendmail
raise SMTPRecipientsRefused(senderrs)
smtplib.SMTPRecipientsRefused: {'wemiefwnifewnwfin@efwfewefw.pl': (450, b'4.1.2 wemiefwnifewnwfin@efwfewefw.pl: Recipient address rejected: Domain not found')}
ERROR 2018-07-29 14:54:55,237 [/home/pmarkowski/workspace/mtechnology/shop/venv/lib/python3.6/site-packages/django/core/servers/basehttp.py:124]
-- "POST /platnosci/accept/6/ HTTP/1.1" 500 226053
`

### DIFF
--- a/src/oscar/apps/customer/utils.py
+++ b/src/oscar/apps/customer/utils.py
@@ -1,4 +1,5 @@
 import logging
+import smtplib
 
 from django.conf import settings
 from django.contrib.auth.tokens import default_token_generator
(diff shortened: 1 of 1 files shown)


solarissmoke:
Not sure this is the right approach here - Oscar should be backend agnostic. I'd prefer that projects customise the SMTP backend if they want to adjust how exceptions raised by the SMTP handler are handled.

pawelmarkowski:
@solarissmoke I understand your point of view. Nevertheless, it is Oscar role to handle exception that is raised by SMTP backend (in that case)...  but it is my opinion.

sasha0:
Agree with @solarissmoke - we can't proceed with this approach, since it's a project specific and takes into account only SMTP backend. Furthermore, user can decide whether to handle exception or to pass `fail_silently=True`.

sasha0:
Closing per reason above.

PR #2779, opened 2018-08-06
(FIX) dashboard reviews sorting by date

dfirst:
Add DatePickerInput widget to dashboard reviews sorting form


solarissmoke:
Thanks!

PR #2784, opened 2018-08-09
Improve charge computation in `WeightBased` shipping method

samitnuk:
When `remaining_weight` is "0" (zero) we should not get band for it.

### DIFF
--- a/src/oscar/apps/shipping/abstract_models.py
+++ b/src/oscar/apps/shipping/abstract_models.py
@@ -151,13 +151,16 @@ class AbstractWeightBased(AbstractBase):
             return D('0.00')
 
         top_band = self.top_band
-        if weight < top_band.upper_limit:
+        if weight <= top_band.upper_limit:
             band = self.get_band_for_weight(weight)
             return band.charge
         else:
             quotient, remaining_weight = divmod(weight, top_band.upper_limit)
-            remainder_band = self.get_band_for_weight(remaining_weight)
-            return quotient * top_band.charge + remainder_band.charge
+            if remaining_weight:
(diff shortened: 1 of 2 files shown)


src/oscar/apps/shipping/abstract_models.py
```diff
@@ -156,8 +156,11 @@ def get_charge(self, weight):
             return band.charge
         else:
             quotient, remaining_weight = divmod(weight, top_band.upper_limit)
-            remainder_band = self.get_band_for_weight(remaining_weight)
```
solarissmoke: Wouldn't be simpler just to change line 154 above to `if weight <= top_band.upper_limit`. That way there is no possibility of ending up inside this `else` block if there is no remainder.

samitnuk: @solarissmoke 

Using `<=` instead of `<` in line 154 useful only when `weight == top_band.upper_limit`:
- with `<=` calculations will be made in `if`-block
- with `<` calculations will be made in `else`-block

But with changes in this PR, we will receive the correct value with both `<` and `<=`.

I will change `<` to `<=`, since now I see that for the case when `weight == top_band.upper_limit` we should use `if`-block (with fewer calculations). Thank you for pointing out this opportunity.

But using `<=` instead of `<` will not help when `weight > top_band.upper_limit`, since we will be in `else`-block in this case. Please see created test (I will add additional checks to it).

Maybe I missed something. Please comment. Thanks.

solarissmoke: Thanks for the explanation - I see what you mean now. We still need your additional logic when `weight` is a multiple of `top_band.upper_limit`.

PR #2795, opened 2018-08-20
Fix address hash generate bug and success_url consistent

davisnando:
Success_url consistent on the views i saw that not every view in the checkout flow uses a success_url so now it will

I also ran into a bug when generating a hash it. It wanted to save a new user address but the hash already existed so it crashed the checkout flow


solarissmoke:
Could you please separate the two changes into individual PRs? 

For the address logic, can you also provide steps to reproduce the error? We will also need unit tests that cover the case you are trying to fix. 

As regards the `success_url` change, you need to use `reverse_lazy`, as `reverse` will run before URLs are loaded.

PR #2796, opened 2018-08-20
How about this pinning strategy?

specialunderwear:
How about making defensive pinning optional, but the default in the documentation?

### DIFF
--- a/docs/source/internals/getting_started.rst
+++ b/docs/source/internals/getting_started.rst
@@ -14,14 +14,22 @@ scratch and have decided to use Oscar.  Let's call this shop 'frobshop'
 Install Oscar and its dependencies
 ==================================
 
+
 Install Oscar (which will install Django as a dependency), then create the
 project:
 
 .. code-block:: bash
 
     $ mkvirtualenv oscar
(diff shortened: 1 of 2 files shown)


solarissmoke:
As the person who currently responds to the majority of support requests, I don't think that defaulting to unpinned is a good idea. I think the default needs to be pinned, with an option to use `unsafe` dependencies if you want to. Otherwise every newbie who installs Oscar without reading the documentation is going to start with a broken installation, and we will be inundated with support issues.

No issues with doing it the other way around though.

specialunderwear:
The other way around does not work.

solarissmoke:
In that case IMHO, this isn't a good idea. Doing `pip install django-oscar` will almost certainly result in unsupported dependencies being installed (hence why the tests fail on this PR), which I don't think is a good experience for a first time installer.

If it's really an issue for you then perhaps a fork which doesn't pin dependencies is a better option?

specialunderwear:
I tried and because they are extras they don't override so you can not remove a defensive pinning.

solarissmoke:
Closing in favour of discussion on #2797.

PR #2797, opened 2018-08-20
How about this pinning strategy then?

specialunderwear:
This one splits oscar in a metapackage with the pins and a code package wihtout them.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@ VENV = venv
 PYTEST = $(PWD)/$(VENV)/bin/py.test
 
 # These targets are not files
-.PHONY: install sandbox docs coverage lint messages compiledmessages css clean sandbox_image
+.PHONY: install sandbox docs coverage lint messages compiledmessages css clean sandbox_image build-tools
 
 install:
 	pip install -r requirements.txt
@@ -83,6 +83,7 @@ clean:
 	# Remove files not in source control
 	find . -type f -name "*.pyc" -delete
 	rm -rf nosetests.xml coverage.xml htmlcov *.egg-info *.pdf dist violations.txt
+	rm -rf dist
 
 todo:
 	# Look for areas of the code that need updating when some event has taken place (like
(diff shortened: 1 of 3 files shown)


specialunderwear:
Obviously this can not be merged as such, because the package name 'oscar' has allready been taken. So someone needs to pick a name for the code package. I just picked the most obvious name.

solarissmoke:
I don't really have an issue with the metapackage approach. 

My main question would still be what the value of a totally unpinned package is? If you try to install this unpinned package without any local pinning, things will definitely break. So, you still need some way to know what Oscar officially supports before you go and tweak your requirements. That being the case, what is wrong with just relying on Oscar's requirements and overriding them at project level where required (this will result in dependency warnings in pip, but it will still let you override)?

specialunderwear:
It is not totally unpinned it has minimum requirements. If you run projects as python eggs with a setup.py and with console_scripts you can not cheat like with pip. If you try to use the console script it will break with an exception about being unable to satisfy dependencies.

solarissmoke:
Hmm. `pip` will still install the package using `setup.py` (not `metasetup.py`) which means that even if you install the "safe" package you end up with unpinned dependencies. This is why the tests are failing - because an unsupported version of django-tables is being installed. I am not very familiar with the internals of `pip`, but am not aware of a way to tell it to use something other than `setup.py`?

Personally I think the most maintainable solution here is just to set up a fork.

solarissmoke:
@craigloftus @sasha0 thoughts?

specialunderwear:
@craigloftus we are in fact building libraries on top of oscar. Oscar is just a library and it should behave well. Currently only a 1 workflow that includes pip is supported, but there are many more ways in which python packages can be used. It would be really strange to just neglect that.

specialunderwear:
I would also like to note that oscar is a library and not some finished product. That means it has to behave nicely together with other libraries, currently it just doesn't. To me this is a rather elegant way to make oscar behave well as a library but still have safe defaults.

specialunderwear:
#3027 can be solved this way also, because you could just add sorl to the django-oscar metapackage as a dependency, but exclude it from the 'oscar-core' or whatever package.

specialunderwear:
I still believe this is the way to go. I suggest we call the new package django-oscar-unpinned the django-oscar package can then depend on django-oscar-unpinned and add all the pins. django-oscar-unpinned should only specify LOWER BOUND pins and not upper bound.

PR #2804, opened 2018-08-24
Added setting for calculating offer discounts over the price including tax.

specialunderwear:
This makes offers including tax accessible.
The incl_tax parameter used to be hardcoded to False in apply_discount. ?

In creating the tests for this thing, I found the current calculation, over
the price excluding tax to be incorrect. I added the test and
marked it as skipped.

### DIFF
--- a/src/oscar/apps/basket/abstract_models.py
+++ b/src/oscar/apps/basket/abstract_models.py
@@ -11,7 +11,7 @@ from django.utils.translation import gettext_lazy as _
 
 from oscar.core.compat import AUTH_USER_MODEL
 from oscar.core.loading import get_class, get_classes
-from oscar.core.utils import get_default_currency
+from oscar.core.utils import get_default_currency, round_half_up
 from oscar.models.fields.slugfield import SlugField
 from oscar.templatetags.currency_filters import currency
 
@@ -736,7 +736,7 @@ class AbstractLine(models.Model):
             item_incl_tax_discount = (
                 self.discount_value / int(self.consumer.consumed()))
             item_excl_tax_discount = item_incl_tax_discount * self._tax_ratio
-            item_excl_tax_discount = item_excl_tax_discount.quantize(D('0.01'))
+            item_excl_tax_discount = round_half_up(item_excl_tax_discount)
             prices.append((self.unit_price_incl_tax - item_incl_tax_discount,
                            self.unit_price_excl_tax - item_excl_tax_discount,
                            self.consumer.consumed()))
@@ -846,8 +846,9 @@ class AbstractLine(models.Model):
             # against tax inclusive prices but we need to guess how much of the
             # discount applies to tax-exclusive prices.  We do this by
             # assuming a linear tax and scaling down the original discount.
-            return self.line_price_excl_tax \
-                - self._tax_ratio * self._discount_incl_tax
+            return self.line_price_excl_tax - round_half_up(
+                self._tax_ratio * self._discount_incl_tax
+            )
         return self.line_price_excl_tax
 
     @property
@@ -855,13 +856,17 @@ class AbstractLine(models.Model):
         # We use whichever discount value is set.  If the discount value was
         # calculated against the tax-exclusive prices, then the line price
         # including tax
-        if self.line_price_incl_tax is not None:
-            return self.line_price_incl_tax - self.discount_value
+        if self.line_price_incl_tax is not None and self._discount_incl_tax:
+            return self.line_price_incl_tax - self._discount_incl_tax
+        elif self.line_price_excl_tax is not None and self._discount_excl_tax:
+            return round_half_up((self.line_price_excl_tax - self._discount_excl_tax) / self._tax_ratio)
+
+        return self.line_price_incl_tax
 
     @property
     def line_tax(self):
         if self.is_tax_known:
-            return self.quantity * self.unit_tax
+            return self.line_price_incl_tax_incl_discounts - self.line_price_excl_tax_incl_discounts
 
     @property
     def line_price_incl_tax(self):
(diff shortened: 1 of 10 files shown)


solarissmoke:
@specialunderwear I don't know what lint errors you are getting, but suggest you ignore the ones from Codacy, as it is currently using daft rules. As long as the Travis lint build passes you're fine.

I would also add that this sort of change will need documentation.

specialunderwear:
Please have a look at the skipped test I added. I don't know if you want me to fix these issues in this PR as that would change the behaviour of the way discounts work now. I do need them fixed though ...

tests/unit/basket/test_offers.py
```diff
@@ -0,0 +1,162 @@
...
+    def test_offer_incl_tax(self):
+        "The offer should be calculated as if it was declared including tax"
+        with self.settings(OSCAR_OFFERS_INCL_TAX=True):
+            self.assertEqual(Basket.objects.count(), 0)
+
+            admin = User.objects.get(username="admin@admin.nl")
+            self.client.force_login(admin)
+
+            # throw an item in the basket
+            basket_add_url = reverse("basket:add", args=(2,))
```
solarissmoke: Rather than faking all this request logic, it will be much easier to user `oscar.test.factories.create_basket` and then manipulate the basket with `oscar.test.basket.add_product`.

tests/unit/fixtures/auth.json
```diff
@@ -0,0 +1,38 @@
+[
+{
+ "model": "myauth.User",
```
solarissmoke: Don't think these are necessary if you use the test utils that oscar provides for testing baskets.

specialunderwear:
I would write documentation, but it would be very hard to describe what oscar does with ``OSCAR_OFFERS_INCL_TAX=False`` (this is the current default settings). So I could describe what I think it SHOULD do, but the tests show there are some issues there.

solarissmoke:
Yeah, it looks like there are several issues with the offer logic which need to be fixed. If you have time to dig into this particular tax-related issue that would be great. Oscar 2 is going to be making various breaking changes, so now is a good time to change behaviour if it needs changing.

specialunderwear:
Ok I will add that to this PR and make the tests pass.

specialunderwear:
@solarissmoke I made the skipped test pass. Some more accuracy could be gained by not computing _tax_ratio and instead storing it on purchase_info, but let's start here.

tests/unit/basket/test_offers.py
```diff
@@ -0,0 +1,159 @@
...
+    basket.reset_offer_applications()
+    if not basket.is_empty:
+        Applicator().apply(basket, user, request)
+    request.session[settings.OSCAR_BASKET_COOKIE_OPEN] = basket.pk
+    request.session.save()
+    return basket
+
+
+class OfferTest(TestCase):
+    fixtures = ["catalogue", "offer"]
```
sasha0: I suppose we can just use factories for catalogue models and generate necessary offers in the test utils instead of using fixture, which would need to be updated within the models changes.

specialunderwear: Hi @sasha0, thanks for your feedback, understandable considering the way the rest of oscar's tests are implemented. However,  I think updating the fixtures is much, much safer because they are actually made with the user interface, which means they are a real representation of data users produce. It would be much, much more difficult to update the factories to reflect real offers, especially after changes to the offer machinery (models included). For example there is a field called 'exclusive' which was recently added. Without understanding what that field is for, factory boy would just put some random value or the default in there. This could lead to testing of non real world scenarios. I know fixtures are frowned upon by many people, but they are explicit and they are explicitly updated when models change and I think that is a good thing instead of a burden, See zen of python.

src/oscar/apps/basket/abstract_models.py
```diff
@@ -736,7 +736,7 @@ def get_price_breakdown(self):
             item_incl_tax_discount = (
                 self.discount_value / int(self.consumer.consumed()))
             item_excl_tax_discount = item_incl_tax_discount * self._tax_ratio
-            item_excl_tax_discount = item_excl_tax_discount.quantize(D('0.01'))
+            item_excl_tax_discount = round_half_up(item_excl_tax_discount)
```
solarissmoke: Just looking at the historic way in which stuff has been rounded, I think we should probably round *down* here and everywhere else that we are working out a discount. That is the more conservative approach (i.e., you will never end up giving a higher discount that you intended to).

specialunderwear: I will have to do some more testing to check that out. Do you mean really round down eg. 1.499 = 1.49 or do you mean round half down eg. 1.499 = 1.50 but 1.495 = 1.49 ?

solarissmoke: Just realised that is this setting for offer rounding - we should probably be using that? https://github.com/django-oscar/django-oscar/blob/f2bb0fb73c83cb879fb2dec9184059dd96d814e0/docs/source/ref/settings.rst#oscar_offer_rounding_function

specialunderwear: I looked into that before and the only place it is used is in AbstractBenefit. Note that I didn't really change anything, I only fixed usage of the really weird and erroneous behaviour of the default rounding function. It has to be investigated if this is the right place to apply the round down. More testing would be needed.

solarissmoke: Spent a bit more time looking at this, and I think what you have is fine.

specialunderwear: The value being rounded is the item_excl_tax_discount, which is the item price. If I round it down there is actually MORE discount given in most cases. I don't think that is great idea.

PR #2812, opened 2018-08-29
allow OSCAR_OFFER_ROUNDING_FUNCTION to be defined as string path to the rounding function

v1kku:
to avoid importing functions in settings file


src/oscar/apps/offer/abstract_models.py
```diff
@@ -625,8 +627,16 @@ def round(self, amount):
         """
         Apply rounding to discount amount
         """
-        if hasattr(settings, 'OSCAR_OFFER_ROUNDING_FUNCTION'):
-            return settings.OSCAR_OFFER_ROUNDING_FUNCTION(amount)
+        rounding_function = getattr(settings, 'OSCAR_OFFER_ROUNDING_FUNCTION', None)
+        if callable(rounding_function):
+            return rounding_function(amount)
```
solarissmoke: Please add a deprecation warning here:

    warnings.warn("Defining OSCAR_OFFER_ROUNDING_FUNCTION as a function is deprecated. Specify a dotted path to the function instead." , RemovedInOscar21Warning, stacklevel=1)

Please also update the [documentation](https://github.com/django-oscar/django-oscar/blob/f2bb0fb73c83cb879fb2dec9184059dd96d814e0/docs/source/ref/settings.rst#offer-settings) to say that this expects a dotted path.

src/oscar/apps/offer/abstract_models.py
```diff
@@ -625,8 +627,16 @@ def round(self, amount):
...
+        elif isinstance(rounding_function, six.string_types):
+            try:
+                rounding_function = import_string(rounding_function)
```
solarissmoke: Think we should consider doing this once and wrapping it in an `lru_cache`, the way we do here: https://github.com/django-oscar/django-oscar/blob/master/src/oscar/core/loading.py#L34-L36

Reference: https://wiki.python.org/moin/PythonSpeed/PerformanceTips#Import_Statement_Overhead

src/oscar/apps/offer/utils.py
```diff
@@ -33,3 +38,20 @@ def load_proxy(proxy_class):
     except AttributeError:
         raise exceptions.ImproperlyConfigured(
             "Module %s does not define a %s" % (module, classname))
+
+
+@lru_cache(maxsize=None)
+def get_offer_rounding_function():
```
craigloftus: I think it would be worth making this a core util usable for loading any dotted path string with caching?

I'd suggest dropping the deprecation warning, and make this a 2.0 breaking change?

PR #2813, opened 2018-08-30
Optimize Range checks, and fix missing excluded_products in range.all_products()

vdboor:
See the commit messages of each individual commit.

This:
* optimized the range queries, so `range.contains_product()` works more efficiently.
* Fixes a inconsistency, where `range.all_products()` missed didn't exclude the blacklisted products when `range.includes_all_products=True`
* Makes sure not the full category data is pulled from the database when only a contains/is-descendant check is performed.

### DIFF
--- a/tests/integration/offer/test_range.py
+++ b/tests/integration/offer/test_range.py
@@ -22,10 +22,12 @@ class TestWholeSiteRange(TestCase):
     def test_whitelisting(self):
         self.range.add_product(self.prod)
         self.assertTrue(self.range.contains_product(self.prod))
+        self.assertIn(self.prod, self.range.all_products())
 
     def test_blacklisting(self):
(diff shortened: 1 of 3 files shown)


tests/integration/offer/test_range.py
```diff
@@ -22,10 +22,12 @@ def test_all_products_excludes_child_products(self):
     def test_whitelisting(self):
         self.range.add_product(self.prod)
         self.assertTrue(self.range.contains_product(self.prod))
+        self.assertTrue(self.prod, self.range.all_products())
```
solarissmoke: Shouldn't this be `assertIn()`?

vdboor: You're totally right! Thanks for spotting

PR #2815, opened 2018-08-31
Update Psycopg2 dependency.

sasha0:


solarissmoke:
@sasha0 `psycopg2` is not deprecated - it just issues a `UserWarning` because they are dropping the wheel package from this package from version 2.8 onwards. Use of the binary package is not recommended - see http://initd.org/psycopg/articles/2018/02/08/psycopg-274-released/ .

We should instead put `psycopg2 --no-binary` in requirements IMO, until they drop the warning.

Issue #2818, opened 2018-09-06
Going back to previous steps in OfferWizard shows confusing error message


SrdjanCosicPrica:
Found a bug? Please fill out the sections below.

### Issue Summary

When creating a new offer you can go back to steps 2 (Incentive) and 3 (Condition) after selecting a range and both will complain that a selected range is an invalid choice.

<img width="689" alt="incentive before" src="https://user-images.githubusercontent.com/15933115/45168853-97b69380-b1fc-11e8-8457-a86c50727cca.png">

### Steps to Reproduce

It's essential that you provide enough information for someone else to replicate the problem you're seeing. Simply describing something that's broken on your current project is not enough!

1. run command `make sandbox`
2. Press the Offers Tab and Select Offers option
3. Press "Create new Offer"
4. Add a name and continue to step 2
5. Press Range and Select "Site"
6. Press Type and Select "Discount is a percentage off of the product's value"
7. Set Value to 10
8. Press "Continue to step 3"
9. In the "Steps" Panel, select "Incentive"

You should see the same message as the image posted.
**Note**: This applies for the Condition Step as well.

I think this is a bug since it is confusing when it says that it is not a valid choice but still shows as selectable and is already selected.

### Technical details

* Python version: 3.6.4
* Django version: 1.11.15 & 2.0.7 & 2.1.1
* Oscar version: 1.6.2 & 2.0.dev0

PR #2819, opened 2018-09-06
Fix OfferWizard Visual Bug in OfferBenefitForm and OfferConditionForm

SrdjanCosicPrica:
A fix for the Issue I created #2818

**Before**
incentive before
 
**After**
incentive after

It also reduced the number of queries by one
incentive query before
incentive query after

### DIFF
--- a/src/oscar/apps/dashboard/offers/views.py
+++ b/src/oscar/apps/dashboard/offers/views.py
@@ -113,8 +113,7 @@ class OfferWizardStepView(FormView):
         form_data = form.cleaned_data.copy()
         range = form_data.get('range', None)
         if range is not None:
-            form_data['range_id'] = range.id
-            del form_data['range']
+            form_data['range'] = range.id
         form_kwargs = {'data': form_data}
         json_data = json.dumps(form_kwargs, cls=DjangoJSONEncoder)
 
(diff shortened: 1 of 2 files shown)


src/oscar/apps/dashboard/offers/views.py
```diff
@@ -113,8 +113,7 @@ def _store_form_kwargs(self, form):
         form_data = form.cleaned_data.copy()
         range = form_data.get('range', None)
         if range is not None:
-            form_data['range_id'] = range.id
```
SrdjanCosicPrica: Since the ID is saved rather than the object, the name might be confusing. I have a solution because I thought this might be down to personal taste.

If you wish, I can leave this part as it is but then on line 128-> I could do this

```
form_kwargs = json.loads(json_data)
if 'range_id' in form_kwargs['data']:
    form_kwargs['data']['range'] = form_kwargs['data']['range_id']
    del form_kwargs['data']['range_id']
    return form_kwargs
```

solarissmoke:
Thanks for this - it looks good to me. Can you please add a test in https://github.com/django-oscar/django-oscar/blob/master/tests/functional/dashboard/test_offer.py that verifies this fix - i.e., that going to step 3 and then going back doesn't result in errors on the step 2 form?

SrdjanCosicPrica:
There, I had to divert from using the form variable from WebTest because I could not find any errors through that. Please tell me what you think of the naming of the test cases.

solarissmoke:
Looks good - thank you!

PR #2820, opened 2018-09-07
Decimal must be imported

Vitalynx:


docs/source/topics/prices_and_availability.rst
```diff
@@ -13,6 +13,7 @@ Simpler e-commerce frameworks often tie prices to the product model directly:
 
 .. code-block:: python
 
+   >>> from decimal import Decimal
```
solarissmoke: `Decimal` is not used in this code sample (the last line is output, not input).

docs/source/topics/prices_and_availability.rst
```diff
@@ -24,6 +25,7 @@ calculating prices and availability information.
 .. code-block:: python
 
    >>> from oscar.apps.partner.strategy import Selector
+   >>> from decimal import Decimal
```
solarissmoke: Same here - we don't actually use `Decimal` anywhere below.

docs/source/topics/prices_and_availability.rst
```diff
@@ -280,6 +282,7 @@ Here's an example ``strategy.py`` module which is used to charge VAT on prices.
     # myproject/partner/strategy.py
 
     from oscar.apps.partner import strategy, prices
+    from decimal import Decimal as D
```
solarissmoke: We should use PEP8 import order - i.e., this should come above the `oscar` import.

Issue #2814, opened 2018-08-30
Adding product variants


Vitalynx:
Why is it nearly impossibly hard to add a product variant through the dashboard? At the very least provide some documentation on how one should go about making product variants, especially considering one of the most sold items on the internet include items with sizes..

I create a new product and the system whines about "One can't add variants to this product at this point. This is likely because this product still has stock records."

Even when creating a product under a product type which doesn't track stock levels, the same problem occurs...

So my question is, how does one create product variants? Like having shirts in multiple sizes.

shadiakiki1986:
The message "One can't add variants to this product at this point. This is likely because this product still has stock records" means that you added an entry under the product's "stock and pricing". If you delete it, you'll be able to add variants. Note that after adding variants, the "stock and pricing" button of the parent product is no longer available

PR #2821, opened 2018-09-09
Fix confusing variants message on new products

SrdjanCosicPrica:
Fix for #2814. It is a bit confusing when creating the product at first but I don't think any more changes are needed.


SrdjanCosicPrica:
Closed because I accidentally branched from another branch that had changes not in master.

PR #2825, opened 2018-09-12
Enhancement #2737 Track order status changes 

### LINKED ISSUE #2737: Track order status changes
Oscar doesn't currently let you track changes to order status. I think it would be useful to add a model that tracks changes made by `Order.set_status` so that a timestamp can be logged for each status change. 

This allows you to do things such as:

- Track how long an order takes to be processed
- Provide much more detailed order history to a customer
- Set up alerts/monitoring when an order stays too long in a certain status

... etc.

Oscar doesn't have to do the more fancy things, but I think it should provide the basic framework for storing status changes.

SrdjanCosicPrica:
Fixes #2737


SrdjanCosicPrica:
I forgot to add a `prefetch_related` in the detail view. I will add it when I have time

src/oscar/apps/order/abstract_models.py
```diff
@@ -430,6 +430,30 @@ def is_editable(self):
...
+
+    class Meta:
+        abstract = True
+        app_label = 'order'
+        verbose_name = _('Order Status Change')
+        verbose_name_plural = _('Order Status Changes')
+        ordering = ['-date_created']
+
+    def __str__(self):
+        return '{order} has changed status from {old_status} to {new_status}'.format(
```
solarissmoke: This string needs to be translated.

src/oscar/apps/order/config.py
```diff
@@ -6,3 +6,7 @@ class OrderConfig(AppConfig):
     label = 'order'
     name = 'oscar.apps.order'
     verbose_name = _('Order')
+
+    def ready(self):
+        from . import receivers  # noqa
+        super(OrderConfig, self).ready()
```
solarissmoke: I don't think we need a `super()` call here - the parent method is a noop.

src/oscar/apps/order/receivers.py
```diff
@@ -0,0 +1,9 @@
+# coding=utf-8
+from django.dispatch import receiver
+
+from .signals import order_status_changed
+
+
+@receiver(order_status_changed)
+def receive_order_status_change(sender, order, old_status, new_status, **kwargs):
```
solarissmoke: I would prefer to do this in an `Order.create_order_status_change()` method that is called directly from `Order.set_status()`. That makes is much easier to override the behaviour instead of having to unhook signals etc.  @craigloftus @sasha0 do you have any thoughts?

craigloftus: Yeah, I think we're increasingly learning that signals are an anti-pattern

src/oscar/apps/order/config.py
SrdjanCosicPrica: Ok, I just do that for good practice but I'll remove it

src/oscar/apps/order/receivers.py
SrdjanCosicPrica: Okay. Just do be clear, it is an instance method of AbstractOrder and not a classmethod?

solarissmoke: Yes please. This allows access to the order object for anyone that wants to modify how status changes are handled.

SrdjanCosicPrica:
Updated

Issue #536, opened 2013-02-26
Add a 'is_active' flag to the Product model


roadsideseb:
A [discussion on the mailing list](https://groups.google.com/forum/?hl=en-US&fromgroups=#!topic/django-oscar/eJ9TYWbtAbQ) revealed that having setting on the `Product` model to activate/deactivate a product is fairly common. I think that makes it reasonable to add it as an attribute in Oscar rather then have everybody add this flag individually. Any objections?

PR #2827, opened 2018-09-14
Add system configuration to dashboard and fields for product and attribute visibilitiy  

frmdstryr:
Should probably be separate PR's, this adds a few things:

1. Adds a `system` app so that basic site settings can be defined from the dashboard (ie name, tagline, editor options).
2. Adds an `is_enabled` field to Products so they can be temporarily disabled (removed from frontend)
3. Adds a `display` field to Attributes so they can be defined but hidden from the product detail page

Suggestions are welcome. Thanks for the great project!


solarissmoke:
Thanks for this. A few thoughts:

1. Regarding the new fields you propose to add, please see the discussion on #536 where the same thing was proposed. I think the reasoning for closing that issue still applies. Oscar provides the core framework for modelling a catalogue, but expects projects to handle specific use-cases.

2. Moving some configuration to the database may not be a bad idea. However I think it needs some careful thought. The dashboard as it currently works is basically a tool for E-commerce administrators (as opposed to web admins). I think it would be a mistake to allow the person who can manage orders to mess around with config such as whether to use Less, or to change the editor configuration, or to change the home page URL.

   That isn't to say we can't expose these settings through the dashboard - but we need different permissions for these views if we do this. 

   We also need to distinguish between presentational (e.g., shop name, shop tagline) and system-level (e.g., use Less) configuration. IMO only the former should be staff-editable.

PR #2829, opened 2018-09-24
(UI) oscar dashboard options

dfirst:
Here is options dashboard ui.

### DIFF
--- a/src/oscar/apps/dashboard/catalogue/app.py
+++ b/src/oscar/apps/dashboard/catalogue/app.py
@@ -62,6 +62,11 @@ class CatalogueApplication(DashboardApplication):
     attribute_option_group_delete_view = get_class('dashboard.catalogue.views',
                                                    'AttributeOptionGroupDeleteView')
 
+    option_list_view = get_class('dashboard.catalogue.views', 'OptionListView')
+    option_create_view = get_class('dashboard.catalogue.views', 'OptionCreateView')
+    option_update_view = get_class('dashboard.catalogue.views', 'OptionUpdateView')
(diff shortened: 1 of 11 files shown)


src/oscar/templates/oscar/dashboard/catalogue/option/form.html
```diff
@@ -0,0 +1,83 @@
+{% extends 'dashboard/layout.html' %}
```
solarissmoke: To be consistent with all the other templates I'd suggest renaming this to `dashboard/option_form.html`. Similarly `option_list.html`, `option_row_actions.html`.

src/oscar/templates/oscar/dashboard/catalogue/option/form.html
```diff
@@ -0,0 +1,83 @@
...
+    </ul>
+{% endblock %}
+
+{% block headertext %}{{ title }}{% endblock %}
+
+{% block dashboard_content %}
+    <form class="form-stacked wysiwyg fixed-actions" method="post" data-behaviour="affix-nav-errors">
+        {% csrf_token %}
+
+        <div class="row">
```
solarissmoke: Please remove these two `div`s (the `.row` and the `.col-md-12`) - it's redundant to have `col-12` inside a `row`.

dfirst:
Fixed

src/oscar/templates/oscar/dashboard/catalogue/option/form.html
dfirst: Maybe it should be in discuss list? Because currently managing all catalog templates in single folder is difficult.

PR #2836, opened 2018-09-28
Releases/1.4

rufelo:


solarissmoke:
@rufelo I don't think you meant to make this PR to django-oscar/django-oscar.

Issue #2843, opened 2018-10-03
Enhancement  : Add availability check for basket in voucher abstract models


Ayub-Khan:
### Enhancement Summary

Add a method in abstractvoucher model to check if voucher is available/valid for a basket or not.

 `https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/voucher/abstract_models.py#L110`

PR #2844, opened 2018-10-03
Added functionality to check if a voucher is available for basket

Ayub-Khan:
[Enhancement-2843](https://github.com/django-oscar/django-oscar/issues/2843)

Added functionality to check if a voucher is available for a basket or not.

Tests : Done
Docs : Not sure!

### DIFF
--- a/src/oscar/apps/voucher/abstract_models.py
+++ b/src/oscar/apps/voucher/abstract_models.py
@@ -219,6 +219,25 @@ class AbstractVoucher(models.Model):
                                 "a previous order")
         return is_available, message
 
+    def is_available_for_basket(self, basket):
+        """
+        Tests whether this voucher is available to the passed basket.
+
+        Returns a tuple of a boolean for whether it is successful, and a
+        availability message.
+        """
+        is_available, message = self.is_available_to_user(user=basket.owner)
+        if not is_available:
+            return False, message
+
+        is_available, message = False, _("This voucher is not available for this basket")
+        for offer in self.offers.all():
+            if offer.is_condition_satisfied(basket=basket):
+                is_available = True
+                message = ''
+                break
+        return is_available, message
+
     def record_usage(self, order, user):
         """
         Records a usage of this voucher in an order.
--- a/tests/integration/voucher/test_models.py
+++ b/tests/integration/voucher/test_models.py
@@ -9,7 +9,16 @@ from django.utils.timezone import utc
 from oscar.apps.voucher.models import Voucher
 from oscar.core.compat import get_user_model
 from oscar.test.factories import (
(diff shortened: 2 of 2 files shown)


solarissmoke:
I am not convince by the utility of this method. It seems like you can achieve the same thing just by calling `voucher.is_available_to_user(basket.owner)`?

Ayub-Khan:
@solarissmoke  this can be used to replace the following code. https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/basket/views.py#L340-L360 . I assume that if a voucher has valid offer it gives discount. Plus this can be used by other projects built on django-oscar and have to check voucher validation for basket in different views. This can also help them to override to add basket base checks. Also adding voucher and thn checking the discount and removing if not valid just to validate the voucher does not seem very appropriate.

src/oscar/apps/voucher/abstract_models.py
```diff
@@ -219,6 +219,24 @@ def is_available_to_user(self, user=None):
                                 "a previous order")
         return is_available, message
 
+    def is_available_for_basket(self, basket):
+        """
+        Tests whether this voucher is available to the pased basket.
```
craigloftus: You have a typo here, `pased`

src/oscar/apps/voucher/abstract_models.py
```diff
@@ -219,6 +219,24 @@ def is_available_to_user(self, user=None):
...
+
+        Returns a tuple of a boolean for whether it is successful, and a
+        availability message.
+        """
+        is_available, message = False, _("This voucher is not available for Basket")
```
craigloftus: Can you change this message to "This voucher is not available for this basket".

src/oscar/apps/voucher/abstract_models.py
```diff
@@ -219,6 +219,24 @@ def is_available_to_user(self, user=None):
...
+        valid_user, __ = self.is_available_to_user(user=basket.owner)
+        if not valid_user:
+            message = _("This voucher is not available to user")
```
craigloftus: And this should be using the message returned by `is_available_to_user`.

Ayub-Khan:
@craigloftus Thanks for review. I have updated the pr. i did run the 'make docs' command. Are there any files where i will have to write those doc details or doc strings are enough.

Ayub-Khan:
@specialunderwear can you review this too?. Let me know if there is any thing else i need to do.

highbiza:
looks good to me

specialunderwear:
looks good to me

Ayub-Khan:
@specialunderwear what do i need to do to get this merged?

specialunderwear:
Ask somebody that can actually commit to master?

PR #2849, opened 2018-10-09
Make file Modifications

Ayub-Khan:
-Added help messages for all targets
-Added help target
-Divided targets into blocks for easy understanding
-Added new target install-js for npm install
-Used new target in install and css
-divided install target into 3 targets
    install-python
    install-test
    install-js
-Updated i18n traget names
-Added install-migrations-testing-requirements target
-Used install-migrations-testing-requirements in test_migrations
-Updated .PHONY

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,15 +2,44 @@ VENV = venv
 PYTEST = $(PWD)/$(VENV)/bin/py.test
 
 # These targets are not files
-.PHONY: install sandbox docs coverage lint messages compiledmessages css clean sandbox_image
+.PHONY: build_sandbox clean compile_translations coverage css docs extract_translations help install install-python \
+ install-test install-js lint release retest sandbox_clean sandbox_image sandbox test todo venv
 
-install:
+help: ## Display this help message
+	@echo "Please use \`make <target>\` where <target> is one of"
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -2,15 +2,44 @@ VENV = venv
 PYTEST = $(PWD)/$(VENV)/bin/py.test
 
 # These targets are not files
-.PHONY: install sandbox docs coverage lint messages compiledmessages css clean sandbox_image
+.PHONY: build_sandbox clean compile_translations coverage css docs extract_translations help install install-python \
+ install-test install-js lint release retest sandbox_clean sandbox_image sandbox test todo venv
 
-install:
+help: ## Display this help message
```
craigloftus: I think the PR generally looks good, but I am not sure about the value of this `help` command, and the complexity of the implementation makes me nervous. Does it add enough value over just reading the make file to justify maintaining it?

Also, I'm assuming from trying to parse it that it requires perl as a system dependency?

Ayub-Khan: ah, i get your point. and yes it requires perl. 
i will remove this.

Ayub-Khan: @specialunderwear Yeah i agree, those are useful. I have used grep to get strings but again i think this will fail on windows. @craigloftus what you think, we can skip this if you think its not worth it.

Ayub-Khan:
@solarissmoke Thanks.

PR #2895, opened 2018-12-04
 Improve efficiency of category_tree template tag.

highbiza:
1. cache full_slug instead of get_absolute_url so parent slug info can be re-used
   to create child slugs, instead of having to query all parents for every child slug.
2. do not hit database or cache for get_absolute_url
3. do not allow expensive tree walking in templates.
4. warmup category url cache while we are generating the urls anyway.

These changes will make the category tree use only 1 query, even when the
cache is not warmed up yet. It also prevents people from making really
expensive queries in templates by using mptt methods.

For sites with a lot of categories (300) this change will shave away hundreds of
useless queries. Now we can go back to having short cache durations and shitloads
of categories.

### DIFF
--- a/src/oscar/apps/catalogue/abstract_models.py
+++ b/src/oscar/apps/catalogue/abstract_models.py
@@ -115,6 +115,19 @@ class AbstractCategory(MP_Node):
         names = [category.name for category in self.get_ancestors_and_self()]
         return self._full_name_separator.join(names)
 
+    def get_full_slug(self, parent_slug=None):
+        if self.is_root():
+            return self.slug
+
+        cache_key = self.get_url_cache_key()
+        full_slug = cache.get(cache_key)
+        if full_slug is None:
+            parent_slug = parent_slug if parent_slug is not None else self.get_parent().full_slug
+            full_slug = "%s%s%s" % (parent_slug, self._slug_separator, self.slug)
+            cache.set(cache_key, full_slug)
+
+        return full_slug
+
     @property
     def full_slug(self):
         """
@@ -125,8 +138,7 @@ class AbstractCategory(MP_Node):
         has been re-purposed to only store this category's slug and to not
         include it's ancestors' slugs.
         """
-        slugs = [category.slug for category in self.get_ancestors_and_self()]
-        return self._slug_separator.join(slugs)
+        return self.get_full_slug()
 
     def generate_slug(self):
         """
@@ -154,6 +166,9 @@ class AbstractCategory(MP_Node):
         if you don't want to include the category itself. It's a separate
         function as it's commonly used in templates.
         """
+        if self.is_root():
+            return [self]
+
         return list(self.get_ancestors()) + [self]
 
     def get_descendants_and_self(self):
@@ -169,7 +184,7 @@ class AbstractCategory(MP_Node):
         cache_key = 'CATEGORY_URL_%s_%s' % (current_locale, self.pk)
         return cache_key
 
-    def get_absolute_url(self):
+    def get_absolute_url(self, parent_slug=None):
         """
         Our URL scheme means we have to look up the category's ancestors. As
         that is a bit more expensive, we cache the generated URL. That is
@@ -178,14 +193,9 @@ class AbstractCategory(MP_Node):
         you change that logic, you'll have to reconsider the caching
         approach.
         """
-        cache_key = self.get_url_cache_key()
-        url = cache.get(cache_key)
-        if not url:
-            url = reverse(
-                'catalogue:category',
-                kwargs={'category_slug': self.full_slug, 'pk': self.pk})
-            cache.set(cache_key, url)
-        return url
+        return reverse('catalogue:category', kwargs={
+            'category_slug': self.get_full_slug(parent_slug=parent_slug), 'pk': self.pk
+        })
 
     class Meta:
         abstract = True
--- a/src/oscar/apps/catalogue/abstract_models.py
+++ b/src/oscar/apps/catalogue/abstract_models.py
@@ -184,7 +184,7 @@ class AbstractCategory(MP_Node):
         cache_key = 'CATEGORY_URL_%s_%s' % (current_locale, self.pk)
         return cache_key
 
-    def get_absolute_url(self, parent_slug=None):
+    def _get_absolute_url(self, parent_slug=None):
         """
         Our URL scheme means we have to look up the category's ancestors. As
         that is a bit more expensive, we cache the generated URL. That is
@@ -196,6 +196,9 @@ class AbstractCategory(MP_Node):
         return reverse('catalogue:category', kwargs={
             'category_slug': self.get_full_slug(parent_slug=parent_slug), 'pk': self.pk
         })
+
+    def get_absolute_url(self):
+        return self._get_absolute_url()
 
     class Meta:
         abstract = True
--- a/src/oscar/templatetags/category_tags.py
+++ b/src/oscar/templatetags/category_tags.py
@@ -86,14 +86,15 @@ def get_annotated_list(depth=None, parent=None):
     else:
         categories = Category.get_tree()
 
+    if max_depth is not None:
+        categories = categories.filter(depth__lte=max_depth)
+
     info = CheapCategoryInfo(parent, url="")
 
     for node in categories:
         node_depth = node.get_depth()
         if start_depth is None:
             start_depth = node_depth
-        if max_depth is not None and node_depth > max_depth:
-            continue
 
         # Update previous node's info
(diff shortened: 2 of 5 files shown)


src/oscar/templatetags/category_tags.py
```diff
@@ -3,10 +3,50 @@
...
+    backwards compatibility.
+    """
+
+    def __init__(self, category, **info):
+        super(CheapCategoryInfo, self).__init__(info)
+        self.category = category
+        self.field_names = set(
+            [field.name for field in category._meta.get_fields(include_parents=True)]
+        )
+
```
jaroel: This + the `__getattr__` stuff can be replaced by setting the data up front.

```python
fieldnames = set([field.name for field in category._meta.get_fields(include_parents=True)])
self.__dict__.update(dict([(name, getattr(category, name)) for name in fieldnames)]))
```
Or 
```python
fieldnames = set([field.name for field in category._meta.get_fields(include_parents=True)])
info.update(dict([(name, getattr(category, name)) for name in fieldnames)]))
super(CheapCategoryInfo, self).__init__(info)
```

This allows for faster access by skipping the `__getattr__` call.

src/oscar/templatetags/category_tags.py
```diff
@@ -3,10 +3,50 @@
...
+    def get_absolute_url(self):
+        return self["url"]
+
+    def __getattr__(self, name):
+        "Only allow access to cheap direct attributes of ``Category``"
+        if name in self.field_names:
+            return getattr(self.category, name)
+
```
jaroel: As most attributes won't be used ever, this should be just fine. We could set the attribute on the instance, so the next access will be local. 

```python
def __getattr__(self, name):
    "Only allow access to cheap direct attributes of ``Category``"
    if name in self.field_names:
        value = getattr(self.category, name)
        setattr(self, name, value)
        return value

```

jaroel:
All in all this should improve things so much, that any other optimisations should be of little value.

src/oscar/templatetags/category_tags.py
highbiza: Ok I added a metaclass that sets up these accessors at type creation time.

highbiza:
what, too late omg

src/oscar/templatetags/category_tags.py
specialunderwear: done with a metaclass

src/oscar/templatetags/category_tags.py
```diff
@@ -1,12 +1,33 @@
...
 register = template.Library()
 Category = get_model("catalogue", "category")
 
 
-class CheapCategoryInfo(dict):
+def create_passthrough(name):
+    def category_passthrough(self):
+        return getattr(self.category, name)
+    return category_passthrough
+
```
jaroel: This is the same as the `__getattr__` implementation, but less readable, and we'll still have the multiple function calls.
You'll want something like https://github.com/Pylons/pyramid/blob/master/src/pyramid/decorator.py#L4 , or keep the `__getattr__` implementation as it is probably fast enough.

specialunderwear: I know this information at parse time, so no need to modify __dict__ at runtime.
Also this is definately not the same as the __getattr__.

specialunderwear: For every field on the model, a passthrough methid is added that directs the call to self.category. so these are in ``__dict__`` at parse time and not added at runtime, which is very ugly.

mvantellingen:
This was originally the case, the save method was overridden to denormalize the slug field. See #1828

specialunderwear:
O lol more people are having troubles with this haha

specialunderwear:
I've got some more optimizations that avoid crazy queries on the paginator, and the COUNT statement,  do you want me to add that in this PR or separately?

src/oscar/templatetags/category_tags.py
```diff
@@ -18,36 +79,55 @@ def get_annotated_list(depth=None, parent=None):
...
+            info["num_to_close"] = list(range(0, depth_difference))
+            tree_slugs = tree_slug.rsplit(node._slug_separator, depth_difference)
+            if tree_slugs:
+                tree_slug = tree_slugs[0]
+            else:
+                tree_slug = node.slug
+
+        info = CheapCategoryInfo(
+            node,
+            url=node.get_absolute_url(tree_slug),
```
solarissmoke: This will break backward compatibility for projects that have overridden `get_absolute_url` and are not accepting any arguments. Given the caching optimisations made there I am not sure that extra optimisation to pass a parent slug is worth a breaking change.

src/oscar/templatetags/category_tags.py
```diff
@@ -18,36 +79,55 @@ def get_annotated_list(depth=None, parent=None):
...
-        info = {'num_to_close': [],
-                'level': node_depth - start_depth}
-        annotated_categories.append((node, info,))
         prev_depth = node_depth
 
     if prev_depth is not None:
         # close last leaf
-        info['num_to_close'] = list(range(0, prev_depth - start_depth))
-        info['has_children'] = prev_depth > prev_depth
+        info["num_to_close"] = list(range(0, prev_depth - start_depth))
```
solarissmoke: Please avoid unnecessary syntax changes - this make it hard to use `blame` to track down changes in future.

src/oscar/apps/catalogue/abstract_models.py
```diff
@@ -115,6 +115,19 @@ def full_name(self):
         names = [category.name for category in self.get_ancestors_and_self()]
         return self._full_name_separator.join(names)
 
+    def get_full_slug(self, parent_slug=None):
+        if self.is_root():
+            return self.slug
+
+        cache_key = self.get_url_cache_key()
```
solarissmoke: `get_url_cache_key()` was used to cache the **URL** of the category. It is now being used to cache the slug, which is not the same thing, and will break projects that have overridden any of this behaviour, e.g., to use a different URL structure. I think we should stick to caching the URL as before, and if there is some additional benefit then caching the slug in a separate key.

src/oscar/templatetags/category_tags.py
```diff
@@ -1,12 +1,73 @@
...
+    this class hides any expensive properties people should not use by accident
+    in templates.
+
+    This replaces both the node as the info object returned by the ``category_tree``
+    templatetag, so it mimics a tuple of 2 items (which are the same) for
+    backwards compatibility.
+    """
+
+    def __init__(self, category, **info):
+        super(CheapCategoryInfo, self).__init__(info)
```
solarissmoke: Please use `super()` for consistency across the project. We don't support Python 2.

src/oscar/templatetags/category_tags.py
solarissmoke: Hmm, now that I am a bit more awake I realise that would defeat the whole purpose of this PR. The fact that it is a breaking change however makes me feel that this should be a new template tag, with the old one being deprecated.

specialunderwear: no I can fix this, no worries, I had it defined as a separate method before.

src/oscar/apps/catalogue/abstract_models.py
specialunderwear: Don't you think it is incredibly _weird_ to have a different method of operation for get_absolute_url and reverse? I think caching the full url is a mistake hereby corrected. Caching the full_slug, amounts to the same speed, but leaves operation of reverse the same as operation of get_absolute_url. Please note that this is also the way it was before. full_slug used to be a property of the product which was removed by normalization. get_url_cache_key merely calculates the **unique key** under which some value is stored in the cache, why would this have any effect on custom url structure?

specialunderwear:
One note: This pull request contains backwards incompatible changes by design. Any customizations of the navigation partials that would destroy performance by calling MPTT methods will break because of these changes. It can be made backwards compatible with just nine characters on line 66 in category_tags.py, but the breakage is by design! People must be discouraged to use these methods because they are sloooooooow, especially in loops they become dangerous.

So maybe you want to put this in minor version update.

src/oscar/apps/catalogue/abstract_models.py
```diff
@@ -154,6 +166,9 @@ def get_ancestors_and_self(self):
         if you don't want to include the category itself. It's a separate
         function as it's commonly used in templates.
         """
+        if self.is_root():
+            return [self]
```
specialunderwear: This guard removed 255 completely useless queries in a site we made, have a look at this:

image

solarissmoke: Looks like something that really needs to be fixed upstream in Treebeard - we shouldn't have to do this. Will add an issue there.

PR #2897, opened 2018-12-05
(UPDATE) normalise_email func to handle None and double @ email values

dfirst:
Update for the case discussed  here: https://github.com/django-oscar/django-oscar/pull/2879#discussion_r237121977

### DIFF
--- a/tests/integration/customer/test_utils.py
+++ b/tests/integration/customer/test_utils.py
@@ -1,7 +1,7 @@
 from django.test import TestCase
 
 from oscar.core.compat import get_user_model
-from oscar.apps.customer.utils import get_password_reset_url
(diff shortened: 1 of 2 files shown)


tests/integration/customer/test_utils.py
```diff
@@ -13,3 +13,7 @@ def test_dispatcher_password_reset_url(self):
         user = User.objects.create_user('testuser', 'text@example.com', 'dummypassword')
         url = get_password_reset_url(user)
         self.assertTrue(url.startswith('/password-reset/confirm/'))
+
+    def test_normalise_email(self):
+        self.assertEqual(normalise_email(None), '')
```
solarissmoke: Why should we accept a `None` value here? I don't think it is unreasonable for the function to expect a string as opposed to any other type.

PR #2907, opened 2019-01-03
Properly populate user instance passed to password validators

craigloftus:
The current user registration form does not correctly populate the user object that is passed to the password validators, consequently any validators relying on properties of the user will not be applied correctly.

The overriding of the undocumented `_post_clean` method follows the approach used upstream in  [django.contrib.auth.forms.UserCreationForm](https://github.com/django/django/blob/master/django/contrib/auth/forms.py#L106)


craigloftus:
@solarissmoke Your comments should now be addressed.

PR #2913, opened 2019-01-14
Use iframe to shield page from leaking css in dashboard email html preview

jaroel:


jaroel:
Fixes #2914

jaroel:
This makes it look like this:
screenshot 2019-01-14 at 12 34 38

Issue #2914, opened 2019-01-14
Dashboard email preview injects html directly into page


jaroel:
### Issue Summary

Dashboard email preview injects html directly into page.

### Steps to Reproduce

Go to the dashboard at http://localhost:8000/shop/dashboard/comms/
Go to an email, ie http://localhost:8000/shop/dashboard/comms/PASSWORD_CHANGED/
Make sure something is in "Email body HTML template".
Click preview and see a mess.

![screenshot 2019-01-14 at 12 29 25](https://user-images.githubusercontent.com/70700/51110203-0f727b80-17f8-11e9-8d6f-3c567bdfd2f5.jpg)

solarissmoke:
Thanks for raising this - it's been something I've wanted to address for some time. As regards your proposed PRs:

- #2913 would resolve the CSS issue, but introduces new problems with iframes, specifically that they will break for any site that is enforcing content security policies that restrict these (which frankly any e-commerce site should be doing).

- Not sure what benefit #2915 provides. Only the most technically savvy end user is going to know that emails are typically sent in plain text and HTML variants. For the rest it is just confusing to see two  versions of the same message for no apparent reason.

I actually think we need to go in the other direction and remove these email views from core. I don't see a compelling use case for them (how many other sites exist that let you browse copies of emails they sent you? - I can think of any). Storing emails is fine in case they need to be resent etc. I would propose that letting users browse them is unnecessary. They have an email client for that.

Keen to hear thoughts from others (@craigloftus , @sasha0, @mvantellingen) before we make a decision on this.

sasha0:
> I actually think we need to go in the other direction and remove these email views from core. I don't see a compelling use case for them (how many other sites exist that let you browse copies of emails they sent you? - I can think of any). Storing emails is fine in case they need to be resent etc. I would propose that letting users browse them is unnecessary. They have an email client for that.

I agree that we'd remove them in the future. Although not sure about doing that in the next release.

solarissmoke:
Fixed in #2913

PR #2915, opened 2019-01-14
Show email sent to customer as text and and html.

jaroel:
More or less the same as #2913 / #2914


jaroel:
Like this:
screenshot 2019-01-14 at 13 40 29

solarissmoke:
Thanks for this. Happy with the iframe change, but I don't think it makes sense to include the plain text version of the email as the average end user is not going to understand why they are seeing plain text and HTML alternatives. Would you be able to amend to only include the iframe?

jaroel:
That should do it :)

PR #2918, opened 2019-01-21
Fix locale email translation for NL.

Temoeri:


solarissmoke:
Closing as this will need to be updated on transifex.

PR #2925, opened 2019-01-29
Make app-forking management command more stable

jwayodi:


src/oscar/core/customisation.py
```diff
@@ -73,11 +73,14 @@ def fork_app(label, local_folder_path, local_app_subpackage=None, logger=None):
...
-    # Check local_folder_path is current folder
+    # Remove trailing slash from folder path
+    local_folder_path = local_folder_path.rstrip('/')
+
+    # Check if local_folder_path is current folder
     if local_folder_path == '.':
         local_folder_path = ''
 
-    local_apps_package = local_folder_path.replace('/', '.')
+    local_apps_package = local_folder_path.lstrip('/').replace('/', '.')
```
sasha0: Why do we need to remove leading slash here?

sasha0: Has been able to re-create the case this line aims to fix, which proves necessity of test coverage.

sasha0:
Closing in favor of PR #2995, which is rebased version of this one.

PR #2935, opened 2019-02-11
is_public field for Products

jaroel:
This allows to disable a product, without having to delete it.
A staff member can still view the product on the website via the dashboard so styling issues or offers can be tested.

This introduces `browsable_dashboard` so the user can see disabled products in the dashboard.


solarissmoke:
The discussion on https://github.com/django-oscar/django-oscar/issues/536 is relevant here.

jaroel:
I understand. I haven't ran a project where I didn't need this so far, so I'm running into migration shizzle though.
This would allow us to stay with upstream.

specialunderwear:
https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/offer/abstract_models.py#L784 has the same flag ...

I've just read the discussion  on #536 and - while extremely useful and powerful - making changes to the oscar models usually leads to problems with updates. From this perspective (maintenance and updates), it would be great to have a basic feature set useful for a webshop as is. Instead of having a very basic framework that MUST be extended before it can be useful.

sasha0:
@specialunderwear, so you agree with `is_public` field introduction, especially since `Offer` model already has the same field and does not comply with sort of guideline in #536?

specialunderwear:
I'm sorry I was not clear, indeed I think it is an important feature with many use cases.

okfish:
And what about backward compatibility? It's easy to speak about new projects you are just begun, but what about legacy ones? Why `is_public` and not `active` or `enable` or `hide`? Why boolean? 
Mike was right, do not hang too many meat on the bones. Oscar still has 'framework' in the title...

specialunderwear:
I think it is very interesting that you mention backwards compatibility. The desirable result of having backwards compatible releases, is for one being able to update without requiring changes to existing code. But in my experience, there are no oscar upgrades without rework on customizations.

Keeping the core lean and mean directly contributes to backwards incompatibility, because the feature has to implemented in a customization.

The semantics of the flag are directly defined in the pull request by updating the manager. And a reason for the name is also provided. What is the problem?

okfish:
I'm keeping my projects up to date with oscar releases since 0.7 and can't remember when I had have to rework my code significally excluding some deprecations.

src/oscar/apps/catalogue/views.py
```diff
@@ -41,6 +41,10 @@ def get(self, request, **kwargs):
         if redirect is not None:
             return redirect
 
+        # Do allow staff members so they can test layout etc.
+        if not product.is_public and not request.user.is_staff:
+            raise Http404(_("Product is not public"))
```
pjstevns: Wouldn't a generic 404 be better? No need to expose the fact that a product exists but isn't public.

okfish: By and large, it is a subject of business logic of each project, not the Oscar framework

dfirst:
I think its project specific feature, for example in my project i got status field with related properties including is_public, the problem is very wide and is_public field can bring restrictions not opportunities for developing.

specialunderwear:
Please if making claims, present examples and explanations, these claims are not immediately obvious to everyone!

specialunderwear:
> Case in my project: i got status field with related properties including is_public, the problem for me is_public field can bring restrictions not opportunities in developing.
> Short example.
> `class Product(AbstractProduct):
> 
> ```
>     STATUSES = (
>         ('public', 'Public'),
>         ('pending', 'Pending'),
>         ('under_review', 'Under review')
>     )
> 
>     status = models.CharField(choices=STATUSES, default='public')
> 
>     @property
>     def is_public(self):
>         return self.status == 'public'
> ```

Thanks for explaining yourself. Would you agree that nomatter what status your product has, it is still either public or hidden on the frontend? I mention this, because you have deifned a property exactly for this use case.

I'm going to exaggerate a bit here, so bear with me, suppose I would tell you that I had a project that looks like this:

```
class Product:
    STATUSES = (
        ('is a car', "Bike"),
        ('could become public on friday the thirteeth', "Michael meyers"),
        ('Definately will become public very soon', "Mueller report"),
        ('is_public', "Is public"),        
    )
    @property
     def is_public(self):
         return self.status == 'public'
```

Would you find this a convincing argument for not including a mechanism for hiding a product?

src/oscar/apps/catalogue/views.py
jaroel: A simple 404 would do just fine for me

dfirst:
I think your comment good addition to my case and it shows why this problem is so project specific and any solution in oscar will bring restrictions in development.

specialunderwear:
O thanks for agreeing this is an important feature that should be merged asap.

solarissmoke:
@specialunderwear I think you will find that @dfirst doesn't agree with you, and still feels this is not a good idea. There is clearly no consensus on this yet and so I think there is scope for some more debate.

I am not convinced either way yet, but a few things to add to the discussion:

1. The logic implemented here for allowing staff users to view non-public products is actually making some assumptions that will not hold for all projects - e.g., that this view is not cached, and that projects will always want staff users to be able to view such products (I can think of projects where permissions are more specific than this).

2. The idea of a `browseable_dashboard()` queryset works for now, but will not work if we switch to using the search backend for searching in the dashboard as well. This is something that is planned elsewhere.

3. You ask for examples and so I would like to ask the same from you - can you illustrate an example where "making changes to the oscar models usually leads to problems with updates"? The whole purpose of Oscar is to try and make it easy to do just this, and generally speaking once you have forked an app, it is not in my experience that difficult to keep it up to date with upstream model changes.

specialunderwear:
3. Example for "making changes to the oscar models usually leads to problems with updates":

Let's say I changed the offer model, so I can assign them to a country. I've overridden the offer app but not the voucher app. When an upstream change to the voucher app introduces a migration that references the offer app, you get a very strange error::

```
django.db.migrations.exceptions.NodeNotFoundError: Migration voucher.0004_auto_20180228_0940 dependencies reference nonexistent parent node ('offer', '0007_conditionaloffer_exclusive')
```

What about data migrations? These can be missed, because nothing ever indicates there was one, django will never warn that you  might miss something in the migrations.

 Migrations can not trivially be added to an existing project without inspecting what the migration is about. If I need to maintain only 10 projects, it is allready work to check the changes between the versions, inspect the migrations, inspect the customizations and then decide how to add the missing migrations. When dealing with 100 sites it is allready a HUGE amount of work to make the decision for each site.

2. ``browseable_dashboard()`` can function just fine using a search engine, just index is_public instead of skipping any non public documents.
I can also think of many ways how I can implement product ``status`` next to ``is_public`` without any problems.

1. This could be just a matter of taste. A default, that can be changed if needed, is that unwelcome?

specialunderwear:
My opinion:

After using oscar for some time now, I see the feature incompleteness as a serious defect. I think we need a more complete webshop framework that can be easily modified and adapted, using the superb mechanisms oscar offers. The "setup phase" is getting tired, let's make it useful out of the box but malleable when needed ...

sasha0:
I mostly agree with the opinion above - I believe it'd be the most efficient to evolve core features.

dfirst:
Want to explain why i started to argue:
1. I agree option to show/hide product is very important.
2. And agree current solution is very useful and fit for most cases.
3. But have doubts about this feature, main - it looks like a step to strict "out of box" solution since oscar called itself as "e-commerce framework".
4. And also Im not convinced what is better to add this feature in projects where i need it or to cut/change where i dont.

pjstevns:
Having also worked with oscar for quite some time now, I completely agree with @specialunderwear and would like to offer some additional considerations:

- development pace over the last 5/6 years has been very slow. When was the last major version release again?
- if the framework doesn't evolve, oscar is going to die.
- if outside contribution is not accepted liberally, oscar is going to die. Contributions should be encouraged, and more core contributors should be accepted.
- backward compatibility is fine within major versions, but should not be a consideration between major versions: there is no reason to *not* massively improve the core of oscar.

Of course all this is outside the scope of this ticket. I'll subscribe to the list...

src/oscar/apps/catalogue/managers.py
```diff
@@ -18,7 +18,16 @@ def base_queryset(self):
 
     def browsable(self):
         """
-        Excludes non-canonical products.
+        Excludes non-canonical products and disabled products
```
solarissmoke: For consistency of documentation please change this to "and non-public products". Disabled does not mean the same as "not public".

src/oscar/apps/catalogue/managers.py
```diff
@@ -18,7 +18,16 @@ def base_queryset(self):
...
+        """
+        return self.filter(parent=None, is_public=True)
+
+    def browsable_dashboard(self):
+        """
+        Products that should be browsable in the dashboard.
+
+        Excludes non-canonical products, but allows diabled products to edit
```
solarissmoke: For consistency of documentation please change this to "but includes non-public products". Disabled does not mean the same as "not public".

src/oscar/apps/catalogue/views.py
```diff
@@ -41,6 +41,10 @@ def get(self, request, **kwargs):
         if redirect is not None:
             return redirect
 
+        # Do allow staff members so they can test layout etc.
+        if not product.is_public and not request.user.is_staff:
```
solarissmoke: To make this logic easier to override I think we should use a separate method, e.g., `self.is_viewable()`.

jaroel: feedback applied

src/oscar/apps/catalogue/managers.py
jaroel: feedback applied

src/oscar/apps/catalogue/managers.py
jaroel: feedback applied

src/oscar/apps/catalogue/views.py
jaroel: feedback applied

okfish:
Apprently, it was inevitable if it happened :) 
So, I reviewed my opinion as I remembered the possible reasons why the changes in the models were treated so scrupulously in the earlier versions of Oscar. Abstract model's fields overriding became possible only since Django 1.10. Until this we had need to think twice and think again everytime we had going to shuffle fields. However, it's a good rule for each commit with migrations and for now.

PR #2942, opened 2019-02-19
update README.rst, add django-oscar-saferpay

yvess:
I added django-oscar-saferpay (payment integration) to the README.rst, section community extensions

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -222,6 +222,8 @@ The following are community-written extensions:
   that makes it possible to use `Wells Fargo Retail Services`_ as an order
   payment method.
 
+* django-oscar-saferpay_ - Integration with the saferpay payment gateway
+
 Let us know if you're writing a new one!
 
 .. _django-oscar-unicredit: https://bitbucket.org/marsim/django-oscar-unicredit/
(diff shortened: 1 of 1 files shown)


sasha0:
Not sure if the package is ready to use already - it provides transaction list and detail views for the dashboard for saferpay payments and provides checkout integration in the readme. Please see https://github.com/django-oscar/django-oscar-docdata for the reference.

yvess:
The heavy lifting part is in [django-saferpay](https://github.com/taywa/django-saferpay) django-oscar-saferpay is only the dashboard part and the oscar specific documentation. I separated it so that django-saferpay can also be used for other projects without the oscar dependency. It's already used in a production. I made the readme link to django-saferpay more visible.

sasha0:
> The heavy lifting part is in django-saferpay django-oscar-saferpay is only the dashboard part and the oscar specific documentation.

That's the thing - `django-oscar-saferpay` could have Saferpay integration into the checkout and etc, described by you in the Readme, directly in the codebase, so that user won't have to copy-paste it.

sasha0:
Due to the lacking feedback from the author and considering my comments above are still relevant (package does not provide any integration in the codebase, but suggests it in the README), I'm inclined to close this PR, but would like to know your opinion on that @solarissmoke

yvess:
No problem, I will close the PR. Perhaps I will create some day a "proper" package, but not sure if/and when I will do it. I had the perception that you almost always need to customise it, so I focused on documenting that part.

PR #2948, opened 2019-02-25
Use {% load sorl_thumbnail %} to avoid conflicts with easy-thumbnails.

sir-sigurd:


sasha0:
Thanks for the contribution.

PR #2951, opened 2019-02-27
Upgrade Bootstrap to 3.4.1.

sasha0:


solarissmoke:
@sasha0 you need to recompile assets - these are committed.

sasha0:
@solarissmoke indeed, thanks.

solarissmoke:
@sasha0 I think this needs to be backported to 1.6.x seeing as it is a security release?

Also I'd suggest that as a matter of best practise we get at least one person to review PRs before they are merged? Does that sound OK to you?

sasha0:
> @sasha0 I think this needs to be backported to 1.6.x seeing as it is a security release?

Yep. I'll backport soon, so you cat cut a release.

> Also I'd suggest that as a matter of best practise we get at least one person to review PRs before they are merged? Does that sound OK to you?

Agree for the most cases, but we can take this discussion into the Slack.

SalahAdDin:
Do you have planned to upgrade to Bootstrap 4?

PR #2965, opened 2019-03-21
Add a node.js installation command within a dockerfile

iz4blue:
Correct errors that occur during the creation of a docker image.


sasha0:
Thanks for the contribution!

PR #2970, opened 2019-03-24
enforce max_affected_items during offer application

### LINKED ISSUE #2788: Offers are consumed twice on basket lines
I am wondering about how to apply offer for consume offer for django oscar 1.6.4,

I was digging in to the code base, and found out that we call `consume` 2 times when applying a benefit. For example
* In `apps.basket.abstract_models.py`
```
 def discount(self, discount_value, affected_quantity, incl_tax=True,
                 offer=None):
        """
        Apply a discount to this line
        """
        if incl_tax:
            if self._discount_excl_tax > 0:
                raise RuntimeError(
                    "Attempting to discount the tax-inclusive price of a line "
                    "when tax-exclusive discounts are already applied")
            self._discount_incl_tax += discount_value
        else:
            if self._discount_incl_tax > 0:
                raise RuntimeError(
                    "Attempting to discount the tax-exclusive price of a line "
                    "when tax-inclusive discounts are already applied")
            self._discount_excl_tax += discount_value
        self.consume(affected_quantity, offer=offer)
```

And then again in `apps.offer.conditions.py` in each condition, they have ` def consume_items(self, offer, basket, affected_lines):` and call `line.consume(quantity_to_consume, offer=offer)` method again.

I don't think it's supposed to call `consume` twice like this?

### Technical details
* Oscar version:  django oscar 1.6.4

pjstevns:
also restrict the usage of offer consumption to line.consume()

fixes #2788

### DIFF
--- a/src/oscar/apps/offer/abstract_models.py
+++ b/src/oscar/apps/offer/abstract_models.py
@@ -661,8 +661,6 @@ class AbstractBenefit(BaseOfferMixin, models.Model):
             if not price:
                 # Avoid zero price products
                 continue
-            if line.quantity_without_offer_discount(offer) == 0:
-                continue
             line_tuples.append((price, line))
 
         # We sort lines to be cheapest first to ensure consistent applications
--- a/src/oscar/apps/basket/utils.py
+++ b/src/oscar/apps/basket/utils.py
@@ -6,6 +6,7 @@ from django.template.loader import render_to_string
 from oscar.core.loading import get_class
 
 Applicator = get_class('offer.applicator', 'Applicator')
+ConditionalOffer = get_class('offer.models', 'ConditionalOffer')
 
 
 class BasketMessageGenerator(object):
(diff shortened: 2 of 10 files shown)


sasha0:
Looks good in general, although I'm not sure about introduction of type annotations. What do you think @solarissmoke ?

solarissmoke:
I think type annotations are a good thing and we should encourage their use - why are you not keen on them @sasha0 ?

sasha0:
I was thinking about this from consistency perspective. Although, if you think PR looks good, we can merge it and introduce type annotations throughout the project later.

solarissmoke:
I'd like to do a bit of testing before we merge - just to verify that the bug reference here is fixed. Should have some time for this later in the week.

solarissmoke:
This works well for me. The documentation for how all of this works could do with improvement, but I think that is a separate issue.

sasha0:
Let's give it a go then.

src/oscar/apps/offer/abstract_models.py
```diff
@@ -661,8 +661,6 @@ def get_applicable_lines(self, offer, basket, range=None):
             if not price:
                 # Avoid zero price products
                 continue
-            if line.quantity_without_offer_discount(offer) == 0:
-                continue
```
maziar-dandc, line 665: curious.

Issue #2772, opened 2018-07-20
"Product analytics" report on dashboard ignores date range


st8st8:
Found a bug? Please fill out the sections below.

### Issue Summary
"Product analytics" report on dashboard ignores date range

### Steps to Reproduce
Visit the dashboard, click on the "Reports" tab.
Select "Product analytics" from the "Report type" dropdown.
Select a date range and click "generate report"
The results returned will always be for all time, despite the date range selected

### Technical details
Any Python / Django version, Oscar 1.6.2

The ProductReportGenerator ignores the date range, and the stored analytics do not keep track of it. It might be nice to disable the date range controls for those reports that will never store date information.  Though a date breakdown of products purchased would be a nice to have...

SrdjanCosicPrica:
I have two solutions if I may.

1. The fields are disabled completely.

2. Add a new model <Insert Name Here> that contains the values on ProductRecord and has a Foreign Key to ProductRecord. This model is created/updated for each day. (If no views/additions/purchases are made then the objects would not be created) The model would have a date for that day for filtering purposes.

What do you think? @sasha0

alxs:
Just a note that "User analytics" behave in the same way

v1kku:
Date filtering is definitely broken for CSV exports. The [function that performs the filtering](https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/dashboard/reports/views.py#L59) is [never called](https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/dashboard/reports/views.py#L48) for CSV downloads. It also wouldn't work because the CSV formatter's `generate()` returns a response.

PR #2974, opened 2019-03-25
Add date range fields for product and user report generators in order to support filter by date range.

sasha0:


sasha0:
Fixes #2772

solarissmoke:
@sasha0 I don't think this is the behaviour that is requested in #2772. The request there is to be able to filter the data on *purchases* made between the selected dates, not on products created between those dates. I think that makes sense and it much more useful. Not sure that filtering on creation date is very useful.

sasha0:
Yeah. Then, let's postpone it for now.

PR #2975, opened 2019-03-25
Implement standalone 500 error template.

### LINKED ISSUE #2971: Default 500 error template should not extend layout.html
The current 500 error template in Oscar extends `error.html`, which in turn extends `500.html`. This is bad practise because it means that if you have an error in the parent template, then the 500 template will fail to render.

The 500 template should be standalone and not extend any other template.

sasha0:
Fixes #2971

### DIFF
--- a/src/oscar/templates/oscar/500.html
+++ b/src/oscar/templates/oscar/500.html
@@ -1,15 +1,20 @@
-{% extends "error.html" %}
-{% load i18n %}
+<!DOCTYPE html>
+<html>
+    <head>
+        <title>Server error!</title>
 
-{% block title %}
-    {% trans 'Server error!' %} | {{ block.super }}
-{% endblock %}
+        <meta http-equiv="content-type" content="text/html; charset=UTF-8" />
+        <meta name="robots" content="NOARCHIVE,NOCACHE" />
+    </head>
 
-{% block error_heading %}
-    {% trans "Server error!" %}
-{% endblock %}
+    <body>
+        <h1>Server error!</h1>
+        <p>We're sorry but something went terribly wrong and we've been unable to generate the page you requested.</p>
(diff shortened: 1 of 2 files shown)


src/oscar/templates/oscar/500.html
```diff
@@ -1,15 +1,34 @@
-{% extends "error.html" %}
 {% load i18n %}
+{% load staticfiles %}
+<!DOCTYPE html>
+<html lang="{{ LANGUAGE_CODE|default:"en-gb" }}" class="{% block html_class %}no-js{% endblock %}">
+    <head>
+        <title>{% block title %}{{ shop_name }} - {{ shop_tagline }}{% endblock %}</title>
```
solarissmoke: I think this should be hard-coded to "Server error" as it's possible that context processors will not have run depending on what caused the 500 error. 

My view is that we should not use any template tags or template context in this template - it should be pure HTML. See https://github.com/wagtail/wagtail/blob/master/wagtail/project_template/project_name/templates/500.html for example.

Projects can always override this to provide their own styling if they want to.

sasha0: You mean we'd not use even styling at all?

solarissmoke: Yeah. The rationale for this is to minimise the chance of causing additional errors in the template itself. Django [tries to do the same thing](https://docs.djangoproject.com/en/2.1/ref/views/#django.views.defaults.server_error):

> The default 500 view passes no variables to the `500.html` template and is rendered with an empty `Context` to lessen the chance of additional errors.

In our projects we provide inline CSS for this view, but don't rely on any static assets.

sasha0: Ok agreed.

sasha0:
Sure. Added.

PR #2976, opened 2019-03-25
Remove form from the basket line formset and bypass its validation if related model instance does not exist any more.

### LINKED ISSUE #2873: RelatedObjectDoesNotExist at /en-gb/basket/ Line has no basket.
### Issue Summary
`RelatedObjectDoesNotExist at /en-gb/basket/ Line has no basket.`
Occurs when updating a basket across multiple tabs. 

### Steps to Reproduce

The issue comes down to having the basket open in two tabs. To reproduce:

1. Add items to the basket and open the basket page in two tabs.
2. Remove an item from the basket in one of the tabs.
3. Try to update that same item in the other tab.

### Technical details

* Python version: 3.7.1
* Django version: 2.1
* Oscar version: 2.0dev0

sasha0:
Fixes #2873

### DIFF
--- a/src/oscar/apps/basket/formsets.py
+++ b/src/oscar/apps/basket/formsets.py
@@ -1,4 +1,5 @@
 from django.forms.models import BaseModelFormSet, modelformset_factory
+from django.utils.functional import cached_property
 
 from oscar.core.loading import get_classes, get_model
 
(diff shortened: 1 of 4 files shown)


src/oscar/apps/basket/formsets.py
```diff
@@ -23,6 +23,10 @@ def _should_delete_form(self, form):
         """
         if super()._should_delete_form(form):
             return True
+
+        # If related form instance already removed, let's skip this form.
```
solarissmoke: This comment is a little misleading - we're not skipping the form, but deleting it.

sasha0: Fixed.

sasha0:
> I think it's worth adding tests for this - to ensure that a formset containing such a form still validates properly.

Tests added.

Issue #2982, opened 2019-03-28
Make Benefit apply methods accept kwargs


fakhar362:
If one decides to use the `max_total_discount` or the `discount_amount` field available on `apply()` methods of `PercentageDiscountBenefit` and `AbsoluteDiscountBenefit`, you run into the issue that the other benefits of `FixedPriceBenefit` etc don't accept these **arguments** or **kwargs** in general, and because of this, in your overridden `apply_benefit()` method on `ConditionalOffer` you have to add in a conditional to check for the proxyclass of your Benefit to only add the additonal kwarg i.e. `max_total_discount` in case of `PercentageDiscountBenefit` or `AbsoluteDiscountBenefit`, so it would be beneficial if all apply methods to start accepting kwargs so one can go without adding in the additional checks for **proxy class** of `Benefit`

Came across this issue in our project where we decided to restrict the max discount per application of an offer, other people will also probably run into similar issues if they want to make use of the `max_total_discount` argument.

sasha0:
Done in PR #2983.

PR #2983, opened 2019-03-28
Modified benefits apply methods to accept kwargs to avoid adding conditional in ConditionalOffer's apply_benefit method

fakhar362:
Refers to https://github.com/django-oscar/django-oscar/issues/2982


pjstevns:
looks good to me.

Issue #2699, opened 2018-05-11
Easier "Colors" option?


khink:
On March 2nd, i asked this on Slack:

> Kees Hink [2:56 PM]
> How would one model a catalogue where all Products of a certain ProductClass are available in all AttributeOptions, and at the same price? Does one have to add all variants manually?
> 
> To take the sandbox as an example: All “T-shirt” products are now available in all Colors, and they all cost the same. So when i add a new T-shirt product, i don’t want to have to add Variants for each individual Color: there might be 20, that’s a lot of work. 
> (For now let’s disregard the “T-shirt size” factor by saying the new product is a size L. But bonus points if there’s a way to also let the customer pick any size as well, without having to add these manually.)
> 
> Is there a shortcut here?
> 
> Kees Hink [3:24 PM]
> I think product.options does what i want. Small issues:
> - I don’t seem to be able to set that through the dashboard (need django-admin)
> - It displays a string input field instead of the desired “colors” dropdown
> But that’s just implementation details, the basics are there!

Now that i've implemented this (https://www.barriquewijnhandel.nl/catalogue/category/glitterfles_7/), i thought i'd report what i had to do to get it working:

- Fork Oscar catalogue
- Add Color model + migration
- Fork Oscar's basket app.
- Override Oscar's SimpleAddToBasketForm
- Modify the AddToBasketForm to show a choice widget instead of text
- Test that we can manage custom colors for adding a Color product to basket
- Use Color objects in add to basket form
- Customize dashboard catalogue nav: also manage colors
- Make Colors manageable in Oscar Dashboard

Is this something we want to make easier? If so, i could share the knowledge, or make a PR or something.

solarissmoke:
@khink thanks for sharing this! A couple of thoughts:

I personally think that `Color` specifically is probably too project-specific for Oscar to implement in core. You are, however, describing a fairly common use case ("choice-based options") and I think it's worth exploring whether Oscar can make it easier to implement that.

That said, the bulk of the work you describe above was in adding a `Color` model and making it dashboard-editable (which for me is quite project-specific).

We have previously used a different approach and used a product attribute to store a list of colors for each product, and then used these to populate a choice widget for the options, just as you have described. The benefit for us with that was that no additional dashboard functionality had to be written to allow management of these choices (they are just attributes) - all we had to do we swap out the widget used to render the options field. 

Would this approach have worked for you?

specialunderwear:
[...]

Issue #1527, opened 2014-10-20
Customizable Products with Catalogue.Options


jlmadurga:
This functionality is focused in adding  not only CharField  _Catalogue.Options_ to Products. This is a key functionality for shops selling customizable Products.
In order to fulfill this requirement just follow the schema and the idea of _ProductAttributes_ which only defines the type of the field and then other instance _ProductAttributeValue_ handles the value. Forms generates fields dynamically (_dashboard/catalogue/forms:ProductForm_)

My proposal:
- catalogue:
  - abstract_models: _AbstractOption_ modify type attr to chose the type of field add required attr to handle actual required/not_required choice field 
- basket
  - forms: _addToBasket_ modify _add_option_field to generate form field dynamically (already done in _dashboard/catalogue/forms:ProductForm_)
  - abstract_models: _AbstractLineAttribute_ modify value attr not to just handle CharField. Follow _ProductAttributeValue_ logic keep an eye to line_reference generation (slug) 
- order
  - abstract_models: _AbstractLineAttribute_ remove type attr set value attr as it is done in _Basket.AttributeLineAttribute_

The same scheme model is repeated in _Basket.LineAttribute_, _Order.LineAttribute_ and _Catalogue.ProductAttributeValue_. We should consider DRY:
- forms/fields.py: Add _FIELD_FACTORIES_ defined in _dashboard/catalogue/forms_
- models/???:  Add _models.Model_ abstract class or mixin to define dynamically value as it is done already in _Catalogue.ProductAttributeValue_

Waiting for your feedback.

Remiz:
Bumping this issue for interest...

I saw that your commits haven't been integrated to the project, is there is any update on that? I like your approach, I too feel that charfield is too limited for product options...

jlmadurga:
Hi Remiz,

I am using those changes in production for oscar 0.7. I know those changes where not integrated. The solution is not as clean as it should be but it works.
If i get time I don't mind to refactor not having same logic solution in three places.

PR #2984, opened 2019-03-28
(FEATURE) new catalogue.Option.types

dfirst:
Hello i want to share draft that currently i'm working on.

This is enchantment for catalogue.Option feature. I want to add more control over options fields.

Currently i added simple option types.


sasha0:
First of all, could you please refer to the related issue or explain goal of this feature.

dfirst:
> First of all, could you please refer to the related issue or explain goal of this feature.

What is it and the goal: This is enchantment for catalogue.Option feature. I want to add more control over options fields
Related issues:.
https://github.com/django-oscar/django-oscar/issues/2699
https://github.com/django-oscar/django-oscar/issues/1527
Also in next phase im planing to implement:
https://github.com/django-oscar/django-oscar/issues/1527

dfirst:
@solarissmoke Please can you looks on this PR when you have time, its ready for review.

IAmNatch:
@dfirst I read over this and it looks really clean! Looking forward to it getting merged in!

Pooch11:
This looks great! Hoping this PR can land soon.

src/oscar/apps/basket/forms.py
```diff
@@ -9,10 +9,38 @@
 
 Line = get_model('basket', 'line')
 Basket = get_model('basket', 'basket')
+Option = get_model('catalogue', 'option')
 Product = get_model('catalogue', 'product')
 
 
+def _option_text_field(option):
```
solarissmoke, line 16: Can I suggest a slightly different API here, something like this, in a separate module:

```
class OptionFieldRenderer:
    field_class = forms.CharField

    def __init__(self, product, option):
        self.product = product
        self.option = option

    def get_field_class(self):
        return self.field_class
    
    def get_field_kwargs(self):
        return {
	    'label': self.option.name,
            'required': self.option.required
        }

    def get_form_field(self):
        field_class = self.get_field_class()
        return field_class(self.get_field_kwargs())


class IntegerOptionFieldRenderer(OptionFieldRenderer):
    field_class = forms.IntegerField


class BooleanOptionFieldRenderer(OptionFieldRenderer):
    field_class = forms.BooleanField
```

I think this would be more flexible because (a) it allows projects to override these classes in the regular way (b) passing the product to this class means you can alter the form fields based on product (or more likely product class).

src/oscar/apps/basket/forms.py
```diff
@@ -182,8 +219,8 @@ def _add_option_field(self, product, option):
         This is designed to be overridden so that specific widgets can be used
         for certain types of options.
         """
-        self.fields[option.code] = forms.CharField(
-            label=option.name, required=option.is_required)
+        option_field = self.OPTION_FIELD_FACTORIES[option.type](option)
```
solarissmoke, line 222: I think we need to default to the charfield if the option type is not found in the map - this can happen if a project overrides the list of types.

src/oscar/apps/catalogue/abstract_models.py
```diff
@@ -1103,17 +1103,27 @@ class AbstractOption(models.Model):
...
+        (FLOAT, _("Float")),
+        (DATE, _("Date")),
     )
-    type = models.CharField(_("Status"), max_length=128, default=REQUIRED,
-                            choices=TYPE_CHOICES)
+
+    name = models.CharField(_("Name"), max_length=128)
+    code = AutoSlugField(_("Code"), max_length=128, unique=True,
+                         populate_from='name')
+    type = models.CharField(_("Type"), max_length=255, default=TEXT, choices=TYPE_CHOICES)
```
solarissmoke, line 1125: Since we are changing this field anyway, let's avoid using reserved words. Prefer `option_type` instead of `type`.

src/oscar/apps/catalogue/abstract_models.py
```diff
@@ -1126,7 +1136,10 @@ def __str__(self):
 
     @property
     def is_required(self):
-        return self.type == self.REQUIRED
+        """
```
solarissmoke, line 1139: Suggest we drop the docstring (most people won't see it, and it's not clear about when/how this will be deprecated) - better we log a [deprecation warning](https://github.com/django-oscar/django-oscar/blob/4644ede05a54e287a2445107b8ef000a70a842e4/src/oscar/core/decorators.py) in here.

src/oscar/templates/oscar/partials/form_field.html
```diff
@@ -22,12 +22,12 @@
             {% endblock %}
 
             {% block controls %}
-                <div class="{% if style|default:"stacked" != 'stacked' %}col-sm-7{% endif %}">
+                <div class="{% if style|default:"stacked" != 'stacked' %}col-sm-7{% endif %}{% if field.widget_type == 'CheckboxInput' %} checkbox{% endif %}">
```
solarissmoke, line 25: We are applying the `checkbox` class to both the column and to the label, which I don't think is a good idea. I think this should be controlled at the field level exclusively.

src/oscar/apps/catalogue/abstract_models.py
```diff
@@ -1103,17 +1103,27 @@ class AbstractOption(models.Model):
...
+        (DATE, _("Date")),
     )
-    type = models.CharField(_("Status"), max_length=128, default=REQUIRED,
-                            choices=TYPE_CHOICES)
+
+    name = models.CharField(_("Name"), max_length=128)
+    code = AutoSlugField(_("Code"), max_length=128, unique=True,
+                         populate_from='name')
+    type = models.CharField(_("Type"), max_length=255, default=TEXT, choices=TYPE_CHOICES)
+    required = models.BooleanField(_("Is option required?"), default=False)
```
solarissmoke, line 1126: Please change label to "Is this option required?"

src/oscar/apps/catalogue/abstract_models.py
specialunderwear: Tried it and a bit too much work at this point.

PR #2991, opened 2019-04-04
Improve code documentation for dashboard's product list view

maikhoepfel:
I spent too much time trying to understand why the UPC queries were
structured the way they are. Now that I understand it, I'm leaving some
notes for the next person.

I also made a small tweak to the search: I'd argue UPCs are not case sensitive in real life, so we should treat them the same way in Oscar.

### DIFF
--- a/src/oscar/apps/dashboard/catalogue/views.py
+++ b/src/oscar/apps/dashboard/catalogue/views.py
@@ -131,8 +131,10 @@ class ProductListView(SingleTableView):
 
     def apply_search(self, queryset):
         """
-        Filter the queryset and set the description according to the search
-        parameters given
+        Search through the filtered queryset.
+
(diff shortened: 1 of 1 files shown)


sasha0:
Looks good, but has lint violations.

I think we'd add test for this change as well. Yours thoughts @solarissmoke ?

maikhoepfel:
Thanks for the quick review. I'll look at the lint violations later. I'm afraid I don't have time to increase test coverage - I mostly wanted to leave the docstrings.

maikhoepfel:
No more lint errors. :+1:

sasha0:
> No more lint errors. 👍

Thanks!

> I'm afraid I don't have time to increase test coverage - I mostly wanted to leave the docstrings.

Ok np, I could add test by myself.

Issue #2250, opened 2017-02-23
Unclear dependency between dashboard and frontend templates


pwalsh:
The so called ["cheeky setting" here](https://github.com/django-oscar/django-oscar/blob/master/src/oscar/__init__.py#L23) enforces an explicit dependency between dashboard and frontend templates in Oscar sites. 

The value is dubious, ("(t)his allows Oscar's templates to be extended by templates with the same filename"), and the impact ~ hidden from users who customise their frontend templates (some important markup not in blocks, etc.).

As a user, I'm not sure what the real problem is:

1. The cheeky setting itself (what is the value of this overloading, and what problem does it solve that Django's template resolution order does not solve?).
2. The fact that at the bottom of the template stack, there is a common dependency between the dashboard and the frontend, and not all the markup can be excluded via blocks.

Is there a design pattern I am missing here?

pwalsh:
@sasha0 @solarissmoke and anyone:

I propose:

1. remove or discourage use of this setting (i.e.: Oscar templates should be available namespaces by `oscar/`, and not pollute `/`).
2. If the above is not possible/desirable, make a break where the Oscar frontend templates do not depend on the same base template as the Oscar dashboard templates. Example: [this](https://github.com/django-oscar/django-oscar/blob/master/src/oscar/templates/oscar/dashboard/base.html#L1) should extend `oscar/base` and not `base`.

solarissmoke:
I can't see what value that setting provides. It might have been because you cannot fully control the order of apps in `INSTALLED_APPS` because of Oscar's `get_core_apps` stuff - but that is tenuous too.

I would like to see both your suggestions implemented - i.e., completely separate dashboard and frontend base templates, and also namespace all of Oscar's templates and remove this setting. This will likely be a breaking change.

solarissmoke:
Looking a bit more closely, the setting is simply a device to let us store all the templates in one directory (ass opposed to per-app directories). I don't think it does any more than that (in terms of enforcing any dependency between frontend/dashboard templates). Either way I think your suggested changes are a good idea.

sasha0:
Interrelated with #1378

solarissmoke:
The "cheeky setting" has been removed in #3000, which I think should address the issue that was raised here.

PR #3000, opened 2019-04-09
Remove `OSCAR_MAIN_TEMPLATE_DIR` setting.

sasha0:
Refs #1378 #2250


sasha0:
@auvipy thanks.

Would appreciate your feedback @django-oscar/developers @django-oscar/core @specialunderwear.

vdboor:
When this change is applied, how does one override the template?
Previously, one could add let's say a `catalogue/product_list.html` template which did `{% extends "oscar/catalogue/product_list.html" %}` and override the blocks. This functionality seems to be lost.

Should one use django-apptemplates `{% extends "oscar:oscar/catalogue/product_list.html" %}`, or does modern Django automatically continue the lookup for a base template in another app?

sasha0:
User will be able to customize template via `{% extends "oscar/catalogue/product_list.html" %}` tag.

specialunderwear:
> does modern Django automatically continue the lookup for a base template in another app?
@vdboor yes it does, so that is awesome right?

sasha0:
@specialunderwear how does the PR look for you?

docs/source/internals/getting_started.rst
```diff
@@ -59,14 +59,11 @@ context processors.
 
 .. code-block:: django
 
-    from oscar import OSCAR_MAIN_TEMPLATE_DIR
-
     TEMPLATES = [
         {
             'BACKEND': 'django.template.backends.django.DjangoTemplates',
             'DIRS': [
                 os.path.join(BASE_DIR, 'templates'),
```
specialunderwear: wouldn't it be nicer to just add "oscar" to INSTALLED_APPS and have no template configuration at all?

auvipy: I guess this is what is going to be achieved by this change

specialunderwear: No I mean the whole ``TEMPLATES =`` thing can be removed from the documentation because it is not needed.

specialunderwear:
> @specialunderwear how does the PR look for you?

Super awesome!! I just wanted to explain what I mentioned in #1378 because a backwards compatible solution is actually possible. If you look at: https://github.com/django/django/blob/master/django/views/generic/base.py#L140

You can see that ``TemplateResponseMixin`` actually returns a list of templates. If you would Define a common mixin to all oscar views like this:

```
class OscarBackwardCompatibleTemplateResponseMixin(TemplateResponseMixin):
    old_template_name = None
    def get_template_names(self):
        templates_names = super(). get_template_names()
        if self. old_template_name is not None:
            return [self. old_template_name] + templates_names
        return templates_names
```

That would allow you to define the ``old_template_name`` on a class level if needed and oscar would look for the old template first, the new one after that.

specialunderwear:
> That would allow you to define the `old_template_name` on a class level if needed and oscar would look for the old template first, the new one after that.

Please note that I am mentioning it, but have never tried it. An approach like I mention could add complexity and trouble, so maybe it is better to not have a backwards compatible solution!
I'm just mentioning it in case you where considering backward compatibility.

docs/source/internals/getting_started.rst
auvipy: oh yeah, it's not needed anymore!! so should be dropped as well

solarissmoke:
I don't think we should try to maintain backward compatibility on this - the original design here was problematic, and we should discourage it.

auvipy:
> I don't think we should try to maintain backward compatibility on this - the original design here was problematic, and we should discourage it.

Strongly agree.

sasha0:
> I don't think we should try to maintain backward compatibility on this - the original design here was problematic, and we should discourage it.

What's your general opinion on PR?

solarissmoke:
Looking good - just need to remove the `TEMPLATES` instructions from the docs/

sasha0:
> Looking good - just need to remove the TEMPLATES instructions from the docs.

Firstly, how about `context_processors` configuration in the `TEMPLATES` setting?

Secondly, how would it work if templates located in the project not inside the application directory? I've re-checked for the sandbox and apparently it did not work without the setting.

solarissmoke:
Sorry, I should have been more specific. I don't think we need to show a full `TEMPLATES` setting in the documentation, because `TEMPLATES['DIRS']` is no longer required to hold `OSCAR_MAIN_TEMPLATES_DIR`. 

We could simplify the instructions just to say what needs to be added to `TEMPLATES['OPTIONS']['context_processors']`?

sasha0:
> We could simplify the instructions just to say what needs to be added to TEMPLATES['OPTIONS']['context_processors']?

Yep, rewrote it a little bit. What do you think?

specialunderwear:
Now we're talking

sasha0:
Thanks everyone for the review, feedback and input: @solarissmoke @specialunderwear @auvipy

auvipy:
thanks for your great work @sasha0

jaroel:
Very nice, thank you all!

PR #3001, opened 2019-04-10
WIP: Improve support of multi-tenancy sites.

sasha0:


specialunderwear:
Since you are not loading the mixin with get_class, what would this improve?

sasha0:
> Since you are not loading the mixin with get_class, what would this improve?

Did not have in mind the case when the user would want to customize the whole mixin class. Please share your thoughts on that anyway @specialunderwear .

specialunderwear:
Well moving code to a mixin by itself is not an improvement by itself right? It just adds another level of indirection. So what exactly would this improve in the context of multi tenancy?

webbuilderhero:
does Oscar have Multi Tenancy support yet?

Issue #2730, opened 2018-06-01
Please include support for Bootstrap 4


ctailo:
### Issue Summary

Please include support for Bootstrap 4. Right now, Oscar appears to support Bootstrap 3. Please include support for Bootstrap 4

SalahAdDin:
All PR are welcome, if you begin it i will help you.

ctailo:
Thank you. I am very new to programming but I am willing to help. Can you please tell me what I would need to do to include Bootstrap 4?

SalahAdDin:
Please see in documentation.

ashishnitinpatil:
@ctailo It will be quite a big change due to the amount of templates and static files that would need updating (all the while taking care of design responsiveness on all screen sizes with the changes). But, in summary, you will have to update the relevant static files (less, js, css, images, fonts) for bootstrap with latest stable release from v4, then update all the templates that use bootstrap classes from v3, to v4 classes (or the corresponding changes if that class is deprecated or something entirely new would be better to use).

SalahAdDin:
@ashishnitinpatil Maybe, `tabler`?

gogobook:
How about `Bulma` ?

lordvcs:
Please lets stick with bootstrap 4

ctailo:
> Please lets stick with bootstrap 4

I agree with you. I think that bootstrap 4 is the easiest one to work with. I do not know of other frameworks. I am fairly new to programming, so I am not sure that I can commit to doing this by myself. However, if other people want to work on this, I can submit pull requests if I am able to contribute in some way.

Aiky30:
This is going to be a big undertaking. I will help towards this but I'm unsure just how much of my spare time I can contribute. I guess the place to start is creating a branch that we can all push changes to. 

feature/bootstrap-4.0

ctailo:
> This is going to be a big undertaking. I will help towards this but I'm unsure just how much of my spare time I can contribute. I guess the place to start is creating a branch that we can all push changes to.
> 
> feature/bootstrap-4.0

Ok. How could this project be managed. Should we list out all of the things that should be done to make this change. As I said before, I am new to programming, so I am not familiar with everything that would be needed to do this.

Aiky30:
@ctailo I've not made any changes to this project before so I have no idea of the procedure for new features in this repo. 

[...]

PR #3005, opened 2019-04-12
Bootstrap 4

sasha0:
Refs #2730


samitnuk:
@django-oscar/core

This PR rebased against the `master` branch. Please review and comment. Thanks.

solarissmoke:
Hey @samitnuk , is this ready for review?

samitnuk:
> Hey @samitnuk , is this ready for review?

Yes, it's ready.

solarissmoke:
@samitnuk third party SCSS files should not be committed - see #3236 where we removed these. Please remove everything in `src/oscar/static_src/oscar/scss/bootstrap4/`. These scss modules should ideally be imported directly from `node_modules` in our own scss files, or, as an intermediate step, copied into `static/` by the gulp task. If you could make this change first it will reduce the diff considerably so that its easier to review.

samitnuk:
> @samitnuk third party SCSS files should not be committed - see #3236 where we removed these. Please remove everything in `src/oscar/static_src/oscar/scss/bootstrap4/`. These scss modules should ideally be imported directly from `node_modules` in our own scss files, or, as an intermediate step, copied into `static/` by the gulp task. If you could make this change first it will reduce the diff considerably so that its easier to review.

@solarissmoke Done. Please continue your review when you will be able. Thanks.

src/oscar/core/context_processors.py
```diff
@@ -25,7 +25,7 @@ def metadata(request):
             'homepage_url': settings.OSCAR_HOMEPAGE,
             'language_neutral_url_path': strip_language_code(request),
             # Fallback to old settings name for backwards compatibility
-            'use_less': (getattr(settings, 'OSCAR_USE_LESS', None)
-                         or getattr(settings, 'USE_LESS', False)),
+            'use_scss': (getattr(settings, 'OSCAR_USE_SCSS', None)
+                         or getattr(settings, 'USE_SCSS', False)),
```
solarissmoke: I think we need to remove this variable and setting entirely - you can't load the scss in the browser any more so it isn't used.

src/oscar/static_src/oscar/css/dashboard.css
```diff
@@ -0,0 +1,11286 @@
+/*!
```
solarissmoke: This file should not be committed - it should be compiled into `static/` at build time. Same for everything in `static_src/oscar/fonts`.

src/oscar/static_src/oscar/scss/fontawesome/_mixins.scss
```diff
@@ -0,0 +1,56 @@
+// Mixins
```
solarissmoke: Everything in `scss/fontawesome` shouldn't be committed - font awesome should be a dependency in `package.json` and compiled at build time the same as we do with Bootstrap SCSS.

src/oscar/static_src/oscar/scss/page/layout.scss
```diff
@@ -0,0 +1,33 @@
...
+}
+
+.row {
+  box-sizing: border-box;
+}
+
+// The page container
+.page .page_inner {
+  background: $white;
+  -webkit-border-radius: 0 0 6px 6px;
```
solarissmoke: `box-shadow` and `border-radius` are now widely supported - we shouldn't need the browser prefixes.

src/oscar/static_src/oscar/scss/page/product_lists.scss
```diff
@@ -0,0 +1,83 @@
+// No margin on lists
```
solarissmoke: Think we should be using the new [`no-gutters` class](https://getbootstrap.com/docs/4.0/layout/grid/#no-gutters) on such lists, instead of styling them manually here.

src/oscar/static_src/oscar/scss/page/reviews.scss
```diff
@@ -0,0 +1,65 @@
...
+  h3 {
+    margin-bottom: $base-margin-bottom;
+    overflow: hidden;
+    text-overflow: ellipsis;
+  }
+
+  // Vote forms inline with "Is this review helpful?"
+  form {
+    display: inline-block;
+    *display: inline;
```
solarissmoke: Is this valid css?

src/oscar/static_src/oscar/scss/page/type.scss
```diff
@@ -0,0 +1,97 @@
+// Re setting margins to the bottom only
```
solarissmoke: Don't think this reset is necessary any more - BS4 doesn't have top margins on headings by default.

src/oscar/static_src/oscar/scss/page/type.scss
```diff
@@ -0,0 +1,97 @@
...
+.page-header,
+.sub-header,
+.table {
+  h1, h2, h3, h4, h5, h6, form {
+    margin-bottom: 0;
+  }
+}
+
+// Heading fonts a little smaller
+h1, .h1 {
```
solarissmoke: We should be modifying the sass variables provided by Bootstrap (`$h1-font-size` etc.), rather than adding additional styles here.

src/oscar/templates/oscar/catalogue/browse.html
```diff
@@ -96,7 +96,7 @@ <h4>{% trans "Refine by" %}</h4>
             <div>
                 <ol class="row">
                     {% for product in products %}
-                        <li class="col-xs-6 col-sm-4 col-md-3 col-lg-3">{% render_product product %}</li>
+                        <li class="col-xs-12 col-sm-6 col-md-4 col-lg-3">{% render_product product %}</li>
```
solarissmoke: `col-xs-*` is not a valid bootstrap class - we can drop `col-xs-12` entirely.

solarissmoke: Is there a reason for changing the column widths of these elements?

src/oscar/templates/oscar/layout.html
```diff
@@ -7,11 +7,8 @@
         files in your project, then set OSCAR_USE_LESS = True in your
         settings file. This will enable the on-the-fly less compiler.
     {% endcomment %}
```
solarissmoke: Please remove this comment.

package.json
```diff
@@ -5,26 +5,29 @@
...
     "inputmask": "^4.0.9",
     "jquery": "^3.4.1",
     "jquery-mousewheel": "^3.1.13",
     "jquery-sortable": "^0.9.13",
     "less": "^3.10.3",
+    "popper.js": "^1.15.0",
     "pump": "^3.0.0",
     "select2": "4.0.12",
     "select2-bootstrap-theme": "0.1.0-beta.10",
+    "tempusdominus-bootstrap-4": "^5.1.2",
```
solarissmoke: This library doesn't look maintained - there are several bugs which have not been addressed for months. We've stopped using it on our projects, and switched to using the native browser date input when available. I think this provides a better user experience, especially on mobile, and would suggest that Oscar defaults to this - projects can then add their own custom datepickers if they want to.

samitnuk: Ok. Will remove it.

src/oscar/templates/oscar/catalogue/browse.html
samitnuk: `col-xs-*` removed.

samitnuk: The reason for changing the column widths - the default font size changed from 14px in Bootstrap 3 to 16px in Bootstrap 4. I kept the coefficients for `h1`, `h2` ..., but now all text bigger in general. With updated column width we will not have text overlapping.

package.json
samitnuk: I found that not all modern browsers support native input types:
* `time` - Safari
* `date` - Safari
* `datetime-local` - FF, Safari

Probably `date` and `time` is not a problem, because we can use `inputmask` for these fields (and they will work in Safari as well) 🤔
But `datetime-local` uses a format like this `2017-06-01T08:30` (I suppose we will need to add this format somewhere in settings globally) and `inputmask` for `datetime` is not "user-friendly" without the date picker (because of hours, minutes and seconds).

I agreed that it is better to use as fewer dependencies as possible, but I am not sure how we can completely remove the date picker.

How do you solve this in your projects? Should we use native input types with `inputmask` for each such field? Thanks.

solarissmoke:
Hey @samitnuk is this ready for another review?

samitnuk:
> Hey @samitnuk is this ready for another review?

@solarissmoke Yes, please continue. Thanks.

solarissmoke:
Thanks @samitnuk! I've done one final review on this and found a few minor issues which I fixed, and then rebased, in #3302.

Issue #3008, opened 2019-04-16
Support for Django 2.2


SrdjanCosicPrica:
Maybe I'm missing something but will there be support for Django 2.2 any time soon?
Can't seem to find any related issues to 2.2 specifically or anything in the docs.

PR #3009, opened 2019-04-17
Django 2.2 support.

sasha0:
Refs #3008

### DIFF
--- a/src/oscar/apps/customer/utils.py
+++ b/src/oscar/apps/customer/utils.py
@@ -5,8 +5,8 @@ from django.contrib.auth.tokens import default_token_generator
 from django.core.mail import EmailMessage, EmailMultiAlternatives
 from django.urls import reverse
 from django.utils.encoding import force_bytes
-from django.utils.http import urlsafe_base64_encode
 
+from oscar.core.compat import urlsafe_base64_encode
 from oscar.core.loading import get_model
 
 
@@ -134,7 +134,7 @@ def get_password_reset_url(user, token_generator=default_token_generator):
     """
     kwargs = {
         'token': token_generator.make_token(user),
-        'uidb64': urlsafe_base64_encode(force_bytes(user.id)).decode(),
+        'uidb64': urlsafe_base64_encode(force_bytes(user.id)),
     }
     return reverse('password-reset-confirm', kwargs=kwargs)
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,18 +12,26 @@ matrix:
       env: TOXENV=py35-django111
     - python: 3.6
       env: TOXENV=py36-django111
+    - python: 3.7
+      env: TOXENV=py37-django111
     - python: 3.5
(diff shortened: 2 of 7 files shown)


pjstevns:
:+1:

SrdjanCosicPrica:
Apologies for a newbie question. But doesn't `setup.py` need to be changed as well?
Right now it is 
```
install_requires = [
    'django>=1.11,<2.2',
    ...
```

maerteijn:
@SrdjanCosicPrica No it should be adjusted indeed as well

src/oscar/apps/customer/utils.py
```diff
@@ -132,9 +132,12 @@ def get_password_reset_url(user, token_generator=default_token_generator):
     """
     Generate a password-reset URL for a given user
     """
+    uidb64 = urlsafe_base64_encode(force_bytes(user.id))
+    if isinstance(uidb64, bytes):
```
solarissmoke: We will almost certainly forget to remove this check when we drop support for Django 2.1 and below. I think it is better to wrap `urlsafe_base64_encode` in a helper method in `oscar.core.compat`, so we know it can be removed later without leaving dead code around.

sasha0:
> This is however mentioned in the comment in catious_slugify and used in templates here and here. So maybe we should add |safe in the templates and update the comment?

Do we really need to add safe filter there, since slugify is not supposed to return any HTML characters?

> I also saw that you added xenial to the travis configuration: Then you could also add python 3.7 as well (that works with xenial).

Ok done.

maerteijn:
No I don’t think so, just mentioned it as an option. Then I would suggest to remove the mark_safe part in the comment.

maerteijn:
> > I also saw that you added xenial to the travis configuration: Then you could also add python 3.7 as well (that works with xenial).
> 
> Ok done.

I think you should adjust `tox.ini`:
`py{35,36,37}-django{111,20,21}` -> add `22` as well

And refer to the correct tox envs in `.travis.yml`:
```
    - python: 3.7
      env: TOXENV=py36-django22
``` 
(should be `TOXENV=py37-django22`)

src/oscar/apps/customer/utils.py
sasha0: Ok done.

sasha0:
> No I don’t think so, just mentioned it as an option. Then I would suggest to remove the mark_safe part in the comment.

Right. Updated.

> I think you should adjust tox.ini:
py{35,36,37}-django{111,20,21} -> add 22 as well
>
>And refer to the correct tox envs in .travis.yml:
>
>    - python: 3.7
>      env: TOXENV=py36-django22
>(should be TOXENV=py37-django22)

Ok, adjusted.

maerteijn:
Looks fine to me, so @solarissmoke should review (and hopefully approve) your `urlsafe_base64_encode` implementation and then we can merge right?

sasha0:
Yep.

PR #3019, opened 2019-05-07
Remove several deprecated model fields from order and partner apps.

### LINKED ISSUE #1962: StockRecord: rename price_excl_tax
The docstring for `price_excl_tax` already explains what it truly is: the base price for tax calculations. I've had two projects where we knew the tax-inclusive price, and only calculated the price excl tax. I'd argue that a stock record is merely a record of a base price, and only the pricing policy gives it any meaning in regards to tax, so the field name should reflect that. 
I realize that the field is probably used directly more often than it should (instead of through pricing strategies), so we need a good idea about being friendly as far as backwards-compatibility is concerned.

solarissmoke:
This PR removes the following model fields:

  - `order.Line.est_dispatch_date`
  - `order.Line.unit_cost_price`
  - `order.Line.unit_retail_price`
  - `partner.StockRecord.cost_price`
  - `partner.StockRecord.price_retail`

and renames `partner.StockRecord.price_excl_tax` to `partner.StockRecord.price`

I have not removed `Line.line_price_incl_tax` and `Line.line_price_excl_tax`, which we had previously deprecated. This is because those fields are used by several properties/methods and to untangle those will require a bit of work - I've left that for later.

Refs #1968. Closes #1962.


solarissmoke:
I'll rebase this and regenerate migrations after we've released v2.1 in a few days.

Issue #2791, opened 2018-08-16
Cannot remove any item from basket if 2 or more items are unavailable


netok421:
### Issue Summary

Cannot remove any item in `BasketView` if 2 or more items in basket became unavailable

### Steps to Reproduce

1. Add 2 different items to your basket
2. Add a different item to your basket
3. Set the stock on items added in step 1 to 0, making them unavailable
4. Go to the cart page and attempt to remove any item

### Result

The JSON response has an error `Your basket couldn't be updated. Please correct any validation errors below.` . The basket is not updated.

### Expected behavior

The item you removed should be removed.

### Related

#1654

sasha0:
> The alternative is to automatically remove unavailable items from baskets (and leave a message saying that this happened), but I don't think that addresses the basic problem of editing individual lines.

Alternatively, we could save only valid basket forms if at least one basket is valid. Of course, not the best approach, but I'm not sure what would be better way around this (using an API is quite a different topic). WDYT @django-oscar/core ?

solarissmoke:
Fixed in #3361.

PR #3025, opened 2019-05-11
Bug Fix: remove any item from basket if 2 or more items are unavailable

prashg008:
Bug #2791 Fix

### DIFF
--- a/src/oscar/apps/basket/forms.py
+++ b/src/oscar/apps/basket/forms.py
@@ -61,11 +61,9 @@ class BasketLineForm(forms.ModelForm):
         # add second time, just updating number of items.
         qty_delta = qty - self.instance.quantity
         is_allowed, reason = self.instance.basket.is_quantity_allowed(qty_delta)
-
         if self.instance.basket.strategy.request and self.instance.basket.strategy.request.is_ajax:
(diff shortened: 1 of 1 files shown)


src/oscar/apps/basket/forms.py
```diff
@@ -61,13 +61,19 @@ def check_max_allowed_quantity(self, qty):
         # add second time, just updating number of items.
         qty_delta = qty - self.instance.quantity
         is_allowed, reason = self.instance.basket.is_quantity_allowed(qty_delta)
+        if self.instance.basket.strategy.request and self.instance.basket.strategy.request.is_ajax:
```
solarissmoke: `is_ajax` is a method, not a property. The callable itself will always evaluate to truthy.

src/oscar/apps/basket/forms.py
```diff
@@ -61,13 +61,19 @@ def check_max_allowed_quantity(self, qty):
...
+            if self.instance.basket.strategy.request.POST.get('form-0-DELETE'):
```
solarissmoke: This also assumes that this form is the first item in the formset - there is no guarantee that this is the case.

solarissmoke:
Closing due to lack of feedback - if you're willing/able to work on this again please let us know.

PR #3027, opened 2019-05-13
Add back sorl-thumbnail as a dependency of Oscar.

### LINKED ISSUE #3023: Failing readthedocs build for 2.0b1
### Issue Summary

The documentation build for the 2.0 branch is failing - this is because the documentation references `sorl`, but this is not a core dependency.

We could add documentation build configuration to handle this, but I feel like regular users are going to hit this same issue. `pip install django-oscar` will not install a working build because there is no thumbnailer installed.

I think we need to revert to having `sort` in `install_requires`, as the default thumbnailer.

solarissmoke:
I think we made a mistake by dropping the default thumbnailer as a dependency. It means that `pip install django-oscar` doesn't give you a working setup, which is going to trip up a lot of newcomers. Since the default is still sorl I think we should keep it as a dependency.

Also fixes #3023.

### DIFF
--- a/docs/source/ref/settings.rst
+++ b/docs/source/ref/settings.rst
@@ -330,7 +330,7 @@ The time to live for the basket cookie in seconds.
 
 Default: ``10000``
 
-The maximum number of products that can be added to a basket at once. Set to 
+The maximum number of products that can be added to a basket at once. Set to
 ``None`` to disable the basket threshold limitation.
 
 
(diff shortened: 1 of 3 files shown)


sasha0:
What would be an issue to install dependency manually?

Secondly, how would be make then sorl optional and allow user only install easy_thumbnails?

solarissmoke:
> What would be an issue to install dependency manually?

The issue for me is that `pip install django-oscar`, which is what most people will do, doesn't work. They will hit an error, and then have to go work out that they need to install sorl. I don't think this is a good experience for the user.

> Secondly, how would be make then sorl optional and allow user only install easy_thumbnails?

Given that sorl is Oscar's default thumbnailer, a project that wants to use something else would be able to uninstall sorl. I think this is preferable to having a setup that doesn't work out of the box.

sasha0:
I'm still not convinced this is the right approach.

We use same solution as django-storages, which raises exception if dependency is not installed - https://github.com/django-oscar/django-oscar/blob/master/src/oscar/core/thumbnails.py#L16-L18 and thus make user acknoledged that this is missing.

solarissmoke:
@sasha0 I don't think the django-storages approach of throwing an exception at runtime is all that friendly. 

I would suggest one of the following two options for consideration, and we go with whichever is most popular:

1. As with the current PR, just make sorl a dependency of Oscar. The vast majority of users will use this anyway, so it eliminates one extra installation hurdle for them (especially people new to Oscar who just want to get it running - I worry that we are just making it harder for such people).
2. Leave sorl as an extra, but add a system check that verifies that `OSCAR_THUMBNAILER` is importable, and that and throw an error if it isn't. This is preferable to raising an exception in the class itself, because system checks happen before the server starts - i.e., we are informing the user as early as possible.

My preference is for (1), but happy to go with (2) if there is more support for that.

solarissmoke:
Closing due to conflicting opinions on whether to do this.

PR #3028, opened 2019-05-13
Update translatable strings

solarissmoke:
This incorporates all the string changes that have been made in the code base in the last year. I've also tweaked the makemessages command not to wrap lines, as that results in a smaller diff.


Makefile
```diff
@@ -92,7 +92,7 @@ test_migrations: install-migrations-testing-requirements ## Tests migrations
 # Translations Handling
 #######################
 extract_translations: ## Extract strings and create source .po files
-	cd src/oscar; django-admin.py makemessages -a
+	cd src/oscar; django-admin.py makemessages -a --no-wrap
```
sasha0: I'd make this separate commit, since the change has nothing to do with the commit message "Update translatable strings.".

solarissmoke: OK, done.

sasha0: Can't see updates yet.

PR #3034, opened 2019-05-16
Add less.js to Dashboard Layout

### LINKED ISSUE #3033: OSCAR_USE_LESS doesn't work on dashboard
### Issue Summary
``OSCAR_USE_LESS`` suppose to enable less on browser for fast development. It works fine with storefront / site, but doesn't work with Dashboard.

### Steps to Reproduce
Just set ``OSCAR_USE_LESS = True`` and go to dashboard. Its broken there.

### Cause of issue
``oscar/dashboard/layout.html`` doesn't include ``less.js`` as it does in ``oscar/layout.html``

### Solution
Following snippet might be included in ``oscar/dashboard/layout.html``

```
{% block cdn_scripts %}
  {{ block.super }}
  {% if use_less %}
    {# Load the on-the-fly less compiler. Never do this in production. #}
    <script src="//cdnjs.cloudflare.com/ajax/libs/less.js/2.5.3/less.min.js"></script>
  {% endif %}
{% endblock %}
```

shimulch:
OSCAR_USE_LESS = True doesn't work in Oscar Dashboard. This commit makes it work by putting less.js on dashboard/layout.html. It resolves #3033


src/oscar/templates/oscar/dashboard/layout.html
```diff
@@ -121,6 +121,14 @@ <h1>{% block headertext %}{% endblock %}</h1>
 
 {% endblock %}
 
+{% block cdn_scripts %}
+  {{ block.super }}
+  {% if use_less and debug %}
+    {# Load the on-the-fly less compiler. Never do this in production. #}
+    <script src="//cdnjs.cloudflare.com/ajax/libs/less.js/2.5.3/less.min.js"></script>
```
solarissmoke: The latest version on the CDN is 3.9.0 - is there a reason you haven't used that?

src/oscar/templates/oscar/dashboard/layout.html
```diff
@@ -121,6 +121,14 @@ <h1>{% block headertext %}{% endblock %}</h1>
 
 {% endblock %}
 
+{% block cdn_scripts %}
+  {{ block.super }}
+  {% if use_less and debug %}
```
solarissmoke: Please drop the `debug` check - we have provided enough warning about this, but should allow the user to test this without all the other debug components.

src/oscar/templates/oscar/dashboard/layout.html
shimulch: Well I have just used the version from ``src/oscar/templates/oscar/layout.html``. I thought that might be more consistent.

src/oscar/templates/oscar/dashboard/layout.html
shimulch: I see. FYI debug is used on oscar/templates/oscar/layout.html too. Should I remove debug from there too? in a single commit?

src/oscar/templates/oscar/dashboard/layout.html
solarissmoke: Could you please update both to use the latest version?

src/oscar/templates/oscar/dashboard/layout.html
solarissmoke: Oh, I had missed that, sorry. Let's leave as-is for now I think.

PR #3037, opened 2019-05-18
Retrieving product options now performs only 1 query.

specialunderwear:
Also it returns all the product options combined. This makes more sense because
as the existing tests prove, there is allready a way to retrieve the product_class
options and the product options separately.


solarissmoke:
Looks good - please could you fix the lint errors?

solarissmoke:
Sorry to be a pain, but can you remove the changes you have committed to package-lock.json? This is not a meaningful change and just seems to happen when different versions of npm are used - we need to look separately into a way of preventing that.

PR #3043, opened 2019-05-30
Upgrade CDN version of jQuery

rik:
https://github.com/django-oscar/django-oscar/pull/3018 updated the in-tree versions


solarissmoke:
Thanks!

PR #3046, opened 2019-06-05
Fix: warning messages of UppercaseCharField and NullCharField

mongkyo:
I fixed deprecation warning.
change Nullcharfield and UppeercaseCharField class in oscar/models/fields.

### DIFF
--- a/src/oscar/models/fields/__init__.py
+++ b/src/oscar/models/fields/__init__.py
@@ -78,7 +78,6 @@ class UppercaseCharField(CharField):
             cls, name, **kwargs)
         setattr(cls, self.name, Creator(self))
 
-    # I fix it!
     # Django 2.0 updates the signature of from_db_value.
     # https://docs.djangoproject.com/en/2.0/releases/2.0/#context-argument-of-field-from-db-value-and-expression-convert-value
     if django.VERSION < (2,):
@@ -100,9 +99,6 @@ class UppercaseCharField(CharField):
                 value = self.to_python(value)
             return value
 
-    # Previous code
-    # def from_db_value(self, value, expression, connection, context):
-    #     return self.to_python(value)
 
     def to_python(self, value):
         val = super().to_python(value)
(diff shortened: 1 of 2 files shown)


src/oscar/models/fields/__init__.py
```diff
@@ -73,14 +74,31 @@ class UppercaseCharField(CharField):
...
         super().contribute_to_class(
             cls, name, **kwargs)
         setattr(cls, self.name, Creator(self))
 
-    def from_db_value(self, value, expression, connection, context):
-        return self.to_python(value)
+    # Django 2.0 updates the signature of from_db_value.
+    # https://docs.djangoproject.com/en/2.0/releases/2.0/#context-argument-of-field-from-db-value-and-expression-convert-value
+    if django.VERSION < (2,):
+        def from_db_value(self, value, expression, connection, context):
```
solarissmoke: I don't think we need to have different definitions based on Django version. We can simplify our code to:

```
def from_db_value(self, value, *args, **kwargs):
```
Since we don't use any of the other args. This will work for all versions of Django.

src/oscar/models/fields/__init__.py
```diff
@@ -73,14 +74,31 @@ class UppercaseCharField(CharField):
...
+            """
+            Converts a value as returned by the database to a Python object. It is
```
solarissmoke: Please remove this comment - looks like it's copied from some other library.

src/oscar/models/fields/__init__.py
```diff
@@ -73,14 +74,31 @@ class UppercaseCharField(CharField):
     Defined with the with_metaclass helper so that to_python is called
     https://docs.djangoproject.com/en/1.6/howto/custom-model-fields/#the-subfieldbase-metaclass  # NOQA
     """
-
```
solarissmoke: Please avoid unnecessary white space changes, thanks!

mongkyo: thank you your comment. I fix right now

mongkyo:
I modified all of thing that your comment. Thank you @solarissmoke

mongkyo:
> I modified all of thing that your comment. Thank you @solarissmoke

PR #3047, opened 2019-06-05
Fix: warning messages of UppercaseCharField and NullCharField

mongkyo:
I fixed deprecation warning.
change Nullcharfield and UppeercaseCharField class in oscar/models/fields.


solarissmoke:
@mongkyo could you clarify why you're opening and closing multiple PRs? If you need to change your code you can just update the branch you've made your PR on.

PR #3050, opened 2019-06-06
Add is_public to Category

jaroel:
Allows for disabling categories without deleting them.
This is so shop owners can build up new categories without them being available for the public.


solarissmoke:
This looks good to me - want to hold off merging for a few days until we ship v2.0.

jaroel:
Not a problem, thanks!

src/oscar/apps/catalogue/abstract_models.py
```diff
@@ -77,6 +77,16 @@ def has_attributes(self):
         return self.attributes.exists()
 
 
+class CategoryQuerySet(MP_NodeQuerySet):
+    def public(self):
```
solarissmoke: Can we change this to `browsable()` to be consistent with `ProductQuerySet`, and add a docstring to this method?

jaroel: Thanks for the feedback. I've renamed `public` to `browsable`.

specialunderwear:
#3160 is a problem with this code

src/oscar/apps/catalogue/abstract_models.py
```diff
@@ -77,6 +77,19 @@ def has_attributes(self):
         return self.attributes.exists()
 
 
+class CategoryQuerySet(MP_NodeQuerySet):
+    def browsable(self):
+        """
+        Excludes non-public categories
+        """
+        return self.filter(is_public=True)
```
specialunderwear: This is not sufficient, all subcategories of non-public categories must be filtered out as well. If not bad thing happen, see #3160

specialunderwear: fixed in 3de975c

specialunderwear:
This is now good to merge, just needs approval from @solarissmoke

specialunderwear:
This PR needs work because it is too slow.

specialunderwear:
@jaroel can you update this PR with your performance fixes on postgres?

specialunderwear:
These changes do indeed solve the performance problem that was only apparent on postgresql.

src/oscar/apps/catalogue/receivers.py
```diff
@@ -1,16 +1,19 @@
...
+
+from oscar.core.loading import get_model
+
+Category = get_model("catalogue", "Category")
 
 if settings.OSCAR_DELETE_IMAGE_FILES:
     from django.db import models
-    from django.db.models.signals import post_delete
 
     from oscar.core.loading import get_model
```
solarissmoke: Duplicate import.

src/oscar/apps/catalogue/abstract_models.py
```diff
@@ -34,6 +36,41 @@
...
+     Would be about the same as the python statement::
+
+        henk.startswith("koe")
+
+    ReverseStartsWith will flip the right and left hand side of the expression,
+    effectively making this the same query as::
+
+    "koe".startswith(henk)
+
+    This is used by the range query below, where we need to flip select children
```
solarissmoke: This comment is no longer valid because the class has moved. Let's remove it.

jaroel:
Agree - I'll get it sorted asap

jaroel:
Awesome, thanks!

solarissmoke:
I've updated and rebased this PR. Note that I've changed how the ``Category.objects`` manager is defined to be consistent with what we do for `Product` - we don't need a separate manager class. I've also moved the queryset into `managers.py` so that it can be overridden.

src/oscar/apps/catalogue/abstract_models.py
```diff
@@ -158,9 +207,34 @@ def save(self, *args, **kwargs):
...
         super().save(*args, **kwargs)
 
+    def set_ancestors_are_public(self):
+        # Update ancestors_are_public for the sub tree.
+        # note: This doesn't trigger a new save for each instance, rather
+        # just a SQL update.
+        included_in_non_public_subtree = self.__class__.objects.filter(
+            is_public=False, path__rstartswith=OuterRef("path"), depth__lt=OuterRef("depth")
+        )
+        self.get_descendants_and_self().update(
```
iamsobanjaved, line 219: @jaroel is it possible to change this query which is mysql compatible as well.

> MySQL doesn't allow updating the table you are already using in an inner select as the update criteria.

jaroel: Dunno, I don't use mysql. What did you already try but didn't work?

PR #3062, opened 2019-06-17
Restore voucher set URLs

rik:
As part of https://github.com/django-oscar/django-oscar/commit/fbf8bac4c5061657879f3f3baa1c1e7078cffb60, the Voucher set URLs disappeared.

The following exception is logged on every Dashboard page:
`Reverse for 'voucher-set-list' not found. 'voucher-set-list' is not a valid view function or pattern name.`

### DIFF
--- a/src/oscar/apps/dashboard/vouchers/apps.py
+++ b/src/oscar/apps/dashboard/vouchers/apps.py
@@ -41,5 +41,16 @@ class VouchersDashboardConfig(OscarDashboardConfig):
                 name='voucher-delete'),
             url(r'^stats/(?P<pk>\d+)/$', self.stats_view.as_view(),
                 name='voucher-stats'),
+
+            url(r'^sets$', self.set_list_view.as_view(),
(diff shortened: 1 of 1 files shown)


rik:
Out of curiosity, why are views collected in `ready` to then be hooked up in `get_urls`? Could it be done in one step in `get_urls`?

solarissmoke:
Thanks for this!

> Why are views collected in `ready` to then be hooked up in `get_urls`

This is so that it is easy for projects to override individual views without having to re-declare the entire set of URLs for an app.

SunilRao01:
Hello all, I believe there is a missing character which, without it, will result in a 404 when trying to access `http://localhost:8000/en-gb/dashboard/vouchers/sets/`.

Here is a link to the line in the src: https://github.com/django-oscar/django-oscar/blob/660dbea2d9a91d429b15b20e4c7415970be1c64b/src/oscar/apps/dashboard/vouchers/apps.py#L45

Please correct me if I'm wrong.

PR #3071, opened 2019-07-04
Reduce amount of queries performed by Range to 1.

specialunderwear:


src/oscar/apps/catalogue/abstract_models.py
```diff
@@ -580,7 +580,7 @@ def get_missing_image(self):
         return MissingProductImage()
 
     def get_all_images(self):
-        if self.is_child and not self.images.exists():
+        if self.is_child and not self.images.exists() and self.parent is not None:
```
vdboor: `and not self.parent_id` avoids another query.

src/oscar/apps/offer/abstract_models.py
```diff
@@ -1025,7 +996,7 @@ def is_reorderable(self):
         """
         Test whether products for the range can be re-ordered.
         """
-        return len(self._class_ids()) == 0 and len(self._included_categories()) == 0
+        return not(self.included_categories.exists() or self.classes.exists())
```
vdboor: add a space, `not (...)`

src/oscar/apps/offer/abstract_models.py
```diff
@@ -991,27 +931,58 @@ def num_products(self):
...
+        return self.product_queryset
+
+    @cached_property
+    def product_queryset(self):
         """
         Return a queryset containing all the products in the range
 
         This includes included_products plus the products contained in the
         included classes and categories, minus the products in
         excluded_products.
```
vdboor: is there a reason this docstring isn't kept in the `all_products()` method?

tests/integration/offer/test_range.py
```diff
@@ -209,11 +209,11 @@ def test_range_is_reordable(self):
             product=product, category=included_category)
         self.range.included_categories.add(included_category)
 
-        self.range.invalidate_cached_ids()
+        self.range.invalidate_cached_queryset()
```
vdboor: I suppose `invalidate_cached_ids()` should remain as alias for `invalidate_cached_queryset()` to ensure backwards compatibility ?

specialunderwear: nah, this goes in 2.0 so we can cleanup and break stuff.

specialunderwear:
> Wonderful work getting this into a single query!
> How much did this improve the performance?

Well depends on what people do with the ranges. But on certain badly constructed sites it got rid of 5000 queries en 12 seconds of request time hahahaha. The range object becomes a lot more useful like this, the ``contains_product`` queryset method is also handy if you want to get to the applicable discounts really fast (can also be done in 1 query).

specialunderwear:
There is 1 thing that can be a regression in performance, methods that return real queries instead of casted to list or set can slow down paginators ans things that access by index, because querysets will perform a query when accessed by index. I haven't checked for this, because I don't think that is in scope of this PR, but worth noting. I think all the model methods should return queryset and never casted to list. That should be done when needed at the point where that is needed.

tests/integration/offer/test_range.py
solarissmoke: Just FYI I'm in the middle of cutting a stable release for 2.0 (and we have already released a beta for it), so this change is going to have to go into 2.1. Will try and test/review in the next few days!

specialunderwear:
How about it? This is a really nice PR all our sites have become much snappier after this improvement. Pages that list products become a lot snappier because of the performance improvement on the strategy.

solarissmoke:
Will try and get this merged next week.

Issue #2916, opened 2019-01-15
Notification body is rendered as plain text, but passed HTML in some cases


solarissmoke:
### Issue Summary

The body for customer notifications are [rendered as plain text](https://github.com/django-oscar/django-oscar/blob/f2bb0fb73c83cb879fb2dec9184059dd96d814e0/src/oscar/templates/oscar/customer/notifications/detail.html#L29) in the detail view.

But we create notifications with HTML in the body - e.g., [for stock alerts](https://github.com/django-oscar/django-oscar/blob/f2bb0fb73c83cb879fb2dec9184059dd96d814e0/src/oscar/apps/customer/alerts/utils.py#L101).

The resulting notifications are rendered with escaped HTML such as `<a href="/catalogue/test-product_6/">test product</a> is back in stock`.

We need to either allow HTML in the body, or ensure we don't provide it. We already allow HTML in the subject so it seems silly not to allow it in the body.

TravellerSam:
I can confirm this bug is still present in 3.0. Also, some strings are not translated (or translatable) in the same template.

<img width="1078" alt="Screenshot 2021-02-04 at 12 59 00" src="https://user-images.githubusercontent.com/872730/106890303-6900b800-66e9-11eb-8ce6-540eec6908e2.png">

specialunderwear:
<img width="836" alt="image" src="https://github.com/django-oscar/django-oscar/assets/52858/a4259185-8100-4514-bf4d-72f98802d73a">
this has been fixed

PR #3076, opened 2019-07-09
Basket is valid even if items are unavailable, so basket can still be edited and items removed.

SmileyJames:
Solution to issue #2791.

This allows items to be removed from a basket without validation errors impeding.

The checkout independently checks that basket is in stock.

I believe deleting this check/validation has no negative side effects, but the maintainers should think about if it does.

Two tests have been changed.

Third commit is a solution to issue #2916.


solarissmoke:
Thanks. I think you're right that removing this check isn't a problem - the user will still get redirected back here if they attempt to proceed with ineligible items in the basket - but what it does mean is that such errors are not salient to the user until they try to do so. Going to give this a bit of thought to see if there are any other implications before we merge.

solarissmoke:
We tested this more carefully, and found that this approach is problematic: removing the validation entirely means that you can add invalid quantities to to the basket from this view - e.g., to change the quantity to be greater than the number available, or to bypass per-user restrictions on quantities - and these will not report any error until you attempt to proceed to the next step.

@jwayodi is working on an alternative approach which ensures that the validation is skipped only for the specific case where items are no longer available, and being removed from the basket.

PR #3077, opened 2019-07-12
Substituted 'catalogue:index' in favour of OSCAR_HOMEPAGE

Andruten:
- Substituted 'catalogue:index' in favour of OSCAR_HOMEPAGE.
- Tests adapted.


src/oscar/apps/catalogue/views.py
```diff
@@ -131,7 +132,7 @@ def get(self, request, *args, **kwargs):
         except InvalidPage:
             # Redirect to page one.
             messages.error(request, _('The given page number was invalid.'))
-            return redirect('catalogue:index')
```
solarissmoke: I don't think we should change this particular URL within the catalogue app. The catalogue app has it's own index URL defined, and we should respect that - it's possible that the home page of a project is different from the catalogue root.

src/oscar/templates/oscar/partials/nav_primary.html
```diff
@@ -42,7 +42,7 @@
                         </a>
                         <ul class="dropdown-menu" data-navigation="dropdown-menu">
                             {% category_tree depth=2 as tree_categories %}
-                            <li><a tabindex="-1" href="{% url 'catalogue:index' %}">{% trans "All products" %}</a></li>
+                            <li><a tabindex="-1" href="{{ OSCAR_HOMEPAGE }}">{% trans "All products" %}</a></li>
```
solarissmoke: This also needs to stay as `'catalogue:index'` as we don't know whether the home page is the catalogue root.

src/oscar/apps/catalogue/views.py
Andruten: Ok! thank you @solarissmoke, i'll update the PR.

Andruten: done!

src/oscar/templates/oscar/partials/nav_primary.html
Andruten: done!

PR #3081, opened 2019-07-20
Add support for spell checking and carry out initial corrections

craigloftus:
As a first step to starting to work on improving the documentation, I thought it made sense to getting the spelling sorted out and consistent. I went with British English as that was predominantly used in the project already.

In terms of whitelisting, I've gone with a split approach, using a wordlist for common words that are not found by enchant spellchecker, and file specific whitelists for names e.g., NGINX.

Potentially controversially, when fixing spelling issues with classes and method names etc, I've used sphinx roles, so that sphinx will cross reference the documentation... I know this isn't to everyone taste, but it makes the documentation a lot more usable.


docs/source/conf.py
```diff
@@ -263,7 +281,7 @@ def process_docstring(app, what, name, obj, options, lines):
     # Only look at objects that inherit from Django's base model class
     if inspect.isclass(obj) and issubclass(obj, models.Model):
 
-        # Ignore abstract models
+        # Ignore abstract models
```
specialunderwear: what happened here?

craigloftus: Some how that was a non-breaking space; flake8 picked it up.

Issue #3080, opened 2019-07-20
`default_access_fn` always returns True for forked apps


rik:
### Issue Summary
I've managed to reproduce this for any dashboard app and will use the Pages dashboard app as the example in this ticket.

I've narrowed the issue down to `apps.get_containing_app_config` returning [Shop](https://github.com/django-oscar/django-oscar/blob/fe983bf7140457d06094d843da0b94ebaea4d6c8/src/oscar/config.py#L13) when it should return `PagesDashboardConfig`.
https://github.com/django-oscar/django-oscar/blob/fe983bf7140457d06094d843da0b94ebaea4d6c8/src/oscar/apps/dashboard/nav.py#L98

I've tried changing the check on line 99 to 
```py
if app_config_instance is None or not isinstance(app_config_instance, OscarDashboardConfig):
```

But then, `get_app_config_class` will return the base `OscarDashboardConfig` class which is also not what we want.

### Steps to Reproduce
1. Fork the Pages dashboard app
`./manage.py oscar_fork_app pages_dashboard . oscar_forks.dashboard.pages`
1. Log into the dashboard with a user with `partner.dashboard_access`
1. Notice the Content submenu is visible while the whole Pages app has this default permission https://github.com/django-oscar/django-oscar/blob/fe983bf7140457d06094d843da0b94ebaea4d6c8/src/oscar/apps/dashboard/pages/apps.py#L13

### Technical details
* Python version: 3.7.4
* Django version: 2.2.3
* Oscar version: 2.0

rik:
I've solved it on my project. I'm not sure if this approach could be generic enough to be used as a basis for a PR.

```py
from functools import lru_cache
from typing import TYPE_CHECKING, Any, Dict, Optional, Sequence, cast

from django.apps import apps
from django.core.exceptions import ImproperlyConfigured
from django.urls import resolve, reverse
from oscar.core.application import OscarDashboardConfig
from oscar.views.decorators import check_permissions

if TYPE_CHECKING:
    from authentication.models import User  # noqa


@lru_cache(maxsize=2)
def _dashboard_urls_to_config() -> Dict[str, OscarDashboardConfig]:
    dashboard_configs = (
        config
        for config in apps.get_app_configs()
        if isinstance(config, OscarDashboardConfig)
    )
    urls_to_config: Dict[str, OscarDashboardConfig] = {}
    for config in dashboard_configs:
        for url in config.urls[0]:
            name = url.pattern.name
            if not name:
[...]

PR #3087, opened 2019-07-22
Fix dashboard permissions for forked apps

rik:
When looking up a URL for a forked app, the previous mechanism fetched the top level AppConfig which allows every user.
The new mechanism creates a dict of dashboard URLs/configs and does a lookup to find a DashboardConfig.

This will also raise NoReverseMatch if the URL does not exist or is not a dashboard URL.

fix https://github.com/django-oscar/django-oscar/issues/3080

### DIFF
--- a/src/oscar/apps/dashboard/nav.py
+++ b/src/oscar/apps/dashboard/nav.py
@@ -1,4 +1,5 @@
 import logging
+import warnings
 from functools import lru_cache
 
 from django.apps import apps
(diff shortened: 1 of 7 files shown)


rik:
Only forking the dashboard app broke two tests (`test_non_staff_user_has_empty_menu` and `test_default_access_fn_non_staff_user`) so there was no need to fork more.

PR #3097, opened 2019-07-30
Updated README with correct link to the slack workspace

maerteijn:


README.rst
```diff
@@ -78,7 +78,7 @@ Docs status:
 .. _`django-oscar-jobs group`: https://groups.google.com/forum/?fromgroups#!forum/django-oscar-jobs
 .. _`PyPI page`: https://pypi.python.org/pypi/django-oscar/
 .. _`Transifex project`: https://www.transifex.com/projects/p/django-oscar/
-.. _`Slack`: https://slack.oscarcommerce.com/
+.. _`Slack`: /http://django-oscar.slack.com/
```
rik: There is an extra leading `/`, I think this will break links.

solarissmoke:
@maerteijn it's a bit more complicated than this unfortunately. The slack.oscarcommerce.com URL was doing something clever to invite the user to the workspace. django-oscar.slack.com cannot be accessed by someone who has not been invited, so pointing people there isn't much help.

I believe @mvantellingen set up that endpoint and he me able to shed light on what it is doing, and whether we need to move that service elsewhere.

solarissmoke:
See https://github.com/django-oscar/django-oscar/commit/b1da8e0f221a5a57a293a6b432a671c0545bfa70 - we have tried this before :).

solarissmoke:
I've pinged Michael on slack and will try and find a way to restore that service. Closing this, as it doesn't fix the problem of people not being able to join the workspace. In the mean time, for anyone wanting to join, [please fill in this form](https://forms.gle/Vpn2EnTjQ68C1jyG8) and we'll invite manually.

PR #3103, opened 2019-08-01
refactor ProductQuerySet.base_queryset to use Exists instead of Count to check product has options

### LINKED ISSUE #3075: ProductQuerySet.base_queryset() should not perform annotations
The annotations performed in [`base_queryset()`](https://github.com/django-oscar/django-oscar/blob/a5096da38a265925f768ec5ce5258e9b7b1a2d72/src/oscar/apps/catalogue/managers.py#L97-L98) are very expensive. 

We have a catalogue of 250,000 products, where loading the dashboard products list view takes about 6-7 seconds because of these annotations - despite the fact that they are not used anywhere in that view. Removing the annotation drops this to less than 0.5 seconds.

I propose that the base queryset should be just that - a base queryset without any annotations. The views that need to use options should add annotations when preparing their context. The same argument could actually apply for the prefetching being done here. I think we need to review all the places that `base_queryset()` is used, and whether it is more optimal to perform the prefetching at point of use.

v1kku:
Fixes #3075.

### DIFF
--- a/src/oscar/apps/catalogue/abstract_models.py
+++ b/src/oscar/apps/catalogue/abstract_models.py
@@ -499,10 +499,10 @@ class AbstractProduct(models.Model):
     def has_options(self):
         # Extracting annotated value with number of product class options
         # from product list queryset.
-        num_product_class_options = getattr(self, 'num_product_class_options', None)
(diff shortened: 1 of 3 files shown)


v1kku:
Supposed to address #3075

PR #3106, opened 2019-08-03
Update sandbox.rst

rogererens:
Use the same option in the virtualenv command as the mkvirtualenv command uses.
apt-get is nowadays shortened to apt.

Might be helpful to advise to run
sudo apt update && sudo apt upgrade
before.


rogererens:
I also needed to run
sudo apt install build-essential
sudo apt install npm
sudo apt install libpq-dev
sudo apt install postgresql-common

(running Ubuntu 18.04 via WSL on Windows10)

PR #3109, opened 2019-08-09
Update several npm dependencies.

solarissmoke:


rik:
Have you looked at https://dependabot.com for this kind of task? It's similar to PyUp but covers more package managers.

solarissmoke:
I was literally in the process of setting up dependabot now!

PR #3126, opened 2019-08-10
Migrate deployed sandbox to heroku

### LINKED ISSUE #3125: Sandbox Site is not showing images (404s)
### Issue Summary

When visiting the Sandbox site some of the images are returning a 404

### Steps to Reproduce

1. Visit https://latest.oscarcommerce.com/en-gb/catalogue/
2. Notice most of the images are missing
3. Open up the Javascript Console and you will 404s

The site should load images and not show 404s. ;) 

Screen Shot 2019-08-09 at 11 48 10 AM

solarissmoke:
The sandbox site currently deployed at https://latest.oscarcommerce.com/ is broken, and it seems nobody with the ability to manage/access it is around any more. 

I have set up a new sandbox on Heroku: https://oscar-sandbox.herokuapp.com/, using the configuration that this PR adds. 

Using Heroku for this will also give us the additional advantage of being able to use [review apps](https://devcenter.heroku.com/articles/github-integration-review-apps) for PRs - i.e., to automatically fire up an instance of the sandbox for any PR, to facilitate easy testing.

Because Heroku uses an ephemeral filesystem, any changes made on the deployed sandbox will be cleared periodically without any intervention required.

Fixes #3125.

### DIFF
--- a/sandbox/settings.py
+++ b/sandbox/settings.py
@@ -10,22 +10,10 @@ location = lambda x: os.path.join(
 
 DEBUG = env.bool('DEBUG', default=True)
 
-ALLOWED_HOSTS = [
-    'latest.oscarcommerce.com',
-    'master.oscarcommerce.com',
-    'localhost',
-    '127.0.0.1',
-]
+ALLOWED_HOSTS = env.list('ALLOWED_HOSTS', default=['localhost', '127.0.0.1'])
 
-# This is needed for the hosted version of the sandbox
-ADMINS = (
-    ('David Winterbottom', 'david.winterbottom@gmail.com'),
-    ('Michael van Tellingen', 'michaelvantellingen@gmail.com'),
-)
 EMAIL_SUBJECT_PREFIX = '[Oscar sandbox] '
 EMAIL_BACKEND = 'django.core.mail.backends.console.EmailBackend'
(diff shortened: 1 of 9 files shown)


sandbox/settings.py
```diff
@@ -119,7 +107,7 @@
 )
 
 # Make this unique, and don't share it with anybody.
-SECRET_KEY = '$)a7n&o80u!6y5t-+jrd3)3!%vh&shg$wqpjpxc!ar&p#!)n1a'
+SECRET_KEY = env.str('SECRET_KEY', default='$)a7n&o80u!6y5t-+jrd3)3!%vh&shg$wqpjpxc!ar&p#!)n1a')
```
craigloftus: Seems like this line is causing an issue with the sandbox and docs suites, though, on the face of it, I can't see what the issue is.

solarissmoke: Apparently django-environ doesn't handle special characters in strings too well.

rik: A $ prefix is a special case for proxy values: https://django-environ.readthedocs.io/en/latest/#proxy-value

solarissmoke: Good to know, thanks!

PR #3137, opened 2019-08-13
Fix coverage reporting on travis

solarissmoke:
Coverage reports on codecov are currently broken - it turns out that the issue was missing `__init__.py` files in some test directories, which for reasons I don't fully understand causes coverage to bork on travis (works fine locally).

While I was there I've also removed a redundant reporting step - codecov-py does this itself.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -77,5 +77,4 @@ script:
 
 
 after_success:
-  - tox -e coverage-report
   - codecov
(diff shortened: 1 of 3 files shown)


solarissmoke:
This doesn't work - coverage is still not being processed by codecov. I can't see what the issue is (the report is being sent to their servers) and have emailed them for help.

solarissmoke:
@craigloftus I figured out the issue in the end - you should see codecov reporting something on this PR. It will say no base coverage because master currently has no coverage report.

PR #3141, opened 2019-08-16
Fix/file field upload to option

### LINKED ISSUE #3057: Changing OSCAR_IMAGE_FOLDER requires migrations
Changing the `OSCAR_IMAGE_FOLDER` setting causes Django to require migrations, which is problematic if a project has not forked the catalogue app.

The solution is the same as what we did in #2078 - to use a callable instead of directly referencing the setting, so that new migrations are not triggered every time the setting is changed.

jwayodi:
Fixes #3057.


tests/_site/apps/catalogue/migrations/0016_auto_20190327_0757.py
```diff
@@ -0,0 +1,40 @@
+# -*- coding: utf-8 -*-
```
solarissmoke: This migration doesn't appear to be related to the change you're making here?

src/oscar/utils/models.py
```diff
@@ -0,0 +1,8 @@
+import os
+
+from django.conf import settings
+from django.utils import timezone
+
+
+def get_image_upload_path(instance, filename):
+    return os.path.join(timezone.now().strftime(settings.OSCAR_IMAGE_FOLDER), filename)
```
solarissmoke: Suggest we use `datetime.datetime.now()` to be consistent with [what Django does](https://github.com/django/django/blob/aed89adad54a977829c4f180b036033e031ebcc7/django/db/models/fields/files.py#L305) - there is a difference in terms of timezone-awareness.

solarissmoke: We should also use `posixpath.join` as Django does.

tests/_site/apps/catalogue/migrations/0016_auto_20190327_0757.py
jwayodi: It is a dependency for `0017_auto_20190816_0938`, which has also been copied from the parent `catalogue` app.

Should I copy it over in a separate PR? I also see a couple of other forked apps whose migrations do not match those in their parent apps?

solarissmoke: Yes please

tests/_site/apps/catalogue/migrations/0017_auto_20190816_0938.py
```diff
@@ -0,0 +1,29 @@
+# Generated by Django 2.2.4 on 2019-08-16 08:38
+
+from django.db import migrations, models
+import oscar.utils.models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('catalogue', '0016_auto_20190327_0757'),
```
jwayodi: This migrations is currently missing, but is added in PR #3143.

Issue #3029, opened 2019-05-13
Transifex push is failing PR builds


solarissmoke:
PR builds are currently failing because the transifex command is asking for an API key by user input. Meanwhile the transifex push doesn't actually happen on commits to master - I don't think it has been working for quite some time.

PR #3170, opened 2019-09-05
Add translation context for "Re-order"

rik:
I have not generated the PO files in light of https://github.com/django-oscar/django-oscar/issues/3029. Should I?


rik:
> Thanks. PO files we fetch from Transifex - I will push the MOs there manually.

Isn't that the opposite ?

solarissmoke:
Yes, sorry.

PR #3177, opened 2019-09-18
Fix for missing default in currency filter

bartekgruszka:
Fallback to default is missing in line 29


src/oscar/templatetags/currency_filters.py
```diff
@@ -10,7 +10,7 @@
 
 
 @register.filter(name='currency')
-def currency(value, currency=None):
+def currency(value, currency=settings.OSCAR_DEFAULT_CURRENCY):
```
solarissmoke: I'd prefer not to specify the default here, because Python evaluates defaults at runtime, meaning overriding settings becomes tricky. Instead it's better to do set this inside the function if `currency` is `None`.

PR #3193, opened 2019-10-04
add OSCAR_THUMBNAIL_DEBUG setting that controls whether oscar_thumbnail templatetag raises errors or not

v1kku:


docs/source/ref/settings.rst
```diff
@@ -426,6 +426,13 @@ Thumbnailer class that will be used to generate thumbnails. Available options:
 ``pip install django-oscar[easy-thumbnails]``. Custom thumbnailer class (based on
 ``oscar.core.thumbnails.AbstractThumbnailer``) can be used as well.
 
+``OSCAR_THUMBNAIL_DEBUG``
+-------------------------
+
+Default: ``Same as 'DEBUG'``
```
solarissmoke: This renders oddly. I think it's better only to render the `DEBUG` bit as a code string.

docs/source/ref/settings.rst
```diff
@@ -426,6 +426,13 @@ Thumbnailer class that will be used to generate thumbnails. Available options:
...
+
+When set to True the ThumbnailNode.render method can raise errors. Django recommends that tags never raise errors in the Node.render method but since the thumbnail generation (which is outside django-oscar's control) can fail with errors, we need to be able to debug that.
```
solarissmoke: Please wrap variables and class names in double backticks. Also suggest we change the last sentence to "Django recommends that tags never raise errors in the `Node.render` method in production.". (The fact that we're allowing debug seems self-evident).

docs/source/ref/templatetags.rst
```diff
@@ -159,6 +159,8 @@ or
 
     {% oscar_thumbnail [source] [size] [options] as [variable] %}
 
+When the `OSCAR_THUMBNAIL_DEBUG` setting is set to True, this templatetag will fail with an error if an exception is raised while generating the thumbnail. If set to False, an empty string is returned.
```
solarissmoke: Please fix rendering - variables including things like `True` and `False` need to be in double backticks, not single. Also "template tag" should be two words.

src/oscar/templatetags/image_tags.py
```diff
@@ -11,6 +12,7 @@
 
 register = template.Library()
 kw_pat = re.compile(r'^(?P<key>[\w]+)=(?P<value>.+)$')
+logger = logging.getLogger()
```
solarissmoke: We should use an explicit namespace for the logger - suggest `oscar.thumbnail`.

tests/integration/templatetags/test_oscar_thumbnail.py
```diff
@@ -59,6 +60,26 @@ def test_oscar_thumbnail_tag_as_context_value(self):
     def test_oscar_thumbnail_sizes(self):
         self._test_oscar_thumbnail_tag_sizes()
 
+    @patch('oscar.templatetags.image_tags.ThumbnailNode._render')
+    def test_ThumbnailNode_doesnt_raise_if_OSCAR_THUMBNAIL_DEBUG_is_false(self, render_mock):
```
solarissmoke: Please use lower case in function names to conform with PEP8.

tests/integration/templatetags/test_oscar_thumbnail.py
```diff
@@ -59,6 +60,26 @@ def test_oscar_thumbnail_tag_as_context_value(self):
...
+        render_mock.side_effect = ValueError()
+        with override_settings(OSCAR_THUMBNAIL_DEBUG=True):
+            with self.assertRaises(ValueError):
```
solarissmoke: This check (for an exception) contradicts what the test name says it's doing - think there should be two tests, one for each setting state.

tests/integration/templatetags/test_oscar_thumbnail.py
```diff
@@ -59,6 +60,26 @@ def test_oscar_thumbnail_tag_as_context_value(self):
...
+                self.template.render(self.context)
+
+        with override_settings(OSCAR_THUMBNAIL_DEBUG=False):
+            self.assertEqual(self.template.render(self.context), '')
+
+    @patch('oscar.templatetags.image_tags.ThumbnailNode._render')
+    def test_ThumbnailNode_defaults_to_DEBUG_if_OSCAR_THUMBNAIL_DEBUG_is_not_set(self, render_mock):
```
solarissmoke: I don't think this test is necessary - its essentially just testing the internals of `getattr`.

v1kku: I think it's needed because we're testing that the code falls back to using DEBUG and that feature could be implemented in different ways and could also be broken if that code is updated.

solarissmoke: OK, suggest we split into two tests as above.

PR #3209, opened 2019-10-23
Integrate Surcharges in oscar

viggo-devries:


src/oscar/apps/checkout/applicator.py
```diff
@@ -0,0 +1,53 @@
...
+    methods = {}
+
+    def __init__(self, methods, total):
+        self.total = total
+        self.methods = methods
+
+
+class SurchargeApplicator(object):
+
+    methods = (
```
specialunderwear: No need to define this on the class, just return it in get_surcharges

src/oscar/apps/checkout/applicator.py
```diff
@@ -0,0 +1,53 @@
...
+        PaymentCharge(charge_excl_tax=D("10.0"), charge_incl_tax=D("10.0"), charge_is_percentage=True),
+        PaymentCharge(charge_excl_tax=D("20.0"), charge_incl_tax=D("20.0")),
+    )
+
+    def get_surcharges(self, basket, shipping_addr=None, **kwargs):
+        return self.methods
+
+    def get_applicable_surcharges(self, basket, shipping_addr=None, **kwargs):
+        methods = [{
```
specialunderwear: Instead of using a dict, please implement an object and implement ```__add__`` so you can just use ```sum``` in get_surcharges_total.

src/oscar/apps/checkout/applicator.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def __init__(self, methods, total):
+        self.total = total
+        self.methods = methods
+
+
+class SurchargeApplicator(object):
+
+    methods = (
+        PaymentCharge(charge_excl_tax=D("10.0"), charge_incl_tax=D("10.0"), charge_is_percentage=True),
```
specialunderwear: please just use excl_tax and incl_tax instead of charge_excl_tax so it behaves more like a price object

src/oscar/apps/checkout/surcharges.py
```diff
@@ -0,0 +1,63 @@
...
+    name = "Default surcharge"
+
+    description = ""
+
+    charge_is_percentage = False
+
+    charge_excl_tax = None
+    charge_incl_tax = None
+
+    charge_is_percentage = False
```
specialunderwear: This should go in the implementation instead of bbeing a flag. EG. implement a PercentageSurcharge and a FlatSurcharge that differ in their calculate method instead of having 2 different surcharges in 1 object.

src/oscar/apps/order/migrations/0009_order_surcharges.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 2.2.6 on 2019-10-21 13:55
```
specialunderwear: PLease add only 1 migration

src/oscar/apps/checkout/applicator.py
```diff
@@ -0,0 +1,81 @@
+from oscar.core import prices
+from oscar.core.loading import get_classes
+
+PercentageCharge, FlatCharge = get_classes("checkout.surcharges", ["PercentageCharge", "FlatCharge"])
```
specialunderwear: these are not used

src/oscar/apps/checkout/calculators.py
```diff
@@ -13,10 +15,18 @@ def __init__(self, request=None):
         # always changes the order total.
         self.request = request
 
-    def calculate(self, basket, shipping_charge, **kwargs):
-        excl_tax = basket.total_excl_tax + shipping_charge.excl_tax
-        if basket.is_tax_known and shipping_charge.is_tax_known:
-            incl_tax = basket.total_incl_tax + shipping_charge.incl_tax
+    def calculate(self, basket, shipping_charge, surcharges, **kwargs):
+        if surcharges is None:
```
specialunderwear: instead of creating a zero value surcharge why not move the surchage check below shipping charge check and add to excl_tax and incl_tax if required

src/oscar/apps/order/abstract_models.py
```diff
@@ -1200,3 +1208,29 @@ def description(self):
...
+        'order.Order',
+        on_delete=models.CASCADE,
+        related_name="surcharges",
+        verbose_name=_("Surcharges"))
+
+    name = models.CharField(
+        _("Surcharge"), max_length=128
+    )
+
+    description = models.TextField(_('Description'), blank=True)
```
specialunderwear: Is this field used anywhere?

src/oscar/apps/checkout/applicator.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+class SurchargeList(object):
```
specialunderwear: maybe it is better to make the SurchargeList a real list so you can loop through the methods directly:
```
class SurchargeList(list):

  @property
  def total(self):
    return sum(self).price
```
less code and no need for a constructor.

src/oscar/apps/order/abstract_models.py
ViggodeVriesUwkm: No it is not. I removed it.

ViggodeVriesUwkm: I also removed it from the the Base object in apps.checkout.surcharges

ViggodeVriesUwkm:
Implements surcharges in Oscar. 

A surcharge, also known as checkout fee, is an extra fee charged by a merchant when receiving a payment by cheque, credit card, charge card or debit card (but not cash) which at least covers the cost to the merchant of accepting that means of payment, such as the merchant service fee imposed by a credit card company.

So this surcharge can be used as a payment fee but also as any other fee that you want to have.
For example service costs.

You can use the calculate method on a Surcharge to do all sorts of calculations. 
Surcharge can even be implemented into a model if you want.

Overwrite the oscar.apps.checkout.applicator.SurchargeApplicator.get_surcharges to implement some surcharges. 
After that you can also overwrite get_applicable and based on the basket return if it's applicable to that person or whatever.

viggo-devries:
> rebase please

It's rebased now, please review.

sasha0:
It will definitely need at least some documentation.

viggo-devries:
I added a new how to to the docs called surcharges.

src/oscar/apps/checkout/applicator.py
```diff
@@ -0,0 +1,55 @@
+class SurchargeList(list):
+    @property
+    def total(self):
+        return sum([surcharge.price for surcharge in self])
+
+
+class SurchargePrice(object):
```
solarissmoke, line 7: We don't need to support old-style classes any more, having dropped Python 2 support  - i.e., this should just be `class SurchargePrice():`

src/oscar/apps/checkout/applicator.py
```diff
@@ -0,0 +1,55 @@
...
+    surcharge = None
+    price = None
+
+    def __init__(self, surcharge, price):
+        self.surcharge = surcharge
+        self.price = price
+
+
+class SurchargeApplicator(object):
```
solarissmoke, line 16: As above, please remove `object`.

src/oscar/apps/checkout/surcharges.py
```diff
@@ -0,0 +1,52 @@
+from decimal import Decimal as D
+
+from oscar.core import prices
+
+
+class Base(object):
```
solarissmoke, line 6: Please remove `object`. I also suggest we name this `BaseSurcharge` to avoid ambiguity when imported by projects.

src/oscar/apps/checkout/surcharges.py
```diff
@@ -0,0 +1,52 @@
...
+
+    The interface is all properties.
+    """
+
+    def calculate(self, basket, **kwargs):
+        raise NotImplementedError
+
+
+class PercentageCharge(Base):
+    name = "Percentage surcharge"
```
solarissmoke, line 21: I think this needs to be translatable?

src/oscar/apps/checkout/surcharges.py
```diff
@@ -0,0 +1,52 @@
...
+        else:
+            return prices.Price(
+                currency=basket.currency,
+                excl_tax=D('0.0'),
+                incl_tax=D('0.0')
+            )
+
+
+class FlatCharge(Base):
+    name = "Flat surcharge"
```
solarissmoke, line 42: Needs translation.

src/oscar/apps/order/admin.py
```diff
@@ -81,3 +82,4 @@ class OrderDiscountAdmin(admin.ModelAdmin):
 admin.site.register(OrderDiscount, OrderDiscountAdmin)
 admin.site.register(CommunicationEvent)
 admin.site.register(BillingAddress)
+admin.site.register(Surcharge)
```
solarissmoke, line 85: We should use a custom admin class here and set `raw_id_fields` for things like Order - otherwise on a site with thousands of orders this becomes unusable.

src/oscar/apps/order/abstract_models.py
```diff
@@ -1169,3 +1177,27 @@ def description(self):
...
+    incl_tax = models.DecimalField(
+        _("Surcharge (inc. tax)"), decimal_places=2, max_digits=12,
+        default=0)
+
+    excl_tax = models.DecimalField(
+        _("Surcharge (excl. tax)"), decimal_places=2, max_digits=12,
+        default=0)
+
+    class Meta:
+        abstract = True
```
solarissmoke, line 1202: Please define a default `ordering` here - otherwise surcharges will be displayed in an unpredictable order. Suggest we order by name.

src/oscar/templates/oscar/order/partials/basket_totals.html
```diff
@@ -84,6 +84,32 @@
...
+                <tr>
+                    <th>&nbsp;</th>
+                    <td></td>
+                </tr>
+                <tr>
+                    <th colspan="2"><h3>{% trans "Surcharges" %}</h3></th>
+                </tr>
+                {% for charge in surcharges %}
+                <tr>
+                        <th class="total">{{charge.name}}</th>
```
solarissmoke, line 99: Minor, but please add spaces for consistency - `{{ charge.name }}`.

tests/integration/checkout/test_mixins.py
```diff
@@ -114,6 +121,12 @@ def test_multi_site(self):
                                  'request': request}
         OrderPlacementMixin().place_order(**order_submission_data)
         order1 = Order.objects.get(number='12345')
+        for charge in surcharges:
```
solarissmoke, line 124: Is this code doing anything? These objects aren't being created/saved?

tests/integration/checkout/test_mixins.py
```diff
@@ -147,6 +162,12 @@ def test_multiple_payment_events(self):
                                  'shipping_address': shipping_address}
         order_placement.place_order(**order_submission_data)
         order1 = Order.objects.get(number='12345')
+        for charge in surcharges:
```
solarissmoke, line 165: As above, these aren't being saved?

src/oscar/apps/checkout/applicator.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def get_surcharges(self, basket, **kwargs):
+        """
+        For example::
+            return (
+                PercentageCharge(percentage=D("2.00")),
+                FlatCharge(excl_tax=D("20.0"), incl_tax=D("20.0")),
+            )
+
+        Surcharges must implement the minimal api in ``oscar.apps.checkout.surcharges.Base``.
```
solarissmoke, line 30: sp. `api` -> `API`.

src/oscar/apps/checkout/surcharges.py
solarissmoke: Given that names need to be translatable I think we also need a `code` here (same as we have for shipping methods) so that the charge can be uniquely identified.

src/oscar/apps/order/abstract_models.py
```diff
@@ -1169,3 +1177,27 @@ def description(self):
...
+
+
+class AbstractSurcharge(models.Model):
+    order = models.ForeignKey(
+        'order.Order',
+        on_delete=models.CASCADE,
+        related_name="surcharges",
+        verbose_name=_("Surcharges"))
+
+    name = models.CharField(
```
solarissmoke, line 1189: Suggest we add a `code` field here, which is the non-translatable identifier for a charge.

src/oscar/templates/oscar/basket/partials/basket_totals.html
```diff
@@ -149,6 +149,36 @@ <h3>{% trans "Shipping" %}</h3>
...
+
+                    <tr>
+                        <th colspan="2">
+                            <h3>{% trans "Surcharges" %}</h3>
+                        </th>
+                    </tr>
+                    {% for surcharge in surcharges %}
+                        <tr> 
+                            <th class="total"> 
+                                {{surcharge.surcharge.name}}
```
solarissmoke, line 167: For consistency please add spaces inside the brackets.

src/oscar/templates/oscar/dashboard/orders/order_detail.html
```diff
@@ -237,6 +237,20 @@ <h3>{% trans "Items ordered" %}</h3>
                                     </tr>
                                 {% endif %}
 
+                                {% with surcharges=order.surcharges.all %}
+                                {% if surcharges %}
+                                    {% for charge in surcharges %}
+                                        <tr>
+                                            <td colspan="9"></td>
+                                            <th>{% trans "Surcharge" %} ({{charge.name}})</th>
```
solarissmoke, line 245: We need to use `blocktrans` here: 

```
{% blocktrans with name=charge.name %}Surcharge ({{ name }}){% endblocktrans %}
```

src/oscar/apps/checkout/surcharges.py
```diff
@@ -0,0 +1,52 @@
+from decimal import Decimal as D
```
solarissmoke, line 1: I think we should have unit tests for each of the built-in charge classes below - percentage and flat - to verify that their output is what we would expect.

viggo-devries:
Hey @solarissmoke thanks for reviewing. I hope to process these on sunday!

src/oscar/apps/order/abstract_models.py
```diff
@@ -1169,3 +1177,27 @@ def description(self):
...
+    )
+
+    incl_tax = models.DecimalField(
+        _("Surcharge (inc. tax)"), decimal_places=2, max_digits=12,
+        default=0)
+
+    excl_tax = models.DecimalField(
+        _("Surcharge (excl. tax)"), decimal_places=2, max_digits=12,
+        default=0)
+
```
viggo-devries, line 1200: Add tax property

src/oscar/apps/checkout/surcharges.py
viggo-devries: Yea I agree. Ik won't be able to do it today. I hope to do this in the upcoming week.

viggo-devries: I added tests in tests/integration/checkout/test_surcharges.py

viggo-devries:
@solarissmoke all your reviews are processed.

src/oscar/apps/checkout/applicator.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def get_surcharges(self, basket, **kwargs):
+        """
+        For example::
+            return (
+                PercentageCharge(percentage=D("2.00")),
+                FlatCharge(excl_tax=D("20.0"), incl_tax=D("20.0")),
+            )
+
+        Surcharges must implement the minimal API in ``oscar.apps.checkout.surcharges.Base``.
```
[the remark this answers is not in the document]
viggo-devries: Fixed

PR #3211, opened 2019-10-23
Exclude products after completing the selection.

### LINKED ISSUE #3210: It is not possible to exclude specific child products from a product range that includes the parent product.
Because the range query selects all children AFTER applying the excludes it is not possible to exclude specific children.

specialunderwear:
fixes #3210


specialunderwear:
Ofcourse

specialunderwear:
Travis crashed

PR #3214, opened 2019-10-24
Pass request object to EmailAuthenticationForm

### LINKED ISSUE #3213: Request not being passed to authenticate method
### Issue Summary

Django-oscar's [`EmailAuthenticationForm`](https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/customer/forms.py#L76) is a subclass of Django's [`AuthenticationForm`](https://github.com/django/django/blob/master/django/contrib/auth/forms.py#L163).
When calling `clean` method, django `authenticate` is called without `request` object.

Looking [at the code](https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/customer/views.py#L136), the `AccountAuthView` is not passing the `request` object to the Form.

Some custom backend, like django-axes's, requires that in order to work properly.
That's why [this issue](https://github.com/django-oscar/django-oscar/issues/2111) was happening

### Steps to Reproduce

1. Create a simple authentication backend that requires the `request` object, and put it before django-oscar's backend:

```
# project/backends.py

from django.contrib.auth.backends import ModelBackend


class CustomBackend(ModelBackend):
    def authenticate(self, request, username: str = None, password: str = None, **kwargs: dict):
        

        if request is None:
            raise Exception('CustomBackend requires a request as an argument to authenticate')
       # some logic

# On settings

AUTHENTICATION_BACKENDS = [
    'project.backends.CustomBackend',
    'oscar.apps.customer.auth_backends.EmailBackend',
]

```
2. Try to login
3. Observe `Exception('CustomBackend requires a request as an argument to authenticate')` is raised

### Technical details

```
$ python --version
Python 3.7.4
$ pip freeze | grep Django
Django==2.2.6
$ pip freeze | grep django-oscar
django-oscar==2.0.3
```

leandro-gomez:
Add request object to login form kwargs

Fixes #3213

### DIFF
--- a/src/oscar/apps/customer/views.py
+++ b/src/oscar/apps/customer/views.py
@@ -135,6 +135,7 @@ class AccountAuthView(RegisterUserMixin, generic.TemplateView):
 
     def get_login_form_kwargs(self, bind_data=False):
         kwargs = {}
+        kwargs['request'] = self.request
         kwargs['host'] = self.request.get_host()
         kwargs['prefix'] = self.login_prefix
         kwargs['initial'] = {
--- a/tests/unit/customer/test_views.py
+++ b/tests/unit/customer/test_views.py
@@ -1,6 +1,6 @@
-from unittest.mock import patch, Mock
+from unittest.mock import Mock, patch
 
-from django.test import TestCase, Client
+from django.test import Client, TestCase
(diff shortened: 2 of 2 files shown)


solarissmoke:
Thanks for this!

PR #3217, opened 2019-10-25
Ignore expired vouchers because basket also ignores them.

### LINKED ISSUE #3216: Vouchers that have expired since adding cause error without explanation on order creation
While the basket ignores vouchers that are expired since the order was created, the OrderPlacementMixin raises an error with an empty explanation.

specialunderwear:
If not the order placement fails with an empty error message.

fixes #3216

### DIFF
--- a/src/oscar/apps/order/utils.py
+++ b/src/oscar/apps/order/utils.py
@@ -66,9 +66,12 @@ class OrderCreator(object):
                 self.update_stock_records(line)
 
             for voucher in basket.vouchers.select_for_update():
-                available_to_user, msg = voucher.is_available_to_user(user=user)
-                if not voucher.is_active() or not available_to_user:
-                    raise ValueError(msg)
+                if not voucher.is_active():  # basket ignores inactive vouchers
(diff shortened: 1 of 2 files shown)


solarissmoke:
Is there an edge case here where the order total was determined before the voucher expired? In which case the total will be wrong? I feel like it would be better in this case to (a) remove expired vouchers from the basket (or report an error there) as soon as they are detected, and throw an error during order creation.

specialunderwear:
I added a test and that test fails without the fix.

solarissmoke:
@specialunderwear I'm not disputing that your changes fix the test you have added. 

What I am saying is that I think this introduces the possibility that Oscar can create an order whose totals are invalid, because they include a discount from a now-removed voucher. I am not convinced that the correct behaviour is to silently ignore the voucher at this stage (because it could have been valid when the order totals were calculated). Doesn't it make more sense to report a more meaningful error here?

specialunderwear:
It can happen only if a voucher expires during the request after computing the basket total.

Issue #3225, opened 2019-11-04
Invalid attributes provide no feedback when creating/modifying a product


rik:
### Issue Summary

The browser HTML5 form validation kicks in but does not focus the Attributes section so it appears broken

### Steps to Reproduce
1. Create a product class "class-for-test" with one required attribute
2. Go to https://your-shop.com/dashboard/catalogue/products/create/class-for-test
3. Click the Save button

### Actual behaviour
No error message is visible and it feels like the Save button is not working. (I've received many complaints from dashboard users)

### Expected behaviour
The attribute section should be focused so that the browser error tooltips are visible.

PR #3226, opened 2019-11-04
Display tabs that have invalid input fields

rik:
Listen to browser form validation events to show tabs with issues

close https://github.com/django-oscar/django-oscar/issues/3225


src/oscar/static/oscar/js/oscar/dashboard.js
```diff
@@ -275,6 +275,12 @@ var oscar = (function(o, $) {
                 if (!form || $(":invalid", form).length == 0)
                     $(this).button('loading');
             });
+
+            // Display tabs that have invalid input fields
+            $('input').on('invalid', function(){
+                var id = $(this).closest('.tab-pane').attr('id');
+                $('.nav-list a[href="#' + id + '"]').tab('show');
```
solarissmoke: Could we add a check here that `id` isn't `undefined` before attempting to show the tab? This code executes for all dashboard forms, including those (e.g., voucher creation form) that don't have a tabbed interface.

rik: We could but it does not trigger an error at the moment. jQuery returns an empty jQuery collection when `id` is `undefined`.

solarissmoke: The string concatenation means that the second line looks for a DOM element with `href="#undefined`". While this doesn't raise an error I would prefer not to rely on there being no such element in the DOM - it's a potential seed for confusion when unrelated things are changed in future.

rik: Fair point :)

PR #3236, opened 2019-11-20
Refactor asset management so that third party assets are not committed.

solarissmoke:
We have been committing third part assets like Bootstrap, jQuery etc, which is not ideal. This PR refactors to put Oscar's own assets in `static_src`, and modifies the asset pipeline to pull in third party assets at build time rather than committing them to the repository. 

This doesn't affect what is shipped in the package - i.e., we will compile and build all the third party assets before packaging.

At this stage I've just refactored the existing Gulp pipeline - I think the next step is to look at migrating to something like webpack.

There are also some assets e.g., from font-awesome, that are committed because we are not loading those libraries from npm. I have left that as-is for now on the basis that eventually we are moving to Bootstrap 4 and likely dropping font-awesome.


solarissmoke:
Worth testing this on the [review app](https://oscar-sandbox-pr-3236.herokuapp.com/) to make sure all expected assets are loading correctly.

craigloftus:
@solarissmoke I've only check the public views. Is there a way to create an admin account in the review apps? I tried the gateway view, but didn't receive an email.

solarissmoke:
@craigloftus review apps use the sandbox fixtures which provide a superuser account (`superuser@example.com`:`testing`).

Issue #1253, opened 2014-03-12
Product ranges and variant products exhibit undefined behaviour


maiksprenger:
#1218 got me to look into the Range model. There's some decisions and documentation needed regarding handling of variant products.
- Can/should one be able to add one, but not all variant products to a range?
- Can/should one apply offers to one, but not all variant products?
- If a range contains a product class or category, currently the associated variant products aren't included as they have neither. The category issue is raised in #1218

PR #3250, opened 2019-12-04
Fixes offers failing on child products for ranges that include all products

specialunderwear:
Leaving out child products from ranges causes offers to not work on child products.
Let's include them and make all products really mean all products.

### DIFF
--- a/src/oscar/apps/offer/abstract_models.py
+++ b/src/oscar/apps/offer/abstract_models.py
@@ -949,8 +949,8 @@ class AbstractRange(models.Model):
         Product = self.included_products.model
 
         if self.includes_all_products:
-            # Filter out child products and blacklisted products
-            return Product.objects.browsable().exclude(
+            # Filter out blacklisted products
+            return Product.objects.all().exclude(
                 id__in=self.excluded_products.values("id")
             )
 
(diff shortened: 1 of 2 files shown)


solarissmoke:
Looking at [where this filter was introduced](https://github.com/django-oscar/django-oscar/commit/da4f0e4dc4893ee97df500022ab21b5a99995746), it was to make this consistent with the `products()` method which [excludes child products](https://github.com/django-oscar/django-oscar/blob/f802188025f513f4e26632db02de9c03a67b246a/src/oscar/apps/offer/abstract_models.py#L482). 

That restriction was added in https://github.com/django-oscar/django-oscar/issues/1253 and the discussion on that ticket suggests that allowing child products to be independently associated with offers introduces other problems. Think it would be good to understand better what those problems are before we make this change? Presumably you have done this on your projects?

specialunderwear:
Ok I read that thread and I am not really impressed by the statements in there that are not carried by examples. Currently the ranges implementation allows including or excluding certain children and that works just fine. It will add all children if a parent is added and if you dont want them all, don't add the parent only children, or exclude children individually. That seems sane to me.

Now the issue this pull request solves, is that offers don't work at all if you are using child products and try to create an offer for all products in the site. It works just fine if you add individual categories or something, only the "include all products" option is broken because it filters out the child products which are now simply not eligable for discount.

These kinds of offers are often done when a new site goes live, eg. "10% discount on all products". So all discussions aside, that seems like an immediate problem to me.

specialunderwear:
Hi can you review this again please

PR #3264, opened 2019-12-20
Use parent class get_users() method in PasswordResetForm.

solarissmoke:
Refs CVE-2019-19844 (https://www.djangoproject.com/weblog/2019/dec/18/security-releases/). Oscar had overridden the `save()` method on `PasswordResetForm`, and was querying for users to send reset emails to in a manner that was not safe from unicode collisions.

This fixes the behaviour of the `save()` method to use the parent class' `get_users()` method, which applies the appropriate measures to avoid unicode collisions.

Oscar's handling of this form was more secure than Django, because it was sending the email to the address stored for the user in the database, rather than the address supplied in the form. As such, Oscar's password reset form has always been secure from the issue in CVE-2019-19844. 

The form handling did however allow for emails to be sent to unintended users (i.e., any user whose address collided with the one supplied in the form), which is what this patch fixes.

### DIFF
--- a/docs/source/releases/v2.1.rst
+++ b/docs/source/releases/v2.1.rst
@@ -115,6 +115,16 @@ Minor changes
 - Fixed the logic of ``StockRequired.parent_availability_policy`` to use
   child products to determine availability of children, rather than the parent.
 
+- ``customer.forms.PasswordResetForm`` now uses the parent class' ``get_users()``
+  method to determine the list of users to send an email to. ``get_users()``
+  filters out users who do not currently have a usable password - which
+  did not happen previously.
+
+  This change was made in response to changes in Django to address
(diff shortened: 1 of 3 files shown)


docs/source/releases/v2.1.rst
```diff
@@ -115,6 +115,9 @@ Minor changes
 - Fixed the logic of ``StockRequired.parent_availability_policy`` to use
   child products to determine availability of children, rather than the parent.
 
+- ``customer.forms.PasswordResetForm`` now uses the parent class' ``get_users()``
+  method to determine the list of users to send an email to.
```
craigloftus: I think you should note that the behaviour changes here, i.e., unicode safe comparison and checking that the user has a usable password.

PR #3275, opened 2020-01-04
Fix time-dependent tests.

solarissmoke:
Switch to using freezegun for tests that depend on date.


tests/integration/payment/test_forms.py
```diff
@@ -45,6 +46,7 @@ def test_rejects_none(self):
             self.field.clean(None)
 
 
+@freeze_time('2016-01-01')
```
craigloftus: I think it would be worth commenting on why a fixed date is necessary here

PR #3302, opened 2020-02-11
Port all core templates and styles to Bootstrap 4.

### LINKED ISSUE #2730: Please include support for Bootstrap 4

solarissmoke:
This is a rebase of #3005, having fixed up a few minor issues I came across when reviewing it. Closes #2730.


solarissmoke:
@samitnuk would be grateful if you can do one more check/test on this before we merge.

craigloftus (dismissed):
I've just reviewed this using the deployed sandbox and noted down a few issues I've spotted:

- [x] The 'View order status' button on the checkout thank you page is not inline with the sentence. 
- [x] The sort form on the reviews list looks unstyled? No spacing between the components, unstyled select and the go button could be less salient.
- [x] The empty state for the notifications list in the accounts page could do with some spacing to separate it from the tabs.
- [x] The search input in the header doesn't have a `name` attribute, which is causing the search form to break.
- [x] The use of outlined stars for reviews in product lists makes them not very salient, I'd suggest swapping for filled stars like on the detail pages.

solarissmoke:
@craigloftus I've addressed the issues you raised. The styling of the reviews form still isn't great - in fact that whole review thing needs an overhaul - but I think that's best left as a separate task.

solarissmoke:
@django-oscar/core this is a sufficiently big change that more testing on the review app will probably throw up a few more issues. Would be grateful if any of you have time to do some testing.

samitnuk:
@solarissmoke 

When you will have a chance, please update these two files [development-environment.rst](https://github.com/solarissmoke/django-oscar/blob/feature/bs4/docs/source/internals/contributing/development-environment.rst) and [how_to_handle_statics.rst](https://github.com/solarissmoke/django-oscar/blob/feature/bs4/docs/source/howto/how_to_handle_statics.rst) because they keep information about `LESS`. Thanks!

solarissmoke:
We've identified a few more issues on this, which we're working on fixing in the next couple of weeks.

sasha0:
We could fix by ourselves if you list out the outstanding issues.

itsthejoker:
Following up on this because I'm looking at integrating Oscar into a site that is already fully invested in bootstrap 4. Do we have an anticipated ETA for this PR?

solarissmoke:
_Possibly_ end of May - but I won't promise it. Definitely within the next couple of months.

daniel-butler:
Is this branch based on the oscar version that is only compatible with django under version 3? When I try to run the sandbox site it gives errors like `No module named 'django.utils.six'` and ` No module named 'django.utils.lru_cache'` which after google searching seem to be only django=2 utility functions.

sasha0 (changes requested):
Looks good, except 2 typos. Also I'd suggest to add Github username mentions into the credits section in the commit message.

sasha0 (dismissed):

specialunderwear (approved):

PR #3316, opened 2020-03-03
Fix Integrity error when deleting and creating a voucher with same name

### LINKED ISSUE #3279: Error if create voucher name matching to voucher name deleted
Got error 500 when create voucher

Screen Shot 2020-01-09 at 9 37 03 AM

### Steps to Reproduce
1. Create new voucher
2. Delete that voucher
3. Create voucher again with same **voucher name** deleted

capelito:
Fixes #3279

Deleting a voucher didn't delete the referenced offers so when deleting a voucher and
creating a new one with the same name, there was a unique name error because the name of an offer is "Offer for voucher [voucher name]". This fix deletes all referenced offers when
deleting a voucher.

### DIFF
--- a/docs/source/releases/v2.1.rst
+++ b/docs/source/releases/v2.1.rst
@@ -162,6 +162,9 @@ Minor changes
 - ``catalogue.Product.is_public`` is now an indexed field. This change requires
   a database migration.
 
+- When a voucher that was created through the Oscar dashboard is deleted, the
+  auto-generated offer that was created with the voucher is also deleted.
+
 Dependency changes
 ~~~~~~~~~~~~~~~~~~
 
--- a/src/oscar/apps/dashboard/vouchers/views.py
+++ b/src/oscar/apps/dashboard/vouchers/views.py
@@ -10,6 +10,7 @@ from django.utils import timezone
 from django.utils.translation import gettext_lazy as _
 from django.views import generic
 
+from oscar.apps.voucher.utils import get_offer_name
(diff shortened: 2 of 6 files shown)


src/oscar/apps/voucher/abstract_models.py
```diff
@@ -177,6 +177,11 @@ def save(self, *args, **kwargs):
         self.code = self.code.upper()
         super().save(*args, **kwargs)
 
+    def delete(self, *args, **kwargs):
+        """Deletes all referenced conditional offers"""
+        self.offers.all().delete()
```
solarissmoke: This is too aggressive and can result in data loss. Oscar's models do not prevent the same offer being used in multiple places - so indiscriminately deleting all offers could affect other vouchers or remove an offer that is used for another purpose. 

We should do something similar to [what we do for benefits/conditions](https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/offer/signals.py#L12-L31), i.e., check that :

1. Only one offer is associated with the voucher, and it is of the `voucher` type.
2. The offer name corresponds to the voucher name.

And only then delete the offer.

I also think this should be done in a `post_delete` signal so that there is no risk of integrity errors if the delete fails for another reason.

capelito: Thanks for your feedback, I will update the code accordingly.

capelito:
Hi @solarissmoke can you please take a look and let me know if I should change anything else?
Thanks

PR #3327, opened 2020-03-23
Only try to serialize PhoneNumber instances

maikhoepfel:
Dear team, I stumbled upon a small issue in my project, and am proposing a fix below.

AbstractBillingAddress has no `phone_number` field. For reasons I'm not sure about, Oscar has code dealing with it's existence. I suppose because it's common to add the field? The existing code assumed that a `PhoneNumberField` model field would be added, which might not always be the case.
I stumbled upon this in a project of mine where `phone_number` was a CharField (for reasons...). That would break with an `AttributeError`. The fix is easy enough, we check if we're indeed dealing with a `PhoneNumber` instance before calling `as_international` on it.

Note that this fix does not need to be applied to shipping addresses, because there we actually define a `PhoneNumberField`, so our assumptions hold.

### DIFF
--- a/src/oscar/apps/checkout/utils.py
+++ b/src/oscar/apps/checkout/utils.py
@@ -1,3 +1,6 @@
+from phonenumber_field.phonenumber import PhoneNumber
+
+
 class CheckoutSessionData(object):
     """
     Responsible for marshalling all the checkout session data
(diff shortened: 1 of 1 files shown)


maikhoepfel:
The test failure seems to be related to `npm`, not my changes.

solarissmoke:
@maikhoepfel thanks for this. The CI build is failing because of your changes - a flake8 error:

`src/oscar/apps/checkout/utils.py:3:1: E302 expected 2 blank lines, found 1`

Could you please update to fix that?

solarissmoke:
(Never mind - I've fixed it for you).

maikhoepfel:
Thanks! And sorry about the flake8 error.

PR #3329, opened 2020-03-24
extra in BasketAdmin Class

pandeyroshan:
Specified the extra value in the BasketAdmin class for LineInLine inlines.


solarissmoke:
@pandeyroshan can you please explain the motivation for this change?

`ModelAdmin` doesn't actually expect an `extra` variable as far as I can see - so this change as it is currently has no effect at all. `extra` is only valid on subclasses of `InlineModelAdmin`, and those default to 3 anyway.

PR #3361, opened 2020-05-11
Delete both valid and invalid basket lines marked for deletion

jwayodi:
Currently, if a valid basket line is marked for deletion, any other
invalid basket lines on the formset prevent it from being deleted.


jwayodi:
Fixes #2791.

solarissmoke:
@sasha0 how does this look to you? We've been running this fix in production for a few days.

sasha0:
Hm, looks good in general, but needs mentioning in the docs about this behaviour.

solarissmoke:
Thanks - we will mention the fix in the release notes.

PR #3364, opened 2020-05-15
Issue/1970/define ordering for models

### LINKED ISSUE #1970: Define ordering for models
I've been thinking about ordering of model instances a bit lately (#1963, https://github.com/django-oscar/django-oscar/pull/1969). I think we might be relying on the implicit order of instances returned by the database a bit too often. From the [PostgreSQL docs](http://www.postgresql.org/docs/9.1/static/queries-order.html): 

> If sorting is not chosen, the rows will be returned in an unspecified order. The actual order in that case will depend on the scan and join plan types and the order on disk, but it must not be relied on. A particular output ordering can only be guaranteed if the sort step is explicitly chosen.

We should make a pass over our models and define a default ordering where it makes sense. I think many times we assume "order of creation", but that is not guaranteed.

SameeranB:
Fixes #1970.

This adds ordering and database indexing to the following fields in their respective models.

|    |               |                                |                 |
|----|---------------|--------------------------------|-----------------|
| #  | App           | Model                          | Ordering Change |
| 1  | analytics     | AbstractUserProductView        | -date_created    |
| 2  | analytics     | AbstractUserSearch             | -date_created    |
| 3  | basket        | AbstractBasket                 | -date_created    |
| 4  | catalogue     | AbstractOption                 | name            |
| 5  | communication | AbstractEmail                  | -date_sent       |
| 6  | communication | AbstractCommunicationEventType | name            |
| 7  | customer      | AbstractProductAlert           | -date_created    |
| 8  | offer         | AbstractRange                  | name            |
| 9  | order         | AbstractOrderNote              | -date_updated   |
| 10 | order         | AbstractOrderDiscount          | pk              |
| 11 | payment       | AbstractSource                 | pk              |
| 12 | payment       | AbstractSourceType             | name            |
| 13 | voucher       | AbstractVoucherSet             | -date_created   |
| 14 | voucher       | AbstractVoucher                | -date_created   |
| 15 | voucher       | AbstractVoucherApplication     | -date_created   |


src/oscar/apps/communication/abstract_models.py
```diff
@@ -25,11 +25,12 @@ class AbstractEmail(models.Model):
     subject = models.TextField(_('Subject'), max_length=255)
     body_text = models.TextField(_("Body Text"))
     body_html = models.TextField(_("Body HTML"), blank=True)
-    date_sent = models.DateTimeField(_("Date Sent"), auto_now_add=True)
+    date_sent = models.DateTimeField(_("Date Sent"), auto_now_add=True, db_index=True)
```
solarissmoke, line 28: Let's drop the index but still retain the ordering.

src/oscar/apps/order/abstract_models.py
```diff
@@ -385,14 +385,15 @@ class AbstractOrderNote(models.Model):
 
     message = models.TextField(_("Message"))
     date_created = models.DateTimeField(_("Date Created"), auto_now_add=True)
-    date_updated = models.DateTimeField(_("Date Updated"), auto_now=True)
+    date_updated = models.DateTimeField(_("Date Updated"), auto_now=True, db_index=True)
```
solarissmoke, line 388: Let's drop the index.

src/oscar/apps/voucher/abstract_models.py
```diff
@@ -247,6 +249,7 @@ def record_usage(self, order, user):
             self.applications.create(voucher=self, order=order)
         self.num_orders += 1
         self.save()
+
```
solarissmoke, line 252: As above please review/remove white space changes where they are not needed.

src/oscar/apps/basket/abstract_models.py
```diff
@@ -66,6 +66,7 @@ class AbstractBasket(models.Model):
     class Meta:
         abstract = True
         app_label = 'basket'
+        ordering = ['-date_created']
```
solarissmoke, line 69: Thinking about this some more and I can't find a reasonable use case for wanting to query and display baskets in any particular order. I think we can probably do without ordering here.

SameeranB:
@solarissmoke, I'm failing Travis checks but I don't think it's got anything to do with my commit. Here is one of the failed tests.  Would you point me in the right direction on this?

image

solarissmoke:
It is to do with your commit. You have changes for which there are migrations missing from the PR.

solarissmoke:
It is to do with your commit. You have changes for which there are migrations missing from the PR.

SameeranB:
Yeah, I've pushed again with the missing migrations.


> @solarissmoke, ~I'm failing Travis checks but I don't think it's got anything to do with my commit. Here is one of the failed tests. Would you point me in the right direction on this?~
> 
> Apologies, It was a problem on my end. I forgot to run make migrations

requirements.txt
```diff
@@ -11,7 +11,7 @@ Pillow==7.1.2
 Whoosh==2.7.4
 django-environ==0.4.5
 django-redis==4.11.0
-pysolr==3.8.1
+pysolr==3.9.0
```
solarissmoke, line 14: This shouldn't be part of this PR?

src/oscar/apps/analytics/abstract_models.py
```diff
@@ -101,11 +102,12 @@ class AbstractUserSearch(models.Model):
         on_delete=models.CASCADE,
         verbose_name=_("User"))
     query = models.CharField(_("Search term"), max_length=255, db_index=True)
-    date_created = models.DateTimeField(_("Date Created"), auto_now_add=True)
+    date_created = models.DateTimeField(_("Date Created"), auto_now_add=True, db_index=True)
```
solarissmoke, line 105: See [my previous comment](https://github.com/django-oscar/django-oscar/pull/3364#discussion_r426118842) - please remove this index.

src/oscar/apps/analytics/abstract_models.py
```diff
@@ -101,11 +102,12 @@ class AbstractUserSearch(models.Model):
...
 
     class Meta:
         abstract = True
         app_label = 'analytics'
+        ordering = ['-date_created']
```
solarissmoke, line 110: Think you may have missed [my previous comment](https://github.com/django-oscar/django-oscar/pull/3364#discussion_r426118907) on this?

src/oscar/apps/analytics/abstract_models.py
SameeranB: Should I change the ordering of this model to 'query' as well ?

src/oscar/apps/analytics/abstract_models.py
SameeranB: The previous comment was for the basket app. Should I change the ordering here to '-pk' as well?

solarissmoke: Yes please.

solarissmoke: Could you please also delete and regenerate the migrations, so that we're left with just one new migration per app.

SameeranB: Of course!....On it...

solarissmoke:
Thanks for this. It'll be a couple of weeks before we can merge (want to get v2.1 out first).

Issue #3401, opened 2020-06-11
Question: Would a PR for Django 2.2 compatibility in the 1.x branch be accepted?


khink:
### Issue Summary

We'd like to create a Django-Oscar 1.x version (based off 1.6.7) that supports Django 2.2. Would the maintainers be willing to accept a PR for this, and make a release?

### Background

We're in the process of upgrading our Django-Oscar+API-based application to a supported version of Django. Unfortunately, while 1.6.7 does support Django 2.1, it does not support 2.2.

So we are faced with a choice:

1. Upgrade to Django-Oscar 2
2. See if we can get Django-Oscar 1 to work with Django 2.2

It seems that the second option would require the least effort. We are expecting a lot of work from the Django-Oscar 2 upgrade, and the changes between Django 2.1 and 2.2 seem minor by comparison.

We'd like to know if the maintainers think that Oscar1.x + Django2.2 compatibility might be useful enough to cut a release. I'm also wondering why there is no such branch/release yet: There might be blocking issues that i'm not aware of, so i'm asking for insights as to what i might run in to. 

### Questions

1. Would a Oscar1.x + Django2.2 release be feasible?
2. Are you aware of blocking issues that make this hard to achieve?

solarissmoke:
> Would a Oscar1.x + Django2.2 release be feasible?

Yes - happy to release this if you are prepared to do the work of making it compatible.

> Are you aware of blocking issues that make this hard to achieve?

I don't think there are any such issues. The reason it wasn't done is mainly one of capacity (i.e., the more active Oscar branches you maintain, the more work is involved), and we don't yet have an explicit [policy](https://github.com/django-oscar/django-oscar/issues/2675) on which versions are actively maintained other than the current one. If the community would benefit from 1.6.x supporting Django 2.2, and you can spend the time required on it, then let's do it. I would say however that 1.6.x is likely only to receive security patches after that.

khink:
Thanks for your speedy reply.

I'm very happy to hear that such a release is feasible. We'll start work for a proper Django-Oscar release. 

> I would say however that 1.6.x is likely only to receive security patches after that.

Totally agreed. It's just that we can't free up the time for the Oscar 2 transition at the moment.

jayvdb:
[...]

PR #3402, opened 2020-06-15
Fix tests on 1.6 branch

khink:
In the process of #3401, i found that the tests for the `releases/1.6` branch currently fail for two reasons:

1. pytest-xdist has sneakily started requiring a higher pytest version in a minor version update (to 1.23.4), see https://github.com/pytest-dev/pytest/issues/3724
2. `test_forms` relied on 2014 being a valid year for a credit card date, which  it isn't anymore.

I'd like to keep the PR for the Django 2.2 compatibility as clean as possible, so i'd like to have this PR merged before submitting the PR with the actual changes.

### DIFF
--- a/tests/integration/payment/test_forms.py
+++ b/tests/integration/payment/test_forms.py
@@ -3,7 +3,7 @@ import datetime
 from django.core.exceptions import ImproperlyConfigured
 from django.test import TestCase
 from django.forms import ValidationError
-
+from freezegun import freeze_time
 
 from oscar.apps.payment import forms, models
 # types=[bankcards.VISA, bankcards.VISA_ELECTRON, bankcards.MASTERCARD,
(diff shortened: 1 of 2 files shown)


tests/integration/payment/test_forms.py
```diff
@@ -11,6 +11,8 @@
 #               bankcards.DISCOVER, bankcards.JCB, bankcards.CHINA_UNIONPAY,
 #               bankcards.SOLO, bankcards.LASER, bankcards.SWITCH]
 
+valid_year = str(datetime.date.today().year)
```
solarissmoke, line 14: Could I request that you use the [same approach](https://github.com/django-oscar/django-oscar/blob/master/tests/integration/payment/test_forms.py#L49) that was used to fix this on master, just for consistency of approach? Using freezegun is also more deterministic.

PR #3403, opened 2020-06-15
Add Django 2.2 compatibility to 1.6 branch

### LINKED ISSUE #3401: Question: Would a PR for Django 2.2 compatibility in the 1.x branch be accepted?

khink:
As discussed in #3401, this makes the 1.6 branch compatible with Django 2.2.

- The only actual code change is backported from https://github.com/django-oscar/django-oscar/commit/4a479e74ab6cf359329f2a539f67e931410347eb
- Test matrices have been augumented with django22 on python3.5 and python3.6 
- Testing with django master branch has been dropped
- Release notes have been updated but will need a release date

Closes #3401.

### DIFF
--- a/docs/source/releases/v1.6.8.rst
+++ b/docs/source/releases/v1.6.8.rst
@@ -2,7 +2,9 @@
 Oscar 1.6.8 release notes
 =========================
 
-:release: 2018-02-28
+:release: ????-??-??
 
-This is Oscar 1.6.8, a security release that updates to Bootstrap to 3.4.1
-and extend 3.0.2 to fix CVE-2019-8331 and CVE-2018-16492 vulnerabilities.
+This is Oscar 1.6.8, a Django 2.2 compatibility release.
+
+It also updates Bootstrap to 3.4.1 and explicitly specifies extend in the
+npm dependencies to fix CVE-2019-8331 and CVE-2018-16492 vulnerabilities.
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,20 +16,28 @@ matrix:
       env: TOXENV=py35-django111
     - python: 3.6
       env: TOXENV=py36-django111
+    - python: 3.7
+      env: TOXENV=py37-django111
     - python: 3.4
       env: TOXENV=py34-django20
     - python: 3.5
       env: TOXENV=py35-django20
     - python: 3.6
       env: TOXENV=py36-django20
+    - python: 3.7
+      env: TOXENV=py37-django20
     - python: 3.5
       env: TOXENV=py35-django21
     - python: 3.6
       env: TOXENV=py36-django21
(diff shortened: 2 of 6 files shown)


khink:
@solarissmoke I think we might also add python3.7 to the test matrix without effort, do you agree?

solarissmoke:
> I think we might also add python3.7 to the test matrix

Yes.

docs/source/releases/v1.6.8.rst
```diff
@@ -2,7 +2,9 @@
...
 =========================
 
-:release: 2018-02-28
+:release: ????-??-??
 
-This is Oscar 1.6.8, a security release that updates to Bootstrap to 3.4.1
-and extend 3.0.2 to fix CVE-2019-8331 and CVE-2018-16492 vulnerabilities.
+This is Oscar 1.6.8, a Django 2.2 compatibility release.
+
+It updates to Bootstrap to 3.4.1 and extends 3.0.2
```
solarissmoke, line 9: It's not clear to me what "extends 3.0.2" is referring to here. Did this make sense to you / could you clarify what it means?

solarissmoke: There is also a typo "updates to Bootstrap to" -> "updates Bootstrap to"

khink: Now that you mention it, this didn't make sense to me. It refers to commit fe9507b8d, which backports #2952. So "Explicitly specify extend in the npm dependencies" instead of "extends 3.0.2" would be better.

solarissmoke:
This looks good - will merge and cut a release within the next week.

solarissmoke:
https://pypi.org/project/django-oscar/1.6.8/

PR #3412, opened 2020-06-19
fork_app: Create init in forked app

jayvdb:


src/oscar/core/customisation.py
```diff
@@ -45,10 +45,21 @@ def inherit_app_config(local_app_folder_path, local_app_name, app_config):
...
+        "    name = '{app_name}'\n"
+        "    label = '{label}'\n".format(
             app_config_class_module=app_config.__module__,
             app_config_class_name=app_config.__class__.__name__,
-            app_name=local_app_name))
+            app_name=local_app_name,
+            label=label,
+            ))
+
+    create_file(
```
solarissmoke, line 56: @jayvdb is this not duplicating the same logic as inlines 39-43 above?

jayvdb: Ahh , this was added since v2.0 which is where I needed it.  Very good.

What about the addition of `label`?

If not, pls close.

If it is beneficial, I'll tidy up the PR.

solarissmoke: The label is inherited from the parent class - I don't think things would work if you change that label at project level.

jayvdb: Ya true. ok, 628bd1e10 makes this redundant then.

PR #3416, opened 2020-06-22
Unlock button for generating report when it starts to download

MOwsianowski:
Now:
smart_generating_report_button
When download is started the state of the button is reset. 

Before:
Button all the time was blocked ("Generating..."), page refresh was needed.


solarissmoke:
Thanks for this. I'm not convinced that the complexity introduced here is worth the benefit though. I think we could just remove the `data-loading` attribute on this button, so that it never gets disabled. The attribute is intended to prevent multiple submissions of forms, and is important when a form modifies state. In this case the action is a read-only query and the worst that would happen is that you run a query twice.

MOwsianowski:
@solarissmoke 
Thanks for opinion. As you said, not sure if that ~risky mechanism is worth the benefit. For sure it will be more user friendly, but this is a admin dashboard so ....  hmhm :)

solarissmoke:
Refactored in #3467.

Issue #3426, opened 2020-06-26
PostActionMixin fails with exception if action=do_foo is not present


jayvdb:
### Issue Summary

Oscar 2.0 smoke tests, but it appears the underlying issue is valid on 2.1.
Submitting a blank post to `/oscar/accounts/orders/id/` fails because it defers to `PostActionMixin`, which skips the `action=do_foo` logic and then tries to find another `.post(...)` implementation which doesnt exist.

```py
ERROR: test_smoke_POST_oscar/^accounts/^orders/(?P<order_number>[\w-]*)/$ (django_smoke_tests.tests.SmokeTests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/lib/python3.8/site-packages/django_smoke_tests/generator.py", line 90, in test
    response = http_method_function(url, {})
  File "/usr/lib/python3.8/site-packages/django/test/client.py", line 526, in post
    response = super().post(path, data=data, content_type=content_type, secure=secure, **extra)
  File "/usr/lib/python3.8/site-packages/django/test/client.py", line 355, in post
    return self.generic('POST', path, post_data, content_type,
  File "/usr/lib/python3.8/site-packages/django/test/client.py", line 421, in generic
    return self.request(**r)
  File "/usr/lib/python3.8/site-packages/django/test/client.py", line 496, in request
    raise exc_value
  File "/usr/lib/python3.8/site-packages/django/core/handlers/exception.py", line 34, in inner
    response = get_response(request)
  File "/usr/lib/python3.8/site-packages/django/core/handlers/base.py", line 115, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "/usr/lib/python3.8/site-packages/django/core/handlers/base.py", line 113, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/lib/python3.8/site-packages/django/contrib/auth/decorators.py", line 21, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "/usr/lib/python3.8/site-packages/django/views/generic/base.py", line 71, in view
    return self.dispatch(request, *args, **kwargs)
  File "/usr/lib/python3.8/site-packages/django/views/generic/base.py", line 97, in dispatch
    return handler(request, *args, **kwargs)
  File "/usr/lib/python3.8/site-packages/oscar/views/generic.py", line 30, in post
    return super().post(request, *args, **kwargs)
[...]

Issue #3420, opened 2020-06-23
model-bakery & automated admin smoke tests


jayvdb:
I have a set of model 'hacks' and other setup so that [`model-bakery`](https://pypi.org/project/model-bakery/) can be used with [`django-admin-smoke-tests` PR 18](https://github.com/greyside/django-admin-smoke-tests/pull/18) & [`django-admin-auto-tests`](https://github.com/alsur/django-admin-auto-tests) to smoke test the Oscar apps that are the core of my system.

It includes registering model bakery generators
```py
    baker.generators.add('phonenumber_field.modelfields.PhoneNumberField',
                         gen_phonenumber)

    baker.generators.add('oscar.models.fields.slugfield.SlugField',
                         gen_slug_starting_with_alpha)

    baker.generators.add('oscar.models.fields.autoslugfield.AutoSlugField',
                         gen_slug_starting_with_alpha)

    baker.generators.add('oscar.models.fields.UppercaseCharField', gen_upper)

    baker.generators.add('oscar.models.fields.ExtendedURLField',
                         default_mapping[URLField])

    baker.generators.add('oscar.models.fields.PositiveDecimalField',
                         gen_positive_decimal)

    baker.generators.add('django.db.models.SlugField',
                         gen_slug_starting_with_alpha)
```

And then the fun one is editing the Oscar models to set `null = False` and `blank = False` on model fields that Oscar treats as mandatory, but has no model definition to enforce that.  Some of these restrict functionality of the model, but it is usually good enough for building test cases.

```py
    require_model_field_value(Basket, 'owner')
    require_model_field_value(BasketLine, 'product')
    require_model_field_value(Benefit, 'range')
    require_model_field_value(Benefit, 'type')
    require_model_field_value(Benefit, 'value')
    require_model_field_value(Condition, 'range')
    require_model_field_value(Condition, 'type')
    require_model_field_value(Condition, 'value')
    require_model_field_value(ConditionalOffer, 'benefit')
    require_model_field_value(ConditionalOffer, 'condition')
    require_model_field_value(ConditionalOffer, 'slug')
    require_model_field_value(Email, 'user')
    require_model_field_value(LineAttribute, 'option')
    require_model_field_value(Notification, 'sender')
    require_model_field_value(Order, 'date_placed')
    require_model_field_value(Order, 'site')
    require_model_field_value(OrderDiscount, 'frequency')
    require_model_field_value(OrderDiscount, 'order')
[...]

PR #3427, opened 2020-06-26
PostActionMixin: Avoid fallback when no exists

jayvdb:
Fixes https://github.com/django-oscar/django-oscar/issues/3426

### DIFF
--- a/src/oscar/views/generic.py
+++ b/src/oscar/views/generic.py
@@ -27,7 +27,14 @@ class PostActionMixin(object):
                 return self.response
             else:
                 messages.error(request, _("Invalid form submission"))
-        return super().post(request, *args, **kwargs)
+                return self.get(request, *args, **kwargs)
+
+        # There may be no fallback implementation at super().post
+        try:
+            return super().post(request, *args, **kwargs)
+        except AttributeError:
+            messages.error(request, _("Invalid form submission"))
+            return self.get(request, *args, **kwargs)
 
 
 class BulkEditMixin(object):


src/oscar/views/generic.py
```diff
@@ -27,7 +28,13 @@ def post(self, request, *args, **kwargs):
                 return self.response
             else:
                 messages.error(request, _("Invalid form submission"))
-        return super().post(request, *args, **kwargs)
+        try:
+            f = super().post
+        except AttributeError:
+            raise SuspiciousOperation(
```
jayvdb, line 34: `messages.error(request, _("Invalid form submission"))` is used above.

happy to change it to that, but it seems `SuspiciousOperation` is actually more appropriate for both, and oscar currently never uses it.

solarissmoke: I prefer to reserve `SuspiciousOperation` for things that are very likely to be malicious, rather than things that can be accidental (e.g., a bad URL generated in a template), because it results in an empty 400 response to the client (and because it gets logged as a likely security issue, which I don't think this would be).

The docstring for the mixin says it is only usable with `DetailView` so I think we should just return `super().get()` if no action or post handler is found (with the "Invalid form submission" message added to the response).

src/oscar/views/generic.py
```diff
@@ -27,7 +27,13 @@ def post(self, request, *args, **kwargs):
                 return self.response
             else:
                 messages.error(request, _("Invalid form submission"))
-        return super().post(request, *args, **kwargs)
+        try:
+            f = super().post
+        except AttributeError:
+            messages.error(request, _("Invalid form submission"))
```
solarissmoke, line 33: It's possible to end up adding this message twice now (if `action` exists in the POST data but it's not a valid action, it will be added in the `else` block above, as well as here). That can be resolved by returning `get()` immediately inside that block.

jayvdb: Done

src/oscar/views/generic.py
```diff
@@ -27,7 +27,15 @@ def post(self, request, *args, **kwargs):
                 return self.response
             else:
                 messages.error(request, _("Invalid form submission"))
-        return super().post(request, *args, **kwargs)
+                return self.get()
```
solarissmoke, line 30: Need to pass `request, *args, **kwargs` here, otherwise this will fail :-). Probably worth adding a test to make sure this works.

src/oscar/views/generic.py
```diff
@@ -27,7 +27,15 @@ def post(self, request, *args, **kwargs):
...
+
+        try:
+            f = super().post
```
solarissmoke, line 33: We can try to directly `return` here rather than setting `f` since the except block is also returning.

jayvdb: That would potentially catch exception from inside the `post(..)`, and add an incorrect message, but done anyway as this is obscure, and doubtless is dev errors anyway not seen by users.

src/oscar/views/generic.py
jayvdb: There are no existing tests for any of this, so here isnt a good spot for me to break my teeth on helping write good *unit* test coverage.

I'd much rather be spending the time on https://github.com/django-oscar/django-oscar/issues/3420 , which would provide adequate test coverage for this specific case, and all of the other errors stemming from throwing ill-formed requests at oscar.

solarissmoke: OK - happy to merge this now. It's an edge case anyway, unlikely to affect a real user.

Issue #3429, opened 2020-06-27
'ProductCreateUpdateView' object has no attribute 'creating'


jayvdb:
### Issue Summary
`apps/dashboard/catalogue/views.py` `ProductCreateUpdateView` does not implement .delete(), and so the usual flow for DELETE would result in an HTTP response.

However `ProductCreateUpdateView` has its own `.dispatch` which calls `super().dispatch()`, and the response saved, but then it does some other work without determining whether that response should have been sent first.  If it was a http error code response, further work should not normally be done.

I've checked there are no other such unsafe behaviour around use of .dispatch in Oscar 2.0.

One approach would be to add status code checking before going into the extra logic.  Unfortunately the error responses and status codes are only clear wrt some codes like 400 & 403, and 500+, and less than 200.  Others depend on context.  Broadening to include all 300 and 400 codes could help avoid 'bugs' if the normal logic responds with a redirect, currently it could be 'overwritten' with the redirect in `check_objects_or_redirect`.  But it can also inhibit desirable behaviour.  The usual approach is something like
```py
if 200 <= response.status_code < 300:
    return self.check_objects_or_redirect() or resp
```

The other, lower impact fix is to prevent the `AttributeError` directly, so the extra work bails out early.

```py
======================================================================
ERROR: test_smoke_DELETE_dashboard/^catalogue/^products/create/(?P<product_class_slug>[\w-]+)/$ (django_smoke_tests.tests.SmokeTests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/lib/python3.8/site-packages/django_smoke_tests/generator.py", line 90, in test
    response = http_method_function(url, {})
  File "/usr/lib/python3.8/site-packages/django/test/client.py", line 565, in delete
    response = super().delete(path, data=data, content_type=content_type, secure=secure, **extra)
  File "/usr/lib/python3.8/site-packages/django/test/client.py", line 394, in delete
    return self.generic('DELETE', path, data, content_type,
  File "/usr/lib/python3.8/site-packages/django/test/client.py", line 421, in generic
    return self.request(**r)
  File "/usr/lib/python3.8/site-packages/django/test/client.py", line 496, in request
    raise exc_value
  File "/usr/lib/python3.8/site-packages/django/core/handlers/exception.py", line 34, in inner
    response = get_response(request)
  File "/usr/lib/python3.8/site-packages/django/core/handlers/base.py", line 115, in _get_response
[...]

PR #3430, opened 2020-06-27
ProductCreateUpdateView: Avoid AttributeError

jayvdb:
DELETE requests were failing with an exception instead
of the appropriate HTTP 405 response.

Fixes https://github.com/django-oscar/django-oscar/issues/3429


src/oscar/apps/dashboard/catalogue/views.py
```diff
@@ -226,6 +226,8 @@ def __init__(self, *args, **kwargs):
                          'image_formset': self.image_formset,
                          'recommended_formset': self.recommendations_formset,
                          'stockrecord_formset': self.stockrecord_formset}
+        self.creating = None
```
solarissmoke, line 229: I think it's better to define these as class variables, for consistency with what is done in the other mixins in this file, but also because it ensures deterministic overriding at instance level.

Issue #3431, opened 2020-06-27
DELETE on accounts/notifications/update causes TypeError


jayvdb:
### Issue Summary

A DELETE on the UpdateView fails, at least on Django 3.  It uses the BulkEditMixin, which I guess is expecting the `delete` to come through a POST, which will then invoke `UpdateView.delete`.

As .dispatch() isnt implemented, the DELETE invokes the `.delete` directly, but it requires arg `notifications` which dispatch doesnt provide.

Probably `delete` action and method here should be renamed to avoid the conflict with the default dispatch mechanism.  More broadly, the oscar generic PostActionMixin has a `do_foo` naming convention - perhaps that might be appropriate here.  

```py
======================================================================
ERROR: test_smoke_DELETE_oscar/^accounts/^notifications/update/$ (django_smoke_tests.tests.SmokeTests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/lib/python3.8/site-packages/django_smoke_tests/generator.py", line 90, in test
    response = http_method_function(url, {})
  File "/usr/lib/python3.8/site-packages/django/test/client.py", line 565, in delete
    response = super().delete(path, data=data, content_type=content_type, secure=secure, **extra)
  File "/usr/lib/python3.8/site-packages/django/test/client.py", line 394, in delete
    return self.generic('DELETE', path, data, content_type,
  File "/usr/lib/python3.8/site-packages/django/test/client.py", line 421, in generic
    return self.request(**r)
  File "/usr/lib/python3.8/site-packages/django/test/client.py", line 496, in request
    raise exc_value
  File "/usr/lib/python3.8/site-packages/django/core/handlers/exception.py", line 34, in inner
    response = get_response(request)
  File "/usr/lib/python3.8/site-packages/django/core/handlers/base.py", line 115, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "/usr/lib/python3.8/site-packages/django/core/handlers/base.py", line 113, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/lib/python3.8/site-packages/django/contrib/auth/decorators.py", line 21, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "/usr/lib/python3.8/site-packages/django/views/generic/base.py", line 71, in view
    return self.dispatch(request, *args, **kwargs)
  File "/usr/lib/python3.8/site-packages/django/views/generic/base.py", line 97, in dispatch
[...]

Issue #3428, opened 2020-06-26
/oscar/checkout/thank-you/` returns 404 when checkout_order_id not in session


jayvdb:
### Issue Summary

Oscar 2.0 smoke tests, `GET /oscar/checkout/thank-you/` returns 404.

The smoke test is obviously requesting this without going through the checkout process, so 403 could be appropriate.  It is a very long stretch of the meaning of 404 to use it here - the *resource* was found - it was the resource is responding that it doesnt like the *request* context.

`ThankYouView` has `raise http.Http404(_("No order found"))`

More interesting, it only does this if a session field `checkout_order_id` wasnt found, which that error message does not convey accurately.

If it is about a missing field, it could be a `SuspiciousOperation`.

However as it is a GET, that page URL could be in a users browser history, in which case it could be inadvertently loaded in a new session, which isnt really suspicious, and the user shouldnt see a http error page.

It should be possible to check if there is a successful order by the user, and thank them again! ;-)
Failing that, a "there is no checkout in progress" error message?

### Steps to Reproduce

c.f. setup in https://github.com/django-oscar/django-oscar/issues/3421

### Technical details

* Python version: 3.8
* Django version: 3.0
* Oscar version: 2.0

jayvdb:
A similar issue I'll dump here instead of creating lots of issues.

410 Gone is described in Wikipedia as

>  The 410 (Gone) status code indicates that access to the target resource is no longer available at the origin server and that this condition is likely to be permanent.  If the origin server does not know, or has no facility to determine, whether or not the condition is permanent, the status code 404 (Not Found) ought to be used instead.

> The 410 response is primarily intended to assist the task of web maintenance by notifying the recipient that the resource is intentionally unavailable and that the server owners desire that remote links to that resource be removed.  Such an event is common for limited-time, promotional services and for resources belonging to individuals no longer associated with the origin server's site.  It is not necessary to mark all permanently unavailable resources as "gone" or to keep the mark for any length of time -- that is left to the discretion of the server owner.

> A 410 response is cacheable by default; i.e., unless otherwise indicated by the method definition or explicit cache controls (see Section 4.2.2 of [RFC7234]).

That is clearly inappropriate here, and even 404 is inappropriate here.  Again, 403 or 400 would be more appropriate.

```py
FAIL: test_smoke_GET_oscar/^accounts/^notifications/update/$ (django_smoke_tests.tests.SmokeTests)
----------------------------------------------------------------------
[...]

PR #3437, opened 2020-07-04
Prevent notifications DELETE

jayvdb:
HTTP method DELETE causes a TypeError, as the delete operation
must be done via POST.

Fixes https://github.com/django-oscar/django-oscar/issues/3431

### DIFF
--- a/src/oscar/apps/communication/notifications/views.py
+++ b/src/oscar/apps/communication/notifications/views.py
@@ -69,11 +69,10 @@ class DetailView(PageTitleMixin, generic.DetailView):
             recipient=self.request.user)
 
 
-class UpdateView(BulkEditMixin, generic.RedirectView):
+class UpdateView(BulkEditMixin, generic.View):
     model = Notification
     actions = ('archive', 'delete')
     checkbox_object_name = 'notification'
-    permanent = False
 
     def get_object_dict(self, ids):
         return self.model.objects.filter(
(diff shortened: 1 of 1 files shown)


jayvdb:
Note there is also a smoke test error wrt GET described at https://github.com/django-oscar/django-oscar/issues/3428#issuecomment-650498588

Happy to include the fix here, with a bit of guidance on what type of fix would be acceptable.

jayvdb:
CI fails with `isort: error: unrecognized arguments: --recursive`

solarissmoke:
> CI fails with `isort: error: unrecognized arguments: --recursive`

Fixed in 3b7a1bc (you'll need to rebase this PR onto master).

> Note there is also a smoke test error wrt GET described

That is coming from the parent `RedirectMixin`, which actually isn't being used here at all (we never define a redirect URL), so it seems to have been a mistake to subclass it. We can just remove that mixin. The implication of that is that only `POST` is allowed on this view.

src/oscar/apps/communication/notifications/views.py
```diff
@@ -69,8 +69,9 @@ def get_queryset(self):
             recipient=self.request.user)
 
 
-class UpdateView(BulkEditMixin, generic.RedirectView):
+class UpdateView(BulkEditMixin, generic.View):
     model = Notification
+    http_method_names = ['post']
     actions = ('archive', 'delete')
     checkbox_object_name = 'notification'
     permanent = False
```
solarissmoke, line 77: Please remove this line (`permanent = False`) - it's also a hangover from the redirect view.

PR #3474, opened 2020-08-07
Making categories mysql compatible

iamsobanjaved:
Trying to make categories compatible for mysql as it is currently broken

### DIFF
--- a/src/oscar/apps/catalogue/abstract_models.py
+++ b/src/oscar/apps/catalogue/abstract_models.py
@@ -227,7 +227,7 @@ class AbstractCategory(MP_Node):
         else:
             included_in_non_public_subtree = self.__class__.objects.filter(
                 is_public=False, path__rstartswith=OuterRef("path"), depth__lt=OuterRef("depth")
-            ).exists()
+            )
             self.get_descendants_and_self().update(
                 ancestors_are_public=Exists(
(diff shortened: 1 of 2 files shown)


iamsobanjaved:
@solarissmoke @jaroel i have translated query into loop but it isn't working

jaroel:
Mysql isn't supported: #2630 
Besides that, I'm not interested in MySQL, nor do I currently use this project.
Use Postgresql and be done.

iamsobanjaved:
@solarissmoke i have converted query to loop for mysql but it isnt working?

solarissmoke:
@iamsobanjaved I'm going to close this for now, for the following reasons:

1. It's not clear what the issue is that you are attempting to fix.

2. It seems your proposed changes don't work, but you're not clear why. Until we know what the issue is its hard to offer suggestions about what isn't working.

Please feel free to reopen if you are able to provide a bit more context. As regards MySQL I would note that Oscar doesn't officially support it at present. If someone is able to spend some time working on addressing the MySQL compatibility issues then we're happy to support that effort - but unfortunately I don't have the time/capacity to debug and fix the issues myself.

awais786:
@solarissmoke the issue is latest django-oscars breaks on mysql. This issue is already reported but unfortunately django-oscar dropped support.

PR #3481, opened 2020-08-18
Add shipping charge to the percentage surcharge

viggo-devries:
Give the shipping charges as kwarg to the applicator and use it to calculate the total surcharge.

This is necessary because before the surcharges were only calculated over the basket items.
For example when using this as a payment surcharge you actually pay the shipping charge but don't pay surcharges over them. This way it will cost the webshop owner some money in transaction cost.

### DIFF
--- a/src/oscar/apps/checkout/surcharges.py
+++ b/src/oscar/apps/checkout/surcharges.py
@@ -27,11 +27,20 @@ class PercentageCharge(BaseSurcharge):
         self.percentage = percentage
 
     def calculate(self, basket, **kwargs):
-        if basket.total_excl_tax:
+        if not basket.is_empty:
+            shipping_charge = kwargs.get("shipping_charge")
+
+            if shipping_charge is not None:
+                total_excl_tax = basket.total_excl_tax + shipping_charge.excl_tax
+                total_incl_tax = basket.total_incl_tax + shipping_charge.incl_tax
+            else:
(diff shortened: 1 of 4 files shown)


viggo-devries:
Give the shipping charges as kwarg to the applicator and use it to calculate the total surcharge

src/oscar/apps/checkout/surcharges.py
```diff
@@ -27,11 +27,20 @@ def __init__(self, percentage):
         self.percentage = percentage
 
     def calculate(self, basket, **kwargs):
-        if basket.total_excl_tax:
+        if not basket.is_empty:
+            shipping_charge = kwargs.get("shipping_charge", None)
```
specialunderwear, line 31: A default of None is obsolete, because that is the default allready.

specialunderwear (changes requested):

src/oscar/apps/checkout/surcharges.py
viggo-devries: Fixed

specialunderwear (approved):

specialunderwear:
I don't think it should be a setting because the surcharges shipped in oscar serve more as an example. People can always implement their own surcharges if different requirements must be met. So this seems ok to me. Calculating surcharges over the total amount payable is sane.

specialunderwear (approved):

PR #3482, opened 2020-08-19
Fix table column misalignment in dashboard order detail page

malefice:
PR #3019 removed some fields, but one of the tables in the dashboard order detail page was not adjusted accordingly when the "Estimated dispatch date" column was removed. This issue can be replicated using the latest sandbox docker image.

image

The `colspan` attribute must be changed so that the column headers and the right edge will be aligned like so:

image


solarissmoke (approved):
Thanks!

PR #3483, opened 2020-08-24
Replace uses of templates/base.html with templates/oscar/base.html

### LINKED ISSUE #3471: docs incorrects for 'how to customize templates' section
Hi there!

There is an error in the docs, specifically in the [how to customize templates section].(https://django-oscar.readthedocs.io/en/2.1.0/howto/how_to_customise_templates.html)

According to issue #1378, the _base.html_ mentioned in the examples should be in '_templates/oscar/base.html_' instead of '_templates/base.html_'.

I'm new to django oscar, and I spent a couple of hours before I could realize what were wrong, so I think it would be very helpful to new devs :)

Regards!

adityasharma3:
Replaced the uses of templates/base.html with templates/oscar/base.html . Fixes #3471.


solarissmoke (approved):
Thanks :).

PR #3484, opened 2020-08-26
Use QuerySet.first() where appropriate instead of fetching by index.

solarissmoke:
Cleanup of some old code that was written before `first()` was introduced to Django's queryset methods. I haven't bothered changing some tests that also do this - just want to reduce clutter in the functional code.

### DIFF
--- a/src/oscar/apps/dashboard/vouchers/views.py
+++ b/src/oscar/apps/dashboard/vouchers/views.py
@@ -171,7 +171,7 @@ class VoucherUpdateView(generic.FormView):
 
     def get_initial(self):
         voucher = self.get_voucher()
-        offer = voucher.offers.all()[0]
+        offer = voucher.offers.first()
         benefit = offer.benefit
         return {
             'name': voucher.name,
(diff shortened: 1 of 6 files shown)


src/oscar/apps/dashboard/vouchers/views.py
```diff
@@ -171,7 +171,7 @@ def get_form_kwargs(self):
 
     def get_initial(self):
         voucher = self.get_voucher()
-        offer = voucher.offers.all()[0]
+        offer = voucher.offers.first()
```
specialunderwear, line 174: When no offers exist this now causes an AttrbuteError which used to be an IndexError. Please confirm this is harmless.

src/oscar/apps/dashboard/vouchers/views.py
```diff
@@ -195,7 +195,7 @@ def form_valid(self, form):
         voucher.end_datetime = form.cleaned_data['end_datetime']
         voucher.save()
 
-        offer = voucher.offers.all()[0]
+        offer = voucher.offers.first()
```
specialunderwear, line 198: same here

src/oscar/apps/dashboard/vouchers/views.py
solarissmoke: The dashboard `VoucherUpdateView` assumes there is always one offer associated with a voucher (which is the only way you can create one via the dashboard). The behaviour is unchanged in that both the old `IndexError` and the new `AttributeError` would be unhandled if you try to edit a voucher whose offer has somehow been deleted. TLDR: yeah it's fine.

specialunderwear (approved):

PR #3485, opened 2020-08-28
Add support for restricted offer combinations

### LINKED ISSUE #2652: Offer Tracking for Stacked Offers

solarissmoke:
Rebase of #2690.

Fixes #2652.


src/oscar/apps/offer/abstract_models.py
```diff
@@ -443,6 +451,13 @@ def products(self):
         return queryset.filter(is_discountable=True).exclude(
             structure=Product.CHILD)
 
+    @cached_property
+    def combined_offers(self):
```
specialunderwear, line 455: I think this can just be done in a single query.

Something like:
```
ConditionalOffer.filter(
    models.Q(id__in=self.pk)
    |models.Q(id__in=self.combinations.values_list("id", flat=True)
    |models.Q(id__in=self.in_combination.values_list("id", flat=True)
)
```
that way you can do more filtering on the resulting queryset.

specialunderwear (changes requested):
I see you've changed the name mangled properties ``__magic_double_underscore`` methods, which I encourage! Is there a reason for leaving some of these methods still with the double underscore?

specialunderwear (dismissed):

PR #3489, opened 2020-09-02
Make category slugs dashboard-editable.

### LINKED ISSUE #561: Allow a category slug to be edited

### LINKED ISSUE #3472: Category slug doesn't get updated
### Issue Summary

When changing the title of a category in the dashboard, the slug doesn't get regenerated.

### Technical details

* Python 3.6.10
* Django version: 2.2.15
* Oscar version: 2.1

solarissmoke:
Make category slugs editable from the dashboard, so that admins can modify them if they change a category name etc.

Fixes #561, fixes #3472.

### DIFF
--- a/src/oscar/apps/dashboard/catalogue/forms.py
+++ b/src/oscar/apps/dashboard/catalogue/forms.py
@@ -24,10 +24,20 @@ ProductSelect = get_class('dashboard.catalogue.widgets', 'ProductSelect')
                                                   ('RelatedFieldWidgetWrapper',
                                                    'RelatedMultipleFieldWidgetWrapper'))
 
-CategoryForm = movenodeform_factory(
+
+BaseCategoryForm = movenodeform_factory(
     Category,
-    fields=['name', 'description', 'image', 'is_public'],
+    fields=['name', 'slug', 'description', 'image', 'is_public'],
     exclude=['ancestors_are_public'])
+
+
+class CategoryForm(BaseCategoryForm):
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        if 'slug' in self.fields:
+            self.fields['slug'].required = False
+            self.fields['slug'].help_text = _('Leave blank to generate from category name')
 
 
 class ProductClassSelectForm(forms.Form):
--- a/docs/source/releases/v3.0.rst
+++ b/docs/source/releases/v3.0.rst
@@ -67,6 +67,8 @@ Minor changes
 - Several models were updated to define a default ordering, to avoid issues with inconsistent ordering of
   items in the dashboard and elsewhere. Database migrations are required for these changes.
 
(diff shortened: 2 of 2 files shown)


src/oscar/apps/dashboard/catalogue/forms.py
```diff
@@ -24,12 +24,22 @@
                                                   ('RelatedFieldWidgetWrapper',
                                                    'RelatedMultipleFieldWidgetWrapper'))
 
-CategoryForm = movenodeform_factory(
+
+BaseCategoryForm = movenodeform_factory(
```
specialunderwear, line 28: From the documentation at https://django-treebeard.readthedocs.io/en/latest/forms.html#treebeard.forms.movenodeform_factory
```
MyNodeForm = movenodeform_factory(MyNode)
```
is equivalent to
```
class MyNodeForm(MoveNodeForm):
    class Meta:
        model = models.MyNode
        exclude = ('sib_order', 'parent')
```

Wouldn't it be mode clear to define CategoryForm by subclassing MoveNodeForm instead of generating it with the factory?

src/oscar/apps/dashboard/catalogue/forms.py
```diff
@@ -24,12 +24,22 @@
...
     exclude=['ancestors_are_public'])
 
 
+class CategoryForm(BaseCategoryForm):
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        if 'slug' in self.fields:
+            self.fields['slug'].required = False
+            self.fields['slug'].help_text = _('Leave blank to generate from category name')
```
specialunderwear, line 40: Is this string allready present in the gettext catalog?

specialunderwear (changes requested):

src/oscar/apps/dashboard/catalogue/forms.py
solarissmoke: No - but I've been periodically updating the gettext catalog separately from functional changes, because it introduces a huge diff that makes review harder.

auvipy (approved):

src/oscar/apps/dashboard/catalogue/forms.py
solarissmoke: There's a warning right above the section in the treebeard documentation you link to: 

> Subclassing `MoveNodeForm` directly is discouraged, since special care is needed to handle excluded fields, and these change depending on the tree type. It is recommended that the `movenodeform_factory()` function is used instead.

Obviously we only use one tree type in Oscar but I think relying on the factory is more resilient to any future upstream changes?

specialunderwear (approved):

PR #3497, opened 2020-09-10
Make ProductAttributesContainer a lazy object so that initialisation is simpler.

### LINKED ISSUE #3258: Writing to products attributes with Product.attr can lead to dataloss
### Issue Summary

Writing to products attributes with `product.attr` can lead to data loss

### Steps to Reproduce

Create a productclass with 3 attributes, the second one you create must be optional.
Create a product and do not fill the second optional attribute. Now write to the non-optional attributes with code like this:

        p = Product.objects.get(pk=4)
        p.attr.subtitle = "lama"
        p.attr.available = False
        p.attr.save()

One of the 2 attributes will not be updated because the attr object will not be intialized. During save it will encounter the second attribute which is not on the object and trigger initialization. Initialization will overwrite one of the 2 values written and reset it to whatever is in the database.

### Technical details

This problem occurs with all python versions and all oscar versions.

solarissmoke:
Fixes #3258.

### DIFF
--- a/src/oscar/apps/catalogue/product_attributes.py
+++ b/src/oscar/apps/catalogue/product_attributes.py
@@ -2,7 +2,7 @@ from django.core.exceptions import ValidationError
 from django.utils.translation import gettext_lazy as _
 
 
-class ProductAttributesContainer(object):
+class ProductAttributesContainer:
     """
     Stolen liberally from django-eav, but simplified to be product-specific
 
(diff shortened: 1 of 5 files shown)


src/oscar/apps/catalogue/product_attributes.py
```diff
@@ -13,25 +13,30 @@ class ProductAttributesContainer(object):
...
 
     def __init__(self, product):
-        self.product = product
-        self.initialised = False
+        self._product = product
+        self._initialised = False
 
     def initiate_attributes(self):
         values = self.get_values().select_related('attribute')
+        self._initialised = True
```
solarissmoke, line 24: You get an infinite loop if this is set *after* the loop below, because that itself is using `setattr`...

auvipy (approved):

solarissmoke:
@auvipy request that you please refrain from approving PRs unnecessarily - if you have feedback on the changes then please feel free to comment, otherwise there's no need to leave a review (it just creates noise for people who receive notifications), as we still need someone from the core team to approve. Thanks.

specialunderwear (approved):
Nice work!

PR #3500, opened 2020-09-21
feature/docs

### LINKED ISSUE #3038: Disable Debug Toolbar in Sandbox
I just tested Oscar for the first time. I created the sandbox locally and was about to conclude that it was too slow for me. I looked at the debug toolbar and realized it could not be the SQL (which is nice and fast). Actually it was the toolbar itself. I deactivated debug mode and realized that Oscar is fast. 

Hence my suggestion: disable the debug toolbar for the sandbox or document that it should be disabled if you want to test Oscar for speed.

0.02

sasha0:
Fixes #3038

### DIFF
--- a/docs/source/internals/sandbox.rst
+++ b/docs/source/internals/sandbox.rst
@@ -88,6 +88,13 @@ at: http://localhost:8000.  A sample superuser is installed with credentials::
     password: testing
 
 
+.. warning::
+
+    The sandbox has Django Debug Toolbar enabled by default, which will affect
+    its performance. You can disable it by setting ``INTERNAL_IPS`` to an
+    empty list in your local settings.
+
+
 Run the sandbox using Docker
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 


docs/source/internals/sandbox.rst
```diff
@@ -88,6 +88,15 @@ at: http://localhost:8000.  A sample superuser is installed with credentials::
     password: testing
 
 
+.. warning::
+
+    Sandbox site has Django Debug Toolbar integrated by default, which might
```
solarissmoke, line 93: Could you please adjust the grammar as follows? I also propose to disable by overriding `INTERNAL_IPS` which is simpler/quicker that removing it from middleware/installed apps:

> The sandbox has has Django Debug Toolbar enabled by default, which will affect its performance. You can disable it by setting \``INTERNAL_IPS\`` to an empty list in your local settings.

solarissmoke (changes requested):

docs/source/internals/sandbox.rst
sasha0: Ok done.

solarissmoke (approved):

PR #3501, opened 2020-09-21
Issues/2772/reports

### LINKED ISSUE #2772: "Product analytics" report on dashboard ignores date range

sasha0:
Fixes #2772

### DIFF
--- a/src/oscar/apps/dashboard/reports/reports.py
+++ b/src/oscar/apps/dashboard/reports/reports.py
@@ -28,8 +28,7 @@ class ReportGenerator(object):
 
         formatter_name = '%s_formatter' % kwargs.get('formatter', 'HTML')
         self.formatter = self.formatters[formatter_name]()
-        if self.queryset is None:
-            self.queryset = self.get_queryset()
+        self.queryset = self.get_queryset()
 
     def report_description(self):
         return _('%(report_filter)s between %(start_date)s and %(end_date)s') \
@@ -39,6 +38,9 @@ class ReportGenerator(object):
                }
 
     def get_queryset(self):
+        if self.queryset is not None:
+            return self.queryset
+
         if not self.model_class:
             raise ValueError(
                 "Please define a model_class property on your report generator class, "
--- a/src/oscar/apps/dashboard/reports/reports.py
+++ b/src/oscar/apps/dashboard/reports/reports.py
@@ -19,6 +19,8 @@ class ReportGenerator(object):
     code = ''
     description = '<insert report description>'
     date_range_field_name = None
+    model_class = None
+    queryset = None
 
     def __init__(self, **kwargs):
         self.start_date = kwargs.get('start_date')
@@ -26,6 +28,8 @@ class ReportGenerator(object):
 
         formatter_name = '%s_formatter' % kwargs['formatter']
         self.formatter = self.formatters[formatter_name]()
+        if self.queryset is None:
+            self.queryset = self.get_queryset()
 
     def report_description(self):
         return _('%(report_filter)s between %(start_date)s and %(end_date)s') \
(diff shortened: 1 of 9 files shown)


src/oscar/apps/dashboard/reports/reports.py
```diff
@@ -19,13 +19,18 @@ class ReportGenerator(object):
...
+    queryset = None
 
     def __init__(self, **kwargs):
         self.start_date = kwargs.get('start_date')
         self.end_date = kwargs.get('end_date')
 
-        formatter_name = '%s_formatter' % kwargs['formatter']
+        formatter_name = '%s_formatter' % kwargs.get('formatter', 'HTML')
         self.formatter = self.formatters[formatter_name]()
+        if self.queryset is None:
```
solarissmoke, line 31: I think this check is better done inside the `get_queryset` method itself, so that projects can override behaviour more easily. See [here](https://github.com/django/django/blob/master/django/views/generic/detail.py#L58-L76) for the pattern Django uses.

solarissmoke (changes requested):
We'll need an entry in the 3.0 release notes for this to describe the API change.

src/oscar/apps/dashboard/reports/reports.py
sasha0: If we do this change, we'd need to add `queryset = self.get_queryset()` in all `generate` method calls, instead of just refer to `self.queryset`, right?

sasha0:
> We'll need an entry in the 3.0 release notes for this to describe the API change.

Ok done.

src/oscar/apps/dashboard/reports/reports.py
solarissmoke: What I had in mind was:

`self.queryset = self.get_queryset()` here, and in `get_queryset()` we do:

```
if self.queryset is not None:
    return self.queryset
```

So that we ensure that `get_queryset()` is *always* called, and if a project wants to override (e.g., to add some filtering or select_relateds) then it can do so without worrying about whether `queryset` has been set on the class. What do you think?

sasha0: Yeah, agreed.

sasha0: Ok done.

solarissmoke:
@sasha0 tests are failing on this - grateful if you can fix.

sasha0:
> @sasha0 tests are failing on this - grateful if you can fix.

@solarissmoke done.

src/oscar/apps/dashboard/reports/reports.py
```diff
@@ -34,8 +38,18 @@ def report_description(self):
                'end_date': date(self.end_date, 'DATE_FORMAT')
                }
 
+    def get_queryset(self):
+        if self.queryset is not None:
+            return self.queryset
+
+        if not self.model_class:
+            raise ValueError(
+                _("Please set model_class property for your report generator class or override get_queryset method.")
```
solarissmoke, line 47: Generally we don't translate developer error notices in Oscar (see other places where we use ValueError) - I think this makes sense as it's not an error that a site user should ever see, and adds more translation work.

solarissmoke: Also please adjust the grammar to "Please define a model_class property on your report generator class, or override the qet_queryset() method".

solarissmoke (changes requested):

src/oscar/apps/dashboard/reports/reports.py
sasha0: Ok done.

Issue #2972, opened 2019-03-25
New Oscar architecture proposal


sasha0:
A while ago we discussed this with the @django-oscar/core team and finally I got to write it down as a proposal.

Since more and more projects implement only API and do not require frontend and/or dashboard, I came up with the raw idea of splitting Oscar into the parts, which would ease it customization:

- `oscar-core` - the main core functionality;
- `oscar-frontend` - basically all storefront views and templates;
- `oscar-dashboard` - all dashboard functionality.

On my latest projects we don't use storefront or dashboard at all (basically only core functionality and API), so in this case it won't need to disable almost all views.

Any feedback appreciated.

maikhoepfel:
David and I always envisioned a split like that. Have a core with a API, and ship the dashboard and front end talking to that API. It would also make sure no business logic lands in the views, which is desirable. 
The front end and dashboard should be forbidden from making database calls and should only communicate via the API. But one could skip serialization - I don’t there’s a need to call the API over REST, I’d be fine with the API returning regular Django model instances.

maerteijn:
I would be a proponent to split-up Oscar in different packages as @sasha0 suggested. From there you can go ahead and build a programmers API when there is a need for it.

Starting with separate packages / modules so it's easy to disable the store front end or dashboard (or not to install them at all) would really make sense, next to the fact that the maintenance and development of the codebase of the different components would be much easier.

maikhoepfel:
What do you see as benefit of having seperate packages, @*maerteijn <https://github.com/maerteijn>*? I absolutely understand your point of wanting to disable them, but I think there's a hidden cost to splitting it up into packages - if you're referring to PyPi packages and seperate GitHub repos, not just Python packages.

If they were to ship as dedicated apps (the dashboard already is), one could easily disable it by not including it in INSTALLED_APPS. The frontend would require splitting up the current apps (this needs to be done either way), but I'm not sure there's a benefit to going for seperate packages. I'd be afraid that it makes maintenance and develeopment trickier, and it would also make releases trickier (because you need to make sure people install the correct versions of frontend - dashboard - core).

sasha0:
> The frontend would require splitting up the current apps (this needs to be done either way), but I'm not sure there's a benefit to going for seperate packages.

[...]

Issue #1468, opened 2014-08-26
Untangle CheckoutSessionMixin


maiksprenger:
I recently had to work on a checkout REST API based on Oscar, and had to get the shipping address and method stored in the session. The mixin currently carries the request and session data as state around and is initialised in `dispatch`. That makes sense when it's used in views, but there's a lot of methods in there that could be helpful outside of that context. 

I'm not entirely sure what the right approach is. Maybe limiting the scope of the mixin to handle the  validating the pre and skip conditions makes sense. All other logic could live in individual methods.

sasha0:
Yes, it does make sense to decouple `CheckoutSessionMixin` into two classes:

- session validation mixin, with all methods which require request parameter and with the initiation in dispatch method;
- order submission data mixin.

Backwards compatibility would be supported by inheriting both mixin classes under the old name, but will make the whole functionality more re-usable.

sasha0:
Made attempt to decouple, hence reviewed mixin again and now not sure if it does make sense, since checkout methods are interrelated, e.g.:

`check_a_valid_shipping_address_is_captured` calls `get_shipping_address`, `skip_unless_payment_is_required` calls `get_shipping_address` and `get_order_totals`.

PR #3503, opened 2020-09-22
Issues/2972/decouple_mixins

sasha0:
Refs #2972 #1468

First attempt to decouple `CheckoutSessionMixin` into multiple mixin classes and try to make it more modular.

### DIFF
--- a/src/oscar/apps/checkout/mixins.py
+++ b/src/oscar/apps/checkout/mixins.py
@@ -129,11 +129,6 @@ class OrderPlacementMixin(CheckoutSessionMixin):
         else:
             status = kwargs.pop('status')
 
-        if 'request' not in kwargs:
-            request = getattr(self, 'request', None)
-        else:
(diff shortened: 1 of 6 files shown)


src/oscar/apps/checkout/mixins.py
```diff
@@ -144,7 +139,6 @@ def place_order(self, order_number, user, basket, shipping_address,
             total=order_total,
             billing_address=billing_address,
             status=status,
-            request=request,
```
specialunderwear, line 147: By not passing the request the order will not be properly associated with the site it originated from.

specialunderwear (changes requested):
I think the removal of request from some of the classes changed in this pull request will definately break multisite oscar instances and make it much harder to implement them

specialunderwear:
Is this now ready for review and complete?

sasha0:
> Is this now ready for review and complete?

@specialunderwear yep.

PR #3504, opened 2020-09-22
issues/3435/basket_form

### LINKED ISSUE #3435: Overriding AddToBasketForm is not Simple

sasha0:
Fixes #3435 , closes #3436.

### DIFF
--- a/src/oscar/apps/basket/forms.py
+++ b/src/oscar/apps/basket/forms.py
@@ -299,7 +299,15 @@ class AddToBasketForm(forms.Form):
         return options
 
 
-class SimpleAddToBasketForm(AddToBasketForm):
+class SimpleAddToBasketMixin:
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        if 'quantity' in self.fields:
+            self.fields['quantity'].initial = 1
+            self.fields['quantity'].widget = forms.HiddenInput()
+
+
+class SimpleAddToBasketForm(SimpleAddToBasketMixin, AddToBasketForm):
     """
     Simplified version of the add to basket form where the quantity is
     defaulted to 1 and rendered in a hidden widget
@@ -307,9 +315,3 @@ class SimpleAddToBasketForm(AddToBasketForm):
     Most of the time, you won't need to override this class. Just change
     AddToBasketForm to change behaviour in both forms at once.
     """
-
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
-        if 'quantity' in self.fields:
-            self.fields['quantity'].initial = 1
-            self.fields['quantity'].widget = forms.HiddenInput()
--- a/src/oscar/apps/basket/forms.py
+++ b/src/oscar/apps/basket/forms.py
@@ -312,6 +312,9 @@ class SimpleAddToBasketForm(SimpleAddToBasketMixin, AddToBasketForm):
     Simplified version of the add to basket form where the quantity is
     defaulted to 1 and rendered in a hidden widget
 
-    Most of the time, you won't need to override this class. Just change
-    AddToBasketForm to change behaviour in both forms at once.
+    If you changed `AddToBasketForm`, you'll need to override this class
+    as well by doing:
+
+    class SimpleAddToBasketForm(SimpleAddToBasketMixin, AddToBasketForm):
+        pass
     """


src/oscar/apps/basket/forms.py
```diff
@@ -299,17 +299,19 @@ def cleaned_options(self):
...
+            self.fields['quantity'].initial = 1
+            self.fields['quantity'].widget = forms.HiddenInput()
+
+
+class SimpleAddToBasketForm(SimpleAddToBasketMixin, AddToBasketForm):
     """
     Simplified version of the add to basket form where the quantity is
     defaulted to 1 and rendered in a hidden widget
 
     Most of the time, you won't need to override this class. Just change
```
solarissmoke, line 315: This docstring is not correct - simply changing `AddToBasketForm` does not work and never has. I think we need to replace this to explain how to override this - i.e., to explain that if you have changed `AddToBasketForm` then you need to redefine `SimpleAddToBasketForm` as well, by doing:

```
class SimpleAddToBasketForm(SimpleAddToBasketMixin, AddToBasketForm):
    pass
```

solarissmoke (changes requested):
I think

src/oscar/apps/basket/forms.py
sasha0: Ok done.

src/oscar/apps/basket/forms.py
```diff
@@ -299,17 +299,22 @@ def cleaned_options(self):
...
+    """
+    Simplified version of the add to basket form where the quantity is
+    defaulted to 1 and rendered in a hidden widget
+
+    If you changed `AddToBasketForm`, you'll need to override this class
+    as well by doing:
+
+    class SimpleAddToBasketForm(SimpleAddToBasketMixin, AddToBasketForm):
+        pass
+    """
```
solarissmoke, line 320: Do we need a `pass` statement here? I'm surprised it doesn't complain without one since there is no other code in the class...

sasha0: Yes, I believe we need it - https://docs.python.org/3/tutorial/classes.html#odds-and-ends.  

I'm surprised too - PyCharm didn't show any syntax violations.

solarissmoke: OK - please add the `pass` statement.

solarissmoke (changes requested):

solarissmoke (approved):

PR #3505, opened 2020-09-23
Issues/3425/invalid page

### LINKED ISSUE #3425: InvalidPage exceptions in CatalogueView are not properly handled

### Issue Summary

Oscar 2.0 (sorry if already fixed..)

If the customer is on page 2 of the Shop product list, and some products are deleted such that there is now only one page, then the customer clicks 'add to cart', they are greeted with a 404 on `
/catalogue/?page=2`

### Steps to Reproduce

1. Create 24 products, and set products to page to 20
2. Navigate to page 2 of the shop
3. Separate tab, delete 20 products but not the one about to be added to cart
4. Back on the shop, click 'add to cart.'

### Technical details

* Python version: 3.8
* Django version: 3.0
* Oscar version: 2.0

sasha0:
Fixes #3425

### DIFF
--- a/src/oscar/apps/search/search_handlers.py
+++ b/src/oscar/apps/search/search_handlers.py
@@ -9,7 +9,43 @@ from . import facets
 FacetMunger = get_class('search.facets', 'FacetMunger')
 
 
-class SearchHandler(object):
+class SearchResultsPaginationMixin(object):
+    paginate_by = None
+    paginator_class = Paginator
+    page_kwarg = 'page'
+
+    def get_paginate_by(self, *args, **kwargs):
+        return self.paginate_by
+
+    def paginate_queryset(self, queryset, page_size):
+        """
+        Paginate the search results. This is a simplified version of
+        Django's MultipleObjectMixin.paginate_queryset
+        """
+        paginator = self.get_paginator(queryset, page_size)
+        page_kwarg = self.page_kwarg
+        page_number = self.request_data.get(page_kwarg, 1)
+        try:
+            page_number = int(page_number)
+        except ValueError:
+            if page_number == 'last':
+                page_number = paginator.num_pages
+            else:
+                raise InvalidPage(_(
+                    "Page is not 'last', nor can it be converted to an int."))
+        # This can also raise an InvalidPage exception.
+        page = paginator.page(page_number)
+        return paginator, page, page.object_list, page.has_other_pages()
+
+    def get_paginator(self, queryset, per_page=None):
+        """
+        Return a paginator. Override this to set settings like orphans,
+        allow_empty, etc.
+        """
+        return self.paginator_class(queryset, per_page)
+
+
+class SearchHandler(SearchResultsPaginationMixin):
     """
     A class that is concerned with performing a search and paginating the
     results. The search is triggered upon initialisation (mainly to have a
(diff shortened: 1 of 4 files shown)


src/oscar/apps/search/search_handlers.py
```diff
@@ -9,7 +9,43 @@
 FacetMunger = get_class('search.facets', 'FacetMunger')
 
 
-class SearchHandler(object):
+class SearchResultsPaginationMixin(object):
```
solarissmoke, line 12: Please remove the `object` arg, not needed in Python 3.

src/oscar/apps/search/search_handlers.py
```diff
@@ -9,7 +9,43 @@
...
+    paginate_by = None
+    paginator_class = Paginator
+    page_kwarg = 'page'
+
+    def get_paginate_by(self, *args, **kwargs):
```
solarissmoke, line 17: This method is not passed any arguments from where we are calling it?

solarissmoke: Do we need this method at all? If someone wanted to override with custom logic they could just define a `paginate_by` property.

src/oscar/apps/search/search_handlers.py
```diff
@@ -9,7 +9,43 @@
...
+        paginator = self.get_paginator(queryset, page_size)
+        page_kwarg = self.page_kwarg
+        page_number = self.request_data.get(page_kwarg, 1)
+        try:
+            page_number = int(page_number)
+        except ValueError:
+            if page_number == 'last':
+                page_number = paginator.num_pages
+            else:
+                raise InvalidPage(_(
```
solarissmoke, line 34: This exception message is never displayed to the user, so we can probably do without it being translated.

src/oscar/apps/search/search_handlers.py
```diff
@@ -9,7 +9,43 @@
...
+                    "Page is not 'last', nor can it be converted to an int."))
+        # This can also raise an InvalidPage exception.
+        page = paginator.page(page_number)
+        return paginator, page, page.object_list, page.has_other_pages()
```
solarissmoke, line 38: Do we need to add the two new items to the returned tuple, given that both of them are accessible from `page` itself? (Would be nice if we don't have to change the API for projects that are calling this method).

solarissmoke (changes requested):

src/oscar/apps/search/search_handlers.py
sasha0: Done.

src/oscar/apps/search/search_handlers.py
sasha0: Agreed. Removed.

src/oscar/apps/search/search_handlers.py
sasha0: Yep, agreed. Done.

src/oscar/apps/search/search_handlers.py
sasha0: Yep agreed. Original intention was to replicate the same behaviour as in the `MultipleObjectMixin.paginate_queryset`. Updated.

sasha0: In fact, no - we still need it, because we inherit both `MultipleObjectsMixin` within this mixin class in the `SimpleProductSearchHandler` and thus this method needs to have the same signature as there.

specialunderwear (approved):
looks good

PR #3506, opened 2020-09-23
issues/3428/thank_you

### LINKED ISSUE #3428: /oscar/checkout/thank-you/` returns 404 when checkout_order_id not in session

sasha0:
Fixes #3428

### DIFF
--- a/src/oscar/apps/checkout/views.py
+++ b/src/oscar/apps/checkout/views.py
@@ -2,6 +2,7 @@ import logging
 from urllib.parse import quote
 
 from django import http
+from django.conf import settings
 from django.contrib import messages
 from django.contrib.auth import login
 from django.shortcuts import redirect
@@ -659,6 +660,13 @@ class ThankYouView(generic.DetailView):
     template_name = 'oscar/checkout/thank_you.html'
     context_object_name = 'order'
 
+    def get(self, request, *args, **kwargs):
+        self.object = self.get_object()
+        if self.object is None:
+            return redirect(settings.OSCAR_HOMEPAGE)
+        context = self.get_context_data(object=self.object)
+        return self.render_to_response(context)
+
     def get_object(self):
         # We allow superusers to force an order thank-you page for testing
         order = None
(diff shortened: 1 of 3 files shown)


src/oscar/apps/checkout/views.py
```diff
@@ -659,6 +660,12 @@ class ThankYouView(generic.DetailView):
     template_name = 'oscar/checkout/thank_you.html'
     context_object_name = 'order'
 
+    def get(self, request, *args, **kwargs):
+        obj = self.get_object()
```
solarissmoke, line 664: Could I suggest we just replace the super method entirely, to avoid the queries in `get_object()` running twice? So just:

```
def get(self, request, *args, **kwargs):
    self.object = self.get_object()
    if self.object is None:
        return redirect(settings.OSCAR_HOMEPAGE)
    context = self.get_context_data(object=self.object)
    return self.render_to_response(context)
```

sasha0: Yes, makes sense.

sasha0: Alright done.

solarissmoke (approved):

PR #3512, opened 2020-09-24
Order reviews by date created by default in the dashboard.

solarissmoke:
The default ordering for reviews is by score, which is fine when listing reviews for a specific product, but not very useful in the dashboard. This changes so that default dashboard ordering is by date created, most recent first.


sasha0 (approved):

sasha0:
Looks good, but tests need attention.

PR #3513, opened 2020-09-24
Fix inflexible product url

MaziyarMK:
This ensures updating the get_absolute_url in Product model won’t cause issues in dashboard reviews list section.


solarissmoke (approved):
Thanks!

PR #3516, opened 2020-09-25
WIP: Add support for Django 3.1

janfabry:
`widget_type` was introduced in Django 3.1, with a slightly different syntax. We switch to this new syntax, with a comment in the release notes.

The rest was fixed by updating imports, one dependency, and an extra setting in the tests.

This passes the tests, but it does not fix the new warnings (`RemovedInDjango40Warning`).

TODO: actual testing of the sandbox

### DIFF
--- a/src/oscar/templatetags/form_tags.py
+++ b/src/oscar/templatetags/form_tags.py
@@ -1,4 +1,8 @@
+import django
 from django import template
+from django.template.base import TextNode
+
+from oscar.core.compat import FormFieldNode
 
 register = template.Library()
 
@@ -9,21 +13,12 @@ def annotate_form_field(parser, token):
     Set an attribute on a form field with the widget type
 
     This means templates can use the widget type to render things differently
-    if they want to.  Django doesn't make this available by default.
+    if they want to. Until 3.1, Django did not make this available by default.
     """
     args = token.split_contents()
     if len(args) < 2:
         raise template.TemplateSyntaxError(
             "annotate_form_field tag requires a form field to be passed")
-    return FormFieldNode(args[1])
-
-
-class FormFieldNode(template.Node):
-    def __init__(self, field_str):
-        self.field = template.Variable(field_str)
-
-    def render(self, context):
-        field = self.field.resolve(context)
-        if hasattr(field, 'field'):
-            field.widget_type = field.field.widget.__class__.__name__
-        return ''
+    if django.VERSION < (3, 1):
+        return FormFieldNode(args[1])
+    return TextNode('')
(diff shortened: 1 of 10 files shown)


src/oscar/templatetags/form_tags.py
```diff
@@ -24,6 +26,6 @@ def __init__(self, field_str):
 
     def render(self, context):
         field = self.field.resolve(context)
-        if hasattr(field, 'field'):
-            field.widget_type = field.field.widget.__class__.__name__
+        if not hasattr(field, 'widget_type') and hasattr(field, 'field'):
```
solarissmoke, line 29: Could we please move this class to `oscar/core/compat.py` with a comment explaining that the tag will be redundant after we drop support for Django 2.2 and lower? That way we will not forget to clean things up when it does become redundant.

tox.ini
```diff
@@ -1,6 +1,6 @@
 [tox]
 envlist =
-    py{36,37,38}-django{22,30}
+    py{36,37,38}-django{22,30,31}
```
solarissmoke, line 3: We also have to update `.travis.yaml` because it runs each tox env in a separate environment.

src/oscar/templatetags/form_tags.py
janfabry: Done in 663b8eb

tox.ini
janfabry: Done in c599d8f

janfabry:
I resolved both issues, and rebased on master. The commits can be squashed once everything is cleared up.

solarissmoke:
There seems to be an issue with how forms are being rendered (both on the front end and the dashboard: `None` is being rendered above the field labels:

image

janfabry:
> There seems to be an issue with how forms are being rendered (both on the front end and the dashboard): `None` is being rendered above the field labels:
> 
> image

My mistake: I return a base `django.template.Node` if Django 3.1 is used, assuming this would not generate any output. But it returns `None`, which is rendered in the template. I will return a `TextNode` instead.

tests/settings.py
```diff
@@ -15,6 +15,8 @@
         'NAME': os.environ.get('DATABASE_NAME', 'oscar'),
         'USER': os.environ.get('DATABASE_USER', None),
         'PASSWORD': os.environ.get('DATABASE_PASSWORD', None),
+        'HOST': os.environ.get('DATABASE_HOST', 'localhost'),
```
solarissmoke, line 18: This causes problems if peer authentication is used instead of password authentication. Could you please leave out this change from this PR, and we can look separately at how to make this work with different auth mechanisms?

solarissmoke (changes requested):
Just one request regarding the change to the database config. Rest looks good and I haven't come across any issues when testing.

solarissmoke (approved):
Looks good, thanks!

Issue #3388, opened 2020-06-04
Basket middleware active in Django admin


jayvdb:
When loading Django admin, the basket middleware failed (below; worthy of a separate less critical issue).  It should detect Django admin and not do anything.

```py
  File "/home/jayvdb/.cache/pypoetry/virtualenvs/django-backend-tUiU8616-py3.8/src/django-oscar/src/oscar/apps/basket/middleware.py", line 41, in load_full_basket
    self.apply_offers_to_basket(request, basket)
  File "/home/jayvdb/.cache/pypoetry/virtualenvs/django-backend-tUiU8616-py3.8/src/django-oscar/src/oscar/apps/basket/middleware.py", line 207, in apply_offers_to_basket
    Applicator().apply(basket, request.user, request)
  File "/home/jayvdb/.cache/pypoetry/virtualenvs/django-backend-tUiU8616-py3.8/src/django-oscar/src/oscar/apps/offer/applicator.py", line 24, in apply
    self.apply_offers(basket, offers)
  File "/home/jayvdb/.cache/pypoetry/virtualenvs/django-backend-tUiU8616-py3.8/src/django-oscar/src/oscar/apps/offer/applicator.py", line 34, in apply_offers
    result = offer.apply_benefit(basket)
  File "/home/jayvdb/.cache/pypoetry/virtualenvs/django-backend-tUiU8616-py3.8/src/django-oscar/src/oscar/apps/offer/abstract_models.py", line 293, in apply_benefit
    if not self.is_condition_satisfied(basket):
  File "/home/jayvdb/.cache/pypoetry/virtualenvs/django-backend-tUiU8616-py3.8/src/django-oscar/src/oscar/apps/offer/abstract_models.py", line 281, in is_condition_satisfied
    return self.condition.proxy().is_satisfied(self, basket)
  File "/home/jayvdb/.cache/pypoetry/virtualenvs/django-backend-tUiU8616-py3.8/src/django-oscar/src/oscar/apps/offer/abstract_models.py", line 48, in proxy
    klass = load_proxy(self.proxy_class)
  File "/home/jayvdb/.cache/pypoetry/virtualenvs/django-backend-tUiU8616-py3.8/src/django-oscar/src/oscar/apps/offer/utils.py", line 29, in load_proxy
    raise exceptions.ImproperlyConfigured(
django.core.exceptions.ImproperlyConfigured: Error importing module oscarbluelight.offer.conditions: No module named 'oscarbluelight'
```

[...]

PR #3518, opened 2020-09-26
WIP Issues/3388/basket middleware

sasha0:


src/oscar/core/application.py
```diff
@@ -93,6 +96,9 @@ def post_process_urls(self, urlpatterns):
                 if decorator:
                     pattern.callback = decorator(pattern.callback)
 
+                if self.is_basket_enabled and pattern.name not in self.views_with_disabled_basket:
+                    pattern.callback.is_basket_enabled = True
```
samitnuk, line 100: Should we add `self.views_with_enabled_basket`? This can be useful in case we want to add the basket to some views where config has `is_basket_enabled = False`.
E.g.
```python
if any([
    pattern.name in self.views_with_enabled_basket,
    self.is_basket_enabled and pattern.name not in self.views_with_disabled_basket,
]):
    pattern.callback.is_basket_enabled = True
```

samitnuk: Looks like we do not need `self.views_with_enabled_basket` because we can use the `enable_basket` views decorator. With it, additionally, we will be able to enable the basket for views in apps that do not use `OscarConfig`.

samitnuk:
@solarissmoke @sasha0 

I have slightly updated this PR. Please review when you will have a chance. Thanks!

src/oscar/core/application.py
sasha0: Yeah, I agree with your last conclusion.

src/oscar/apps/basket/middleware.py
```diff
@@ -19,78 +20,86 @@ def __init__(self, get_response):
...
-    def process_response(self, request, response):
-        # Delete any surplus cookies
-        cookies_to_delete = getattr(request, 'cookies_to_delete', [])
-        for cookie_key in cookies_to_delete:
-            response.delete_cookie(cookie_key)
-
-        if not hasattr(request, 'basket'):
-            return response
+    @staticmethod
+    def is_basket_enabled(request):
```
sasha0, line 67: Is `staticmethod` really necessary? We call this method as generic class method and pass instance within this middleware class (`self.is_basket_enabled(request)`). E.g. can't find static method calls.

tests/unit/core/test_application.py
```diff
@@ -40,4 +43,4 @@ def test_post_process_urls_adds_decorator(self):
         processed_patterns = self.myapp.post_process_urls([pattern])
 
         self.myapp.get_url_decorator.assert_called_once_with(pattern)
-        self.assertEqual(processed_patterns[0].callback, 'fake_callback')
+        self.assertEqual(processed_patterns[0].callback, fake_callback)
```
sasha0, line 46: Good finding. Let's extract this change into the separate PR.

src/oscar/core/application.py
sasha0: Thus, let's remove `views_with_disabled_basket` as well.

sasha0 (commented):
Also we need to think about the case when basket is disabled within the view, but called in the templates. Easiest approach I have in mind is to throw exception/message that basket is disabled. Why do you think guys @samitnuk @solarissmoke?

src/oscar/core/application.py
samitnuk: > Thus, let's remove `views_with_disabled_basket` as well.

How in this case will be better to disable basket for views for apps where config has `is_basket_enabled = True`?

sasha0: We can use next approach - if basket is enabled on the application basis and disabled on the view basis just override application settings by the view. Which would finally disable basket for this view.

samitnuk: Not sure how we can do this on the view level. Could you please explain a bit more detailed?

sasha0: On the view level we'll just add decorator and it will supersede application config.

tests/unit/core/test_application.py
samitnuk: Done in https://github.com/django-oscar/django-oscar/pull/3556

samitnuk:
> Also we need to think about the case when basket is disabled within the view, but called in the templates. Easiest approach I have in mind is to throw exception/message that basket is disabled. Why do you think guys @samitnuk @solarissmoke?

Yea, good idea, but I am not sure how can we check that the basket is called in a template.

From the docs:
> Many templates, including some of Django’s, rely upon the silence of the template system when a nonexistent variable is encountered.

solarissmoke:
> Also we need to think about the case when basket is disabled within the view, but called in the templates. Easiest approach I have in mind is to throw exception/message that basket is disabled.

I don't think we should raise an exception for the reasons mentioned by @samitnuk - it should be up to projects to ensure they only disable the basket on views that really don't use it at all.

specialunderwear:
I'm a little late to the party, I'm sorry, but I really don't like this approach.

The issue this PR would solve (#3388  ) show in the stacktrace that ``load_full_basket`` is being called which somehow loads custom proxy models that breaks due to a cyclic import. Since ``load_full_basket`` is wrapped in a ``SimpleLazyObject`` means that the request.basket was actually accessed by that users code even if it was an admin url! Disabling the basket would not solve the problem this user is having, because it would simpy cause an AttributeError! (request has no attribute basket).

I think that basketmiddleware is just fine, since it is lazy and doesn't really do anything util somebody actually accessed the basket.

Issue #3414, opened 2020-06-20
Allowing OSCAR_HIDDEN_FEATURES apps to be excluded from URLConf


jayvdb:
I have config
```py
OSCAR_HIDDEN_FEATURES = ['reviews', 'voucher', 'wishlists']
```

Voucher is very deeply coupled with basket internals, so couldnt possibly be removed entirely.

I can appreciate that "hidden" means the apps `oscar.apps.wishlist` & `oscar.apps.catalogue.reviews` still need to be in `INSTALLED_APPS`, however I believe they should be excluded from URL paths.  Exposing these features on the site isnt really 'hidden' in my perspective.  I am assuming that those views will be inoperative, to some extent, but they really should not be exposed at all.  Leaving them on the site, but hidden, is a concern that they might be used by hackers, but they are out of sight, out of mind, from the site admins.

`CustomerConfig` depends on `wishlists`, and `DashboardConfig` tries to load `reviews_dashboard`.  `BasketConfig` doesnt require voucher, but deeper down there were some `reverse` which require the voucher views.

I have hacked those so they dont emit url patterns for those apps, if they are in `OSCAR_HIDDEN_FEATURES`.  Is that a direction Oscar wants to go in?  If so, I am happy to contribute that.

If there is interest, I would be happy to take it a bit further to allow excluding reviews or wishlists apps from `INSTALLED_APPS`, however I suspect this will be harder to build tests for.  This is less important afaics - my primary goal was removing those apps from the URLConf, but removing them from INSTALLED_APPS was a way to validate that there wasnt any remaining unknown hooks into those hidden features.  we are running without reviews or wishlists in INSTALLED_APPS - vouchers is not worth the pain to remove, and likely we would want its functionality in the near future anyway.

solarissmoke:
A couple of thoughts on this:

1. In the long run we need to try to decouple Oscar's apps as much as possible, and make them pluggable. `reviews` and `wishlists` could without too much effort be made completely standalone and optional. `voucher` is trickier but there are still unnecessary interdependencies which could be improved.

2. I would like to drop `OSCAR_HIDDEN_FEATURES` entirely and instead move to a pluggable app architecture: to disable you reviews you would simply remove it app from `INSTALLED_APPS`. In templates, `iffeature` would then just become a wrapper for `if apps.is_installed()` (or replaced by some equivalent).

The reviews and/or wishlist apps would I think be a good test case to validate that this approach is viable, so if you have time to work on it that would be great.

[...]

PR #3519, opened 2020-09-28
WIP issues/3414/decouple apps

sasha0:
Refs #3414


solarissmoke:
@sasha0 this is a good start - a few comments:

1. We can't use `apps.is_installed()` to check whether an oscar app is installed - that will not work with forked apps because the argument is the app name (i.e., module path). We will instead have to do something similar to what the dynamic class loader does to find a matching app.

2. I think it would be good to do this in very discrete steps. In this PR I would suggest we just focus on decoupling the voucher app from the basket app, and don't touch the `iffeature` template tag, as that is not related. I feel it is leave that tag untouched and just deprecate it in favour of a new approach (see point 3 below).

3. It looks like you've been able to decouple voucher and basket pretty well! I think we need to provide a data migration for the `Basket.vouchers` field to the new `Voucher.baskets` field.

4. If we allow decoupling then we also need to perform checks in the template for whether the voucher app is installed - otherwise there will be errors when rendering the voucher form on the basket page, etc. I don't think we should use `iffeature` for this, but instead introduce a new template tag that checks whether an app is installed (same logic as we need for point 1 above).

5. There are a couple of additional methods in the `Basket` model that need conditionals - e.g., `Basket.contains_voucher()`.

samitnuk:
> 3. It looks like you've been able to decouple voucher and basket pretty well! I think we need to provide a data migration for the `Basket.vouchers field` to the new `Voucher.baskets` field.

I am imaging this migration in a next way:
1. Add `baskets` field to `Voucher`, but w/o `related_name='vouchers'`, because with this `related_name` we will have something similar to:
    ```
   voucher.Voucher.baskets: (fields.E302) Reverse accessor for 'Voucher.baskets' clashes with field name 'Basket.vouchers'.
	HINT: Rename field 'Basket.vouchers', or add/change a related_name argument to the definition for field 'Voucher.baskets'.
    ```
2. Add migration with:
    ```python
    for basket in Basket.objects.filter(vouchers__isnull=False):
        for voucher in basket.vouchers.all():
            voucher.baskets.add(basket)
    ```
3. Remove `vouchers` field from `Basket`.
4. Update `baskets` field of `Voucher` with `related_name='vouchers'`.

But the 3rd step's migration (`basket`) will be dependant on migration in the 1st/2nd steps (`voucher`). For instance, [this `basket` app migration](https://github.com/django-oscar/django-oscar/blob/c91f9104456e532754c833bc5d467ca763d9a3cd/src/oscar/apps/basket/migrations/0003_basket_vouchers.py) is dependant on the initial migration in the `voucher` app. I suppose we will not be able to use the mentioned migrations in case `voucher` app is not installed.

@solarissmoke @sasha0 Did I missed something in my considerations? Please comment.

solarissmoke:
@samitnuk your proposed approach makes sense to me. We'll have to properly document this in the release notes for projects that have forked either the basket or the voucher app.

Andruten:
@samitnuk @solarissmoke I would also add an iterator to iterate querysets in order to avoid huge memory consumption.

PR #3520, opened 2020-09-28
Add test to check adding/removing of a voucher

samitnuk:


tests/integration/basket/test_views.py
```diff
@@ -109,3 +111,36 @@ def test_default_shipping_address_for_anonymous_user(self):
...
+    def test_add_voucher(self):
+        """
+        Checks that voucher can be added to basket through appropriate view.
+        """
+        self.add_product_to_basket()
+
+        assert self.voucher.basket_set.count() == 0
+
+        response = self.post(reverse('basket:vouchers-add'), params={'code': self.voucher.code})
+        assert response.status_code == 302  # Redirect to basket summary page
```
sasha0, line 132: We can use some dedicated methods for this:

```python
self.assertIsRedirect(response)
self.assertRedirectsTo(response, 'basket:index')
```

sasha0 (changes requested):

tests/integration/basket/test_views.py
samitnuk: Yes, you are right. Thanks!

sasha0 (approved):

PR #3522, opened 2020-09-29
Add SEO fields to product and category models and make editable in the dashboard

ajharry69:


src/oscar/apps/dashboard/catalogue/forms.py
```diff
@@ -27,17 +29,31 @@
...
-class CategoryForm(BaseCategoryForm):
+class _SEOFormMixin(object):
+    def visible_fields(self):
+        return [field for field in self if not field.is_hidden and not self.is_seo_field(field)]
+
+    def seo_fields(self):
+        return [field for field in self if self.is_seo_field(field)]
+
+    @staticmethod
+    def is_seo_field(field):
```
solarissmoke, line 44: It is better to explicitly define on the class itself which fields are considered SEO fields, so that projects can override.

src/oscar/apps/dashboard/catalogue/forms.py
```diff
@@ -27,17 +29,31 @@
...
+
+
+class CategoryForm(_SEOFormMixin, BaseCategoryForm):
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         if 'slug' in self.fields:
             self.fields['slug'].required = False
             self.fields['slug'].help_text = _('Leave blank to generate from category name')
+        if 'meta_description' in self.fields:
```
solarissmoke, line 55: Don't think we need to do this at init - can do it when defining the field itself, i.e., 

```
meta_description = models.TextField(_('Meta Description'), blank=True, widget=forms.TextArea(attrs={'class': 'no-widget-init'})
```

src/oscar/templates/oscar/base.html
```diff
@@ -7,6 +7,7 @@
 
         <meta http-equiv="content-type" content="text/html; charset=UTF-8" />
         <meta name="created" content="{% now "jS M Y h:i" %}" />
+        <meta name="title" content="{% block meta_title %}{% endblock %}" />
```
solarissmoke, line 10: This is not a valid element. The title needs to go into the `<title>` tag, for which we already have a `{% block title %}`. The `catalogue/detail.html` and `catalogue/browse.html` templates override this block and we need to adjust the content inside each.

src/oscar/templates/oscar/catalogue/detail.html
```diff
@@ -12,8 +12,12 @@
     {{ product.get_title }} | {{ block.super }}
 {% endblock %}
 
+{% block meta_title %}
```
solarissmoke, line 15: Need to change `{% block title %}` above, rather than adding a new block.

src/oscar/apps/catalogue/abstract_models.py
```diff
@@ -128,6 +128,8 @@ class AbstractCategory(MP_Node):
 
     name = models.CharField(_('Name'), max_length=255, db_index=True)
     description = models.TextField(_('Description'), blank=True)
+    meta_title = models.CharField(_('Meta Title'), max_length=255, blank=True, db_index=True)
```
solarissmoke, line 131: We don't need a database index on this field.

src/oscar/apps/dashboard/catalogue/forms.py
```diff
@@ -27,17 +29,31 @@
...
 BaseCategoryForm = movenodeform_factory(
     Category,
-    fields=['name', 'slug', 'description', 'image', 'is_public'],
+    fields=['name', 'slug', 'description', 'meta_title', 'meta_description', 'image', 'is_public'],
     exclude=['ancestors_are_public'])
 
 
-class CategoryForm(BaseCategoryForm):
+class _SEOFormMixin(object):
+    def visible_fields(self):
```
solarissmoke, line 37: `visible_fields` has a specific meaning in Django forms, and I don't think we should override that here. Suggest we have a separate method e.g., `primary_form_fields` which returns the fields we render in the main tab, and then use that in the template.

src/oscar/templates/oscar/dashboard/catalogue/category_form.html
```diff
@@ -23,10 +23,72 @@
...
+    <form action="{% if request.GET.urlencode %}?{{ request.GET.urlencode }}{% endif %}" method="post" class="form-stacked wysiwyg fixed-actions" enctype="multipart/form-data" data-behaviour="affix-nav-errors" autocomplete="off">
+        {% csrf_token %}
+        <div class="row">
+            {% block tab_nav %}
+                <div class="col-md-3">
+                    <div data-spy="affix" class="affix-top" data-offset-top="200">
+                        <div class="table-header">
+                            <h3>{% trans "Sections" %}</h3>
+                        </div>
+                        <ul class="nav nav-list bs-docs-sidenav" id="product_update_tabs">
```
solarissmoke, line 35: `id="product_update_tabs"` has been copied from the product form, but isn't semantically correct for a category form.

solarissmoke (changes requested):

solarissmoke:
@ajharry69 something has gone wrong with the migrations for this change - you seem to have deleted the migrations that actually create the new fields. Please delete all your migrations, and regenerate afresh. Also please rebase this PR onto master - currently the diff if showing lots of changes that are not related to this feature.

src/oscar/apps/catalogue/abstract_models.py
```diff
@@ -128,6 +129,8 @@ class AbstractCategory(MP_Node):
 
     name = models.CharField(_('Name'), max_length=255, db_index=True)
     description = models.TextField(_('Description'), blank=True)
+    meta_title = models.CharField(_('Meta-title'), max_length=255, blank=True)
```
solarissmoke, line 132: Please change the label to "Meta title" instead of "Meta-title". Some for all other other fields below - please remove hyphen.

src/oscar/apps/dashboard/catalogue/forms.py
```diff
@@ -1,3 +1,5 @@
+import re
```
solarissmoke, line 1: Unused import.

ajharry69:
Created a [new PR](https://github.com/django-oscar/django-oscar/pull/3534).

PR #3523, opened 2020-09-30
Cleanup tests

samitnuk:


tests/functional/test_basket.py
```diff
@@ -93,21 +92,15 @@ def test_basket_is_empty(self):
...
+    @override_settings(OSCAR_MAX_BASKET_QUANTITY_THRESHOLD=3)
     def test_adding_more_than_threshold_raises(self):
         dummy_product = create_product(price=D('10.00'), num_in_stock=10)
         url = reverse('basket:add', kwargs={'pk': dummy_product.pk})
         post_params = {'product_id': dummy_product.id,
                        'action': 'add',
                        'quantity': 2}
         response = self.app.post(url, params=post_params)
-        self.assertTrue('oscar_open_basket' in response.test_app.cookies)
+        assert 'oscar_open_basket' in response.test_app.cookies
```
sasha0, line 103: Mb better to use `self.assertIn`?

samitnuk: Not sure. Since we already use `pytest`, I think it is better to use `assert` where possible (at least for syntax highlighting).

For example compare this
```python
assert 'oscar_open_basket' in response.test_app.cookies
```
and this
```
self.assertIn('oscar_open_basket', response.test_app.cookies)
```

sasha0: > Not sure. Since we already use pytest, I think it is better to use assert where possible (at least for syntax highlighting).

Yes, but we already mixing up `assert` and `self.assertIn`.

samitnuk: I think this is fine and step-by-step we will replace all "simple" `self.asser*`s (e.g. `self.assertEqual`) with `assert`.

sasha0: I'm fine with that. WDYT @solarissmoke ?

solarissmoke: Fine with me.

sasha0 (dismissed):

tests/functional/test_basket.py
specialunderwear: I would do it the other way around. In unittest act like unittest does and use self.assert* in functional pytest, use assert

solarissmoke: Hmm yeah, I suppose that would be more consistent. I'm not too fussed either way, but I think what you suggest would be easier for other contributors to follow.

sasha0: So we all agree with my initial comment?

specialunderwear: If needed I can contribute a rant about why pytest is not pythonic anytime

solarissmoke: @specialunderwear may be worth filing an issue to propose a standard approach? The test suite is currently all over the place with inconsistent use of pytest vs test classes. I am not particularly opinionated about either, but I do think the inconsistency makes it harder for new contributors to know what to do.

samitnuk: > I would do it the other way around. In unittest act like unittest does and use self.assert* in functional pytest, use assert

Did not think about this previously. I prefer `assert`s because of "more clear lines". But your suggestion has much more sense than just using `assert`s everywhere. Thanks!

Will update the PR.

sasha0 (approved):

PR #3524, opened 2020-09-30
Update settings.py

sahil564:
you comment styles not good


solarissmoke:
This may be of interest to you: https://stackoverflow.com/questions/45346575/what-does-noqa-mean-in-python-comments .

PR #3531, opened 2020-10-04
Added the customization for favicons

ramanaditya:
## Customization for the favicons

Now we can customize the favicons of the website

- Default is `oscar/favicon.ico`
- Need to add OSCAR_FAVICON = `path to the favicon` in the `settings.py`


solarissmoke:
@ramanaditya thanks for this. I'm not convinced however that this is much easier than just overriding the `block favicon` in the template though - could you explain why you feel that is not sufficient? 

The other point in this context is that `.ico` images are not really considered the best practice any more for HTML5 - we should instead be encouraging use of PNGs etc defined in multiple sizes. I would be more inclined to go that way than to further entrench the `favicon.ico`.

solarissmoke:
Also there is already a default favicon in the repository - we don't need another one for the sandbox.

ramanaditya:
@solarissmoke 
Since we tend to use one favicon for the entire website. I thought of putting it in the settings and not uselessly override the existing block.

Yes, to support `png or other types`, we can add a type in the existing code. I mean I will have to update this PR.

The reason I made the changes in the` sandbox` folder is, I wanted to let people know. We can do this in this way as `sandbox` is the implementation of the app. Probably the better option would be to update the docs.

Once you find it relevant I will make the change in the docs as well.

solarissmoke:
I'm still not convinced by the use case here. There are two situations I can see:

1. You want a custom favicon in the same format that Oscar uses. In this case all you have to do is specify your own static asset for `oscar/favicon.ico` and this will be collected by `collectstatic` rather than the default. I still don't see why a setting is necessary to control the path specifically.

2. You want to completely replace the block and insert say a series of icons of different sizes - in that can you just override the template block which is already very easy to do.

sasha0:
I agree with @solarissmoke - template customization should do it.

ramanaditya:
In that case, we can close this issue.

PR #3534, opened 2020-10-05
Add SEO fields to product and category models and make editable in the dashboard

### LINKED ISSUE #3510: Add SEO fields to product and category models and make editable in the dashboard
There are a few SEO components that we regularly find clients asking for the ability to control on product and category pages:

1. Meta title
2. Meta description
3. Slug

For products, slug, meta title and description are all auto-generated from the product title/description, with no option to override any of these. While this is fine a lot of the time, Oscar should provide the ability to override these.

For categories, meta title and description are also auto-generated from the product title/description, and there is no option to override.

I think we should have a "Search Engine Optimisation" tab on the dashboard edit views for both of these, which allow override of these parameters.

ajharry69:
Closes #3510.


src/oscar/apps/catalogue/abstract_models.py
```diff
@@ -367,8 +369,10 @@ class AbstractProduct(models.Model):
     # Title is mandatory for canonical products but optional for child products
     title = models.CharField(pgettext_lazy('Product title', 'Title'),
                              max_length=255, blank=True)
-    slug = models.SlugField(_('Slug'), max_length=255, unique=False)
+    slug = models.SlugField(_('Slug'), max_length=255, unique=False, blank=True)
```
solarissmoke, line 372: Why has this field changed? This PR should not be altering slug handling at all?

src/oscar/apps/catalogue/migrations/0022_auto_20201006_1246.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 3.0.10 on 2020-10-06 11:46
+
+from django.db import migrations, models
```
solarissmoke, line 3: Please generate a single migration for all the changes, to replace the two you currently have.

src/oscar/templates/oscar/dashboard/catalogue/product_update.html
```diff
@@ -81,6 +81,7 @@ <h3>{% trans "Sections" %}</h3>
                                 {% if not parent %}
                                     <li><a href="#product_recommended" data-toggle="tab">{% trans 'Upselling' %}</a></li>
                                 {% endif %}
+                                <li><a href="#seo" data-toggle="tab">{% trans 'Search Enging Optimization' %}</a></li>
```
solarissmoke, line 84: Spelling - `Enging`

solarissmoke: Also please use the British english spelling for Optimisation everywhere, as en-gb is the default language for Oscar.

src/oscar/apps/dashboard/catalogue/forms.py
```diff
@@ -27,11 +27,27 @@
...
+
+    def seo_form_fields(self):
+        return [field for field in self if self.is_seo_field(field)]
+
+    def is_seo_field(self, field):
+        return field.name in self.seo_fields
+
+
+class CategoryForm(SEOFormMixin, BaseCategoryForm):
+    meta_description = forms.CharField(required=False, label=_('Meta description'), widget=forms.Textarea(attrs={
```
solarissmoke, line 48: This could be moved into `SEOFormMixin` to avoid duplication.

src/oscar/templates/oscar/dashboard/catalogue/product_update.html
```diff
@@ -305,6 +306,24 @@ <h3>{% trans "Upload, change or remove images" %}</h3>
                                 {% endblock recommended_products_content %}
                             </div>
                         {% endblock recommended_products %}
+
+                        {% block seo %}
+                            <div class="tab-pane" id="seo">
+                                <div class="table-header">
+                                    <h3>{% trans "Search Engine Optimization" %}</h3>
+                                </div>
+                                <div class="well product-seo-details">
```
solarissmoke, line 315: `.product-seo-details` isn't being used anywhere it seems - if not then we don't need the class.

src/oscar/templates/oscar/catalogue/detail.html
```diff
@@ -9,11 +9,11 @@
 {% load purchase_info_tags %}
 
 {% block title %}
-    {{ product.get_title }} | {{ block.super }}
+    {{ product.meta_title|default:product.get_title|default:""|striptags }} | {{ block.super }}
```
solarissmoke, line 12: I suggest that instead of doing this we define a `get_meta_title` method on the product. This method should:

1. Check the `product.meta_title` and return that if it exists.
2. If it's a child product, then check the `parent.meta_title` if it exists.
3. If we still have nothing, return `product.get_title`.

See the implementation of `get_title()` for how to check parent products.

We also don't need `striptags` here - the title does not contain HTML.

src/oscar/templates/oscar/catalogue/detail.html
```diff
@@ -9,11 +9,11 @@
...
 {% endblock %}
 
 {% block description %}
-    {{ product.description|default:""|striptags }}
+    {{ product.meta_description|default:product.description|default:""|striptags }}
```
solarissmoke, line 16: As above, please define a `get_meta_description` method to handle obtaining the correct meta description.

solarissmoke (changes requested):

src/oscar/apps/catalogue/abstract_models.py
ajharry69: Without it, the dashboard's `ProductForm` tests would fail since `slug` wasn't part of the form's fields before and it's a required field in the model class which would make it required in the form without the change.

src/oscar/apps/dashboard/catalogue/forms.py
ajharry69: With the move, I would lose the `no-widget-init`  effect on the widget.

solarissmoke: OK - looks like the modelform is cobbering what it defined in the mixin. Having looked a bit closer I think we can define the widget on the `Meta.widgets` property (see line 225 below), and then we don't have to redefine the field itself at all:

```
'meta_description': forms.Textarea(attrs={'class': 'no-widget-init'})
```

You can do the same on the category form.

src/oscar/templates/oscar/dashboard/catalogue/product_update.html
```diff
@@ -81,6 +81,7 @@ <h3>{% trans "Sections" %}</h3>
                                 {% if not parent %}
                                     <li><a href="#product_recommended" data-toggle="tab">{% trans 'Upselling' %}</a></li>
                                 {% endif %}
+                                <li><a href="#seo" data-toggle="tab">{% trans 'Search Engine Optimisation' %}</a></li>
```
solarissmoke, line 84: This is getting a bit nitpicky now, but can we use "Search engine optimisation" (lower case e,o) everywhere just for consistency with all the other tab titles?

solarissmoke (changes requested):

src/oscar/apps/catalogue/abstract_models.py
```diff
@@ -128,6 +128,8 @@ class AbstractCategory(MP_Node):
 
     name = models.CharField(_('Name'), max_length=255, db_index=True)
     description = models.TextField(_('Description'), blank=True)
+    meta_title = models.CharField(_('Meta title'), max_length=255, blank=True)
+    meta_description = models.TextField(_('Meta description'), blank=True)
```
specialunderwear, line 132: Please also allow null=True that will avoid having to make a copy of the entire database during the migration.

src/oscar/apps/catalogue/abstract_models.py
```diff
@@ -369,6 +377,8 @@ class AbstractProduct(models.Model):
                              max_length=255, blank=True)
     slug = models.SlugField(_('Slug'), max_length=255, unique=False)
     description = models.TextField(_('Description'), blank=True)
+    meta_title = models.CharField(_('Meta title'), max_length=255, blank=True)
+    meta_description = models.TextField(_('Meta description'), blank=True)
```
specialunderwear, line 381: Add null=true for the same reason

src/oscar/apps/catalogue/abstract_models.py
```diff
@@ -613,6 +623,20 @@ def get_title(self):
         return title
     get_title.short_description = pgettext_lazy("Product title", "Title")
 
+    def get_meta_title(self):
+        title = self.meta_title
+        if not title and self.is_child:
+            title = self.parent.meta_title
```
specialunderwear, line 629: Why not add a fallback to the product title here?

src/oscar/apps/catalogue/abstract_models.py
```diff
@@ -613,6 +623,20 @@ def get_title(self):
...
+        return title or self.get_title()
+    get_meta_title.short_description = pgettext_lazy("Product meta title", "Meta title")
+
+    def get_meta_description(self):
+        meta_description = self.meta_description
+        if not meta_description and self.is_child:
+            meta_description = self.parent.meta_description
+        return meta_description or self.description
```
specialunderwear, line 637: I think you would have to make it like this:
```
 meta_description = self.meta_description or striptags(self.description)
if not meta_description and self.is_child:
    return self.parent.get_meta_description()
```
Note the use of striptags!!!!

src/oscar/apps/dashboard/catalogue/forms.py
```diff
@@ -220,6 +235,9 @@ def __init__(self, product_class, data=None, parent=None, *args, **kwargs):
             self.instance.product_class = product_class
         self.add_attribute_fields(product_class, self.instance.is_parent)
 
+        if 'slug' in self.fields:
```
specialunderwear, line 238: Is this really something that is part of this PR?

src/oscar/templates/oscar/catalogue/category.html
```diff
@@ -3,11 +3,11 @@
 {% load i18n %}
 
 {% block title %}
-    {{ category.name }} | {{ block.super }}
+    {{ category.get_meta_title }} | {{ block.super }}
 {% endblock %}
 
 {% block description %}
-    {{ category.description|default:""|striptags }}
+    {{ category.get_meta_description|striptags }}
```
specialunderwear, line 10: The tag stripping should allready be done in the method to make the result of the method to make sense at all times.

specialunderwear (changes requested):

src/oscar/apps/catalogue/abstract_models.py
solarissmoke: @specialunderwear this is the `Category` model, not product.

solarissmoke: @specialunderwear we do fall back to the title, below, but I think parent meta title should take precedence over child title?

src/oscar/apps/dashboard/catalogue/forms.py
solarissmoke: Product slug was not previously editable at all, and if we make it editable we have to allow it to be empty (for auto-generation).

ajharry69: Yeah, it wasn't there before and there was a need for it.

src/oscar/apps/catalogue/abstract_models.py
specialunderwear: ok

specialunderwear: but also fall back to the parent title if the title is empty as well?

src/oscar/apps/dashboard/catalogue/forms.py
specialunderwear: ok

src/oscar/apps/catalogue/abstract_models.py
solarissmoke: Yes - `get_title()` will handle the fallback to parent title.

specialunderwear: ok

solarissmoke (approved):

PR #3538, opened 2020-10-08
Make inactive products and categories identifiable in dashboard list …

### LINKED ISSUE #3536: Make inactive products and categories identifiable in dashboard list views
Currently, non-public products are not easy to identify in the dashboard list views - there is nothing to distinguish them from public products and you have to view the edit form to determine whether the item is public or not. 

I think it would be useful to have a column in the table to indicate whether the product is active. Same applies for categories.

capelito:
Closes #3536

### DIFF
--- a/src/oscar/apps/dashboard/catalogue/tables.py
+++ b/src/oscar/apps/dashboard/catalogue/tables.py
@@ -44,9 +44,9 @@ class ProductTable(DashboardTable):
 
     class Meta(DashboardTable.Meta):
         model = Product
-        fields = ('upc',  'is_public', 'date_updated')
-        sequence = ('title', 'upc', 'image', 'product_class', 'variants',
-                'stock_records', '...', 'is_public', 'date_updated', 'actions')
+        fields = ('upc', 'is_public', 'date_updated')
+        sequence = ('title', 'upc', 'image', 'product_class', 'variants', 
+                    'stock_records', '...', 'is_public', 'date_updated', 'actions')
         order_by = '-date_updated'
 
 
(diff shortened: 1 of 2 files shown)


src/oscar/apps/dashboard/catalogue/tables.py
```diff
@@ -71,7 +71,8 @@ class CategoryTable(DashboardTable):
 
     class Meta(DashboardTable.Meta):
         model = Category
-        fields = ('name', 'description')
+        fields = ('name', 'description', 'is_public')
+        sequenze = ('name', 'description', '...', 'is_public', 'actions')
```
solarissmoke, line 75: `sequenze`  should be `sequence`. Also please check the failing lint errors.

solarissmoke (changes requested):

solarissmoke (changes requested):
This looks good for the main tables. We also need to update the table that displays variants for a product (the "Variants" tab of the product edit view) - could you please add a column there as well?

capelito:
Thanks for your feedback, I used the true/false icons to make the status identifiable. Some other views use like  /dashboard/ranges/ use Yes or No. What do you prefer?

solarissmoke:
> I used the true/false icons to make the status identifiable. Some other views (for example /dashboard/ranges/) use Yes or No.

Hmm, I think the true/false are easier to parse visually, so would suggest we use those. I'm going to wait until #3302 is merged before merging this though, so that we can check that the icons still work after we switch to Bootstrap 4.

solarissmoke:
That said I'm not sure whether unicode icons are good from an accessibility perspective. There are probably some guidelines on this sort of thing somewhere which we can refer to - will see if I can find any.

solarissmoke (approved):
Thanks for this - looks good. The accessibility question I mentioned earlier is best addressed separately, as there are currently a lot of accessibility issues in the dashboard.

PR #3540, opened 2020-10-15
Remove utilization of connection.cursor and start using Django ORM fo…

Andruten:
…r get_unused_code method. Added separator kwarg in the docstring


solarissmoke (approved):
Thanks!

PR #3543, opened 2020-10-16
WIP: Fixes contains_product

specialunderwear:
This makes the function actually work.

I will add some test soon

### DIFF
--- a/tests/integration/offer/test_range.py
+++ b/tests/integration/offer/test_range.py
@@ -261,13 +261,13 @@ class TestRangeQuerySet(TestCase):
     def test_excluded_product(self):
         ranges = models.Range.objects.contains_product(self.excludedprod)
         self.assertEqual(
-            ranges.count(), 0, "No ranges shoudl contain the excluded product"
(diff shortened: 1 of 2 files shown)


solarissmoke:
@specialunderwear do you think you will have time to add tests in the next few days (i.e., before we cut 3.0)?

tests/integration/offer/test_range.py
```diff
@@ -231,3 +232,52 @@ def test_ensures_unique_slugs_are_used(self):
         first_range.name = "Bar"
         first_range.save()
         models.Range.objects.create(name="Foo")
+
+
+class TestRangeQuerySet(TestCase):
```
solarissmoke, line 237: @specialunderwear all these tests are passing for me without any change to `apps/offer/queryset.py`? 

I think it would be helpful to add a test that fails before applying the change so that we (a) know what has been fixed and (b) ensure it doesn't regress in future.

Also there's a couple of typos in the strings below "shoudl" -> "should"

solarissmoke (changes requested):

Issue #3530, opened 2020-10-03
Upsell messages do not get updated for exclusive offers


malefice:
### Issue Summary

Upsell messages do not get updated. I have tested using the sandbox site, and so far the issue seems present for product count conditions and basket value conditions.

If you remove the exclusive flags, the upsell messages do get updated.

### Steps to Reproduce

1. Visit an offers page. For this step, use the normal site offer [page](https://latest.oscarcommerce.com/en-gb/offers/dave/) of the sandbox site.
2. The normal site offer requires 3 products, and the upsell message is correct at this point. Add any product from that page.
3. Since there is already a product in the basket, the expectation is that you only need to buy 2 more products, but the upsell message does not change.

### Technical details

Present in the sandbox site.

PR #3548, opened 2020-10-26
Fix upsell messages for exclusive offers (issue #3530)

samitnuk:
Closes https://github.com/django-oscar/django-oscar/issues/3530

### DIFF
--- a/src/oscar/apps/basket/abstract_models.py
+++ b/src/oscar/apps/basket/abstract_models.py
@@ -772,6 +772,9 @@ class AbstractLine(models.Model):
     def is_available_for_offer_discount(self, offer):
         return self.consumer.available(offer) > 0
 
+    def quantity_available_for_offer(self, offer):
+        return self.quantity_without_offer_discount(offer) + self.quantity_with_offer_discount(offer)
+
     # ==========
     # Properties
     # ==========
--- a/src/oscar/apps/offer/conditions.py
+++ b/src/oscar/apps/offer/conditions.py
@@ -57,9 +57,8 @@ class CountCondition(Condition):
             return getattr(self, '_num_matches')
         num_matches = 0
         for line in basket.all_lines():
-            if (self.can_apply_condition(line)
-                    and line.quantity_without_offer_discount(offer) > 0):
-                num_matches += line.quantity_without_offer_discount(offer)
+            if self.can_apply_condition(line):
+                num_matches += line.quantity_available_for_offer(offer)
         self._num_matches = num_matches
         return num_matches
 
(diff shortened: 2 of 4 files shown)


sasha0 (dismissed):

solarissmoke:
@samitnuk does this completely fix #3530, or is there still issues to be resolved there?

samitnuk:
@solarissmoke I think this PR completely fix #3530.

Updated my initial description from __Refs: #3530__ to __Closes #3530__.

src/oscar/apps/basket/abstract_models.py
```diff
@@ -772,6 +772,9 @@ def quantity_without_offer_discount(self, offer):
     def is_available_for_offer_discount(self, offer):
         return self.consumer.available(offer) > 0
 
+    def quantity_matches_offer_condition(self, offer):
```
solarissmoke, line 775: The name of this method is not clear to me - could you clarify what its return value represents?

src/oscar/apps/offer/conditions.py
```diff
@@ -70,9 +69,12 @@ def is_partially_satisfied(self, offer, basket):
     def get_upsell_message(self, offer, basket):
```
solarissmoke, line 69: This method used to previously always return a string - now it can return `None`. Will that not cause problems [here](https://github.com/django-oscar/django-oscar/blob/81ec4da73acb703c94ebf9214d4e54aad2e5721a/src/oscar/apps/basket/views.py#L121-L122), because now `None` will get passed to the template instead of a string?

solarissmoke: (If not, is it worth adding a test to verify that `BasketView.get_upsell_messages` doesn't include any `None` values?)

solarissmoke (changes requested):

src/oscar/apps/offer/conditions.py
samitnuk: As I can see returning `None` won't cause any problems - `None` will be returned only for already applied offers  (only applied offers can have `delta <= 0`).

Please see the code below (lines 87-88). For applied offers both `offer.is_condition_partially_satisfied(basket)` and `offer not in applied_offers` are `False`.
https://github.com/django-oscar/django-oscar/blob/81ec4da73acb703c94ebf9214d4e54aad2e5721a/src/oscar/apps/basket/views.py#L81-L93

Will try to add some test to ensure my assumption is correct.

src/oscar/apps/basket/abstract_models.py
samitnuk: The return value represents the line's quantity that is available for the offer - probably `quantity_available_for_offer` is a better name. Thanks!

src/oscar/apps/offer/conditions.py
samitnuk: @solarissmoke

Added additional test in https://github.com/django-oscar/django-oscar/pull/3548/commits/de86d7f43371f58d919d9d23e682b9cef0d2bbc8. Please re-check this PR when you will have a chance. Thanks!

solarissmoke (dismissed):

solarissmoke:
@samitnuk looks good - could you please rebase to fix conflicts?

samitnuk:
@solarissmoke Thanks! Rebased.

PR #3551, opened 2020-10-29
Add caching to Travic configuration to speed up builds.

solarissmoke:
See https://docs.travis-ci.com/user/caching/#pip-cache .


solarissmoke:
Comparing the build logs to before/after, it looks like Travis does this intelligently anyway, so it's not needed.

PR #3554, opened 2020-10-30
Fix checkout.ThankYouView.get_object to never raise DoesNotExist error

jwayodi:
Also fix "is_anonymous" attribute value in WebTestCase subclass.

### DIFF
--- a/src/oscar/apps/checkout/views.py
+++ b/src/oscar/apps/checkout/views.py
@@ -671,17 +671,17 @@ class ThankYouView(generic.DetailView):
         # We allow superusers to force an order thank-you page for testing
         order = None
         if self.request.user.is_superuser:
+            kwargs = {}
             if 'order_number' in self.request.GET:
-                order = Order._default_manager.get(
-                    number=self.request.GET['order_number'])
+                kwargs['number'] = self.request.GET['order_number']
             elif 'order_id' in self.request.GET:
-                order = Order._default_manager.get(
-                    id=self.request.GET['order_id'])
+                kwargs['id'] = self.request.GET['order_id']
+            order = Order._default_manager.filter(**kwargs).first()
 
         if not order:
             if 'checkout_order_id' in self.request.session:
-                order = Order._default_manager.get(
-                    pk=self.request.session['checkout_order_id'])
+                order = Order._default_manager.filter(
+                    pk=self.request.session['checkout_order_id']).first()
         return order
 
     def get_context_data(self, *args, **kwargs):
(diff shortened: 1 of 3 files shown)


src/oscar/apps/checkout/views.py
```diff
@@ -672,16 +672,16 @@ def get_object(self, queryset=None):
...
             if 'order_number' in self.request.GET:
-                order = Order._default_manager.get(
-                    number=self.request.GET['order_number'])
+                order = Order._default_manager.filter(
+                    number=self.request.GET['order_number']).first()
             elif 'order_id' in self.request.GET:
-                order = Order._default_manager.get(
-                    id=self.request.GET['order_id'])
+                order = Order._default_manager.filter(
+                    id=self.request.GET['order_id']).first()
```
sasha0, line 679: Can be a bit simpler:

```python
kwargs = {}
if 'order_number' in self.request.GET:
    kwargs['number'] = self.request.GET['order_number']}
elif 'order_id' in self.request.GET:
    kwargs['id'] = self.request.GET['order_id']).first()
order = Order._default_manager.filter(**kwargs).first()
```

sasha0 (changes requested):

src/oscar/apps/checkout/views.py
jwayodi: Thanks. Fixed.

tests/integration/checkout/test_views.py
```diff
@@ -5,7 +5,7 @@
 
 
 class ThankYouViewTestCase(WebTestCase):
-    is_anonymous = False
+    is_anonymous = True
```
sasha0, line 8: Hm, aren't existing test methods supposed to be run for anonymous users?

jwayodi: Yes, the test methods all use `OSCAR_ALLOW_ANON_CHECKOUT = True`, and make anonymous (without specifying a user) requests. So I am not sure why `is_anonymous` would be `False`?

jwayodi:
@sasha0 I have fixed the tests to properly use Django's test client to make requests after modifying the session. I was previously using WebTest to make these requests, which was incorrect.

Also, `tests.integration.checkout.test_views.ThankYouViewTestCase.test_missing_order_id_in_the_session` looks similar to `tests.functional.checkout.test_customer_checkout.TestThankYouView.tests_gets_a_302_when_there_is_no_order`, only that the former makes an anonymous request. I added `tests.integration.checkout.test_views.ThankYouViewTestCase.test_order_id_in_the_session_is_for_a_non_existent_order` in addition to `tests.functional.checkout.test_customer_checkout.TestThankYouView.tests_gets_a_302_when_the_order_id_in_the_session_is_for_a_non_existent_order`, to match the previous case.

solarissmoke (approved):

PR #3556, opened 2020-10-30
Improve `test_post_process_urls_adds_decorator`

samitnuk:
Use function as the return value of `get_url_decorator`. This change needed because `get_url_decorator` method of `OscarConfigMixin` class returns function or `None` and not string.

Refs: https://github.com/django-oscar/django-oscar/pull/3518/commits/5a10e17b8c3b38f10fe6f4241f1f38f249f2726a#r515003098


sasha0:
Looks good. Please add more descriptive commit message, explaining motivation to do this change (e.g. how you got to it and why fixed this way).

samitnuk:
@sasha0 commit message updated.

sasha0 (approved):

PR #3557, opened 2020-10-30
Render only top level categories in primary navigation menu

solarissmoke:
Support for multi-level dropdowns was removed from Bootstrap 4, with the justification being that these are difficult to use on mobile (and BS4 was designed with a mobile-first approach). Our primary navigation template got a bit mangled during the port to BS4 as a result - resulting in 1st and 2nd level categories all being rendered in a flat list.

This changes the primary navigation only to render the top level categories. I think this is fine given that the menu is already inside a dropdown, and that the full category tree is available on search/category pages.


solarissmoke:
In fact, support for the submenu dropdowns was removed in Bootstrap 3, not 4!

specialunderwear (approved):

sasha0 (approved):

PR #3566, opened 2020-11-07
Create a mixin for filtering partner's products so it's extendable

farooqaaa:
I wanted to customize the queryset for partners and after checking the code, I realized I had to extend three classes to do that: `ProductListView`, `ProductCreateUpdateView` and `ProductDeleteView`. It's because the `filter_products` function is not extendable.

This pull request fixes that by creating a mixin that you can easily extend.

### DIFF
--- a/src/oscar/apps/dashboard/catalogue/views.py
+++ b/src/oscar/apps/dashboard/catalogue/views.py
@@ -9,7 +9,7 @@ from django.utils.translation import gettext_lazy as _
 from django.views import generic
 from django_tables2 import SingleTableMixin, SingleTableView
 
-from oscar.core.loading import get_classes, get_model
+from oscar.core.decorators import deprecated
 from oscar.views.generic import ObjectLookupView
 
 (ProductForm,
@@ -65,6 +65,7 @@ AttributeOptionGroup = get_model('catalogue', 'AttributeOptionGroup')
 Option = get_model('catalogue', 'Option')
 
 
+@deprecated
 def filter_products(queryset, user):
     """
     Restrict the queryset to products the given user has access to.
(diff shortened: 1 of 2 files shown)


src/oscar/apps/dashboard/catalogue/views.py
```diff
@@ -65,20 +66,7 @@
 Option = get_model('catalogue', 'Option')
 
 
-def filter_products(queryset, user):
```
solarissmoke, line 68: Could you please leave this function in place for now, and apply the [deprecated](https://github.com/django-oscar/django-oscar/blob/master/src/oscar/core/decorators.py#L6) decorator to it? This will warn any projects that rely on it that it will be removed in the next version of Oscar.

src/oscar/apps/dashboard/catalogue/views.py
```diff
@@ -9,7 +9,7 @@
 from django.views import generic
 from django_tables2 import SingleTableMixin, SingleTableView
 
-from oscar.core.loading import get_classes, get_model
+from oscar.core.loading import get_classes, get_class, get_model
```
solarissmoke, line 12: Please sort these alphabetically.

solarissmoke (changes requested):
Thanks @farooqaaa , this is a useful change.  A couple of minor changes requested.

src/oscar/apps/dashboard/catalogue/views.py
farooqaaa: That's a good idea. I have done that. Please check the commits now.

src/oscar/apps/dashboard/catalogue/views.py
farooqaaa: Done. I have used isort to sort it.

solarissmoke (approved):

PR #3572, opened 2020-11-13
Fix non-exclusive offers with count condition (3162)

### LINKED ISSUE #3162: Exclusive with Count Condition not work anymore
# Issue Summary

I create 2 offers with count condition, **Both are not exclusive**, After oscar version 2.0, It seem like it just apply only 1 offers not apply both anymore

### Steps to Reproduce
1. Create 2 offers like this 
Screen Shot 2019-08-29 at 17 02 19

2. Only 1 offer apply
Screen Shot 2019-08-29 at 17 06 09

**Expected**
1. Should apply both offers if it does not exclusive.?

Note
I try to dig into the issue and  compare the load, I found out that, in condition.py they are some changed about is_satisfied like this image below

Screen Shot 2019-08-29 at 17 08 11

When I reverse code back it work like a charm, Any idea why you guy change that?

### Technical details

* Python version: 3.6.8
* Django version: 1.11.5
* Oscar version: 2.0.1.

samitnuk:
Fixes #3162

* Checking of `offer.benefit.max_affected_items` removed from the `available` method of `LineOfferConsumer` class - available items should not be limited by related benefit (it should be considered when applying the offer).
* Improved `apply` method of `PercentageDiscountBenefit` - affected items must include "quantities with discount applied".
* `is_satisfied` method of `CountCondition` updated to pass `offer` instead of `None` to `line.quantity_without_offer_discount` - this change is required to fix the initial issue.
* Updated `test_available_with_offer` test - initially this test was added to check the lines that we removed from the `available` method of `LineOfferConsumer` class (See fda27a1).
* Added tests to confirm that the initial issue is fixed.

### DIFF
--- a/src/oscar/apps/basket/utils.py
+++ b/src/oscar/apps/basket/utils.py
@@ -143,7 +143,7 @@ class LineOfferConsumer(object):
     def consumers(self):
         return [x for x in self._offers.values() if self.consumed(x)]
 
-    def available(self, offer=None) -> int: # noqa (too complex (11))
+    def available(self, offer=None) -> int:
         """
         check how many items are available for offer
 
@@ -185,9 +185,5 @@ class LineOfferConsumer(object):
                 check = offer.combinations.count() or x.combinations.count()
                 if check and offer not in x.combined_offers:
                     return 0
-
-            # respect max_affected_items
-            if offer.benefit.max_affected_items:
-                max_affected_items = min(offer.benefit.max_affected_items, max_affected_items)
 
         return max_affected_items - self.consumed(offer)
--- a/src/oscar/apps/offer/benefits.py
+++ b/src/oscar/apps/offer/benefits.py
@@ -67,6 +67,7 @@ class PercentageDiscountBenefit(Benefit):
         max_affected_items = self._effective_max_affected_items()
         affected_lines = []
         for price, line in line_tuples:
+            affected_items += line.quantity_with_offer_discount(offer)
             if affected_items >= max_affected_items:
                 break
             if discount_amount_available == 0:
--- a/src/oscar/apps/offer/conditions.py
+++ b/src/oscar/apps/offer/conditions.py
@@ -47,7 +47,7 @@ class CountCondition(Condition):
         num_matches = 0
         for line in basket.all_lines():
             if self.can_apply_condition(line):
-                num_matches += line.quantity_without_offer_discount(None)
+                num_matches += line.quantity_without_offer_discount(offer)
             if num_matches >= self.value:
                 return True
         return False
--- a/tests/integration/basket/test_utils.py
+++ b/tests/integration/basket/test_utils.py
@@ -76,7 +76,7 @@ class TestLineOfferConsumer:
         offer1 = ConditionalOfferFactory(name='offer1', benefit=benefit)
         lines = basket.all_lines()
         assert lines[0].consumer.available(offer1) == 1
-        assert lines[1].consumer.available(offer1) == 5
+        assert lines[1].consumer.available(offer1) == 10
 
     def test_consumed_with_offer(self, filled_basket):
         offer1 = ConditionalOfferFactory(name='offer1')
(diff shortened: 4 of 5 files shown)


bruecksen:
@samitnuk since this is pretty important for my current project I would like to help. It seems to me that the problem comes from the missing offer argument of `line.quantity_without_offer_discount(offer)` in the `is_satisfied` method of `CountCondition`. When I add the offer as an argument, my second offer gets correctly applied to the basket.

Does it make sense that I push the change to this PR and we see If the changes passes the test? Or how could I help?

samitnuk:
Hi @bruecksen :wave:

With `offer` instead of `None` here:
https://github.com/django-oscar/django-oscar/blob/aadacc3c469982a8a617e30dc0cf51486ab6d5a0/src/oscar/apps/offer/conditions.py#L50

These tests fail:
```
tests/integration/offer/test_combination.py::TestACountConditionWithPercentageDiscount
    - test_consumes_correct_number_of_products_for_3_product_basket
    - test_consumes_correct_number_of_products_for_4_product_basket
    - test_consumes_correct_number_of_products_for_6_product_basket
```

Will try to check the related issue one more time asap.

bruecksen:
@samitnuk thanks for trying this. I made the test suite now work locally and could also reproduce that the three tests fail. Unfortunately I understand to little of the detailed implementation to understand why offer should not be passed to the `available` method. Anyways if I can help with anything let me know. Otherwise I'm happy to hear back from you :)

samitnuk:
@bruecksen I am not sure for 100% at the moment, but I believe that we should pass the `offer` to the `available` method.

When we pass `offer` instead of `None` the three tests fail because of these lines (since for these tests used benefit with `max_affected_items = 1`):
https://github.com/django-oscar/django-oscar/blob/aadacc3c469982a8a617e30dc0cf51486ab6d5a0/src/oscar/apps/basket/utils.py#L189-L191

Probably this should be a start point for the "fix". 🤔

bruecksen:
@samitnuk thanks! I investigated a bit further. In the test the offer is set up with a CountCondition of `value = 3` and a PercentageDiscountBenefit with `max_affected_items = 1`. 

`quantity_available_for_offer ` method should return the available quantity for the BasketLine
https://github.com/django-oscar/django-oscar/blob/aadacc3c469982a8a617e30dc0cf51486ab6d5a0/src/oscar/apps/basket/abstract_models.py#L769-L770

but the available method is limited to the `max_affected_items = 1` and so always returns 1. 
https://github.com/django-oscar/django-oscar/blob/aadacc3c469982a8a617e30dc0cf51486ab6d5a0/src/oscar/apps/basket/utils.py#L189-L191

So the following condition is never meet (in the testcase we have two lines in the basket, returned as available is only 1 so we will never get to the count condition 3)
https://github.com/django-oscar/django-oscar/blob/aadacc3c469982a8a617e30dc0cf51486ab6d5a0/src/oscar/apps/offer/conditions.py#L50-L52

Do you know why the available method should limit against the `max_affected_items`? Shouldn't this only be considered when applying the offer?

samitnuk:
@bruecksen 👍 
> Do you know why the available method should limit against the max_affected_items? Shouldn't this only be considered when applying the offer?

I was also wondering why we need to check `max_affected_items` in the `available` method. When I removed this check from the `available` method - there were more other failed tests. Unfortunately,  I did not have time to make a proper investigation based on them. Will try to check when will be able to.

samitnuk:
@bruecksen Looks like this issue fixed now 😉  (except code coverage 🤔 )

@solarissmoke @sasha0 Please check when you will have a chance.

bruecksen:
@samitnuk thanks a lot, I have already seen the commit and looks good :) 

Is it possible that this will make it into the 3.0 release? And when will there be the 3.0 release?

samitnuk:
@bruecksen 
> Is it possible that this will make it into the 3.0 release? And when will there be the 3.0 release?

I do not know for the moment - It depends on the plans of Oscar's core team.

solarissmoke (approved):
Thanks for working on this @samitnuk. The changes you've made here make sense logically. At some point I think we need to try and refactor this logic to make it more easily understood and maintainable, but this is good fix for now.

PR #3573, opened 2020-11-16
Fix wrong Slack URL in README.rst

maerteijn:


solarissmoke:
@maerteijn the URL isn't wrong - it's pointing to a relay server that auto-generates an invitation to the user to enable them to join the workspace. If you go straight to slack there is no way to register unless you've already been invited by another user.

This service was being maintained by @mvantellingen - pinging him to ask if he can check what's up with it.

maerteijn:
@solarissmoke Ok didn't know that. Now you are directed to https://slack.oscarcommerce.com/ which has an expired certificate, is that something you can fix?

solarissmoke:
No - I've pinged @mvantellingen who controls it.

PR #3576, opened 2020-11-18
use aggregation for calculating basket's totals

jolob5l:
I was learning the project and found out that `basket_total_before_discounts_incl_tax` and `basket_total_before_discounts_excl_tax` are calculated in a quite strange way, it's easy to read but i think there is a better way to do it using aggregation.


solarissmoke (approved):
Thanks!

PR #3589, opened 2020-12-01
Use block for header in basket totals template

iqqmuT:
Allows overriding **Basket** header in `templates/oscar/basket/partials/basket_totals.html`.


solarissmoke (approved):
:+1:

PR #3591, opened 2020-12-02
Make layout_2_col template more customizable

iqqmuT:
This adds blocks that allow `layout_2_col` template to be more customizable.

### DIFF
--- a/src/oscar/templates/oscar/layout_2_col.html
+++ b/src/oscar/templates/oscar/layout_2_col.html
@@ -5,13 +5,14 @@ Keep the same header and footer as the standard layout template but override the
 main content area to split it into a sidebar and a content block
 {% endcomment %}
 {% block content_wrapper %}
-<div class="container page">
+<div class="{% block content_wrapper_classes %}container page{% endblock %}">
     <div class="page_inner">
         {% block breadcrumbs %}{% endblock %}
         <div class="row">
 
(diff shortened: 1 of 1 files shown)


solarissmoke:
@iqqmuT thanks for this. I can see how the `content_wrapper_classes` block is useful, but am not convinced about the other two. You could inject content at the bottom of `column_left` by doing:

```
{% block column_left %}
    {{ block.super }}
    <!-- insert any extra content here -->
{% endblock %}
```

... without the need for an additional block to be defined in Oscar. Same applies for the `content` block.

iqqmuT:
The problem is that other Oscar templates override `column_left` block without using `{{ block.super }}`. For example:
* [`templates/oscar/catalogue/browse.html`](https://github.com/django-oscar/django-oscar/blob/master/src/oscar/templates/oscar/catalogue/browse.html#L23)
* [`templates/oscar/search/results.html`](https://github.com/django-oscar/django-oscar/blob/master/src/oscar/templates/oscar/search/results.html#L25)

solarissmoke:
Ah OK, I see the problem now. Could you change the block names to be more generic? The word `promotions` is tied to your specific use case and not necessarily representative of how others may use them. I think `column_left_extra` and `pre_content`  would work.

solarissmoke (approved):

PR #3593, opened 2020-12-03
More customizable navbar

iqqmuT:
Added more blocks so that navbar can be customized easier.

### DIFF
--- a/src/oscar/templates/oscar/partials/nav_primary.html
+++ b/src/oscar/templates/oscar/partials/nav_primary.html
@@ -2,50 +2,62 @@
 {% load category_tags %}
 {% load i18n %}
 
-<nav class="navbar navbar-expand-lg navbar-dark bg-dark">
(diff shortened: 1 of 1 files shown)


src/oscar/templates/oscar/partials/nav_primary.html
```diff
@@ -2,50 +2,62 @@
...
-        {% if not request.basket.is_empty %}
-            {% if request.basket.is_tax_known %}
-                {% blocktrans with total=request.basket.total_incl_tax|currency:request.basket.currency %}
-                    Total: {{ total }}
-                {% endblocktrans %}
-            {% else %}
-                {% blocktrans with total=request.basket.total_excl_tax|currency:request.basket.currency %}
-                    Total: {{ total }}
-                {% endblocktrans %}
+    {% block navbar_cart %}
```
solarissmoke, line 12: Please change this to `navbar_basket`. Oscar uses "basket" everywhere instead of "cart", thanks to it's British origins.

solarissmoke (changes requested):

PR #3596, opened 2020-12-05
Remove setting of default value via Javascript in time picker inputs.

### LINKED ISSUE #3590: DateTimePickerInput setting initial value for empty fields
### Issue Summary

I have found that when using a `DateTimePickerInput` in the dashboard, the field is always automatically initialised with the current date/time, even when it is not a required field, and even when editing an item where the field was saved empty.

This is not so much a problem when first creating an item, but when you go back to edit something later, unless you clear the default value it is saved with the current date. This could easily be overlooked and cause problems.

### Steps to Reproduce

1. Create a basic model that has an optional date field e.g. `optional_date = models.DateTimeField(blank=True, null=True)`.
2. Create a basic dashboard `CreateView` where the `form_class` is a `ModelForm` for the above model. Set the field widget to `DateTimePickerInput`.
2. Fill out the form. Notice the date field is auto-populated. Clear the field and save.
3. Create and go to the `UpdateView`. The date field is re-populated even though the field was saved as `None`.

### Technical details

- `django-oscar` v3.0a0
- Setting the initial value to empty when intialising the `ModelForm` does not work as the behaviour is caused by the line `'initialDate': new Date(new Date().setSeconds(0))` in `dashboard.js`. I have added my own JS workaround to remedy it, but the default behaviour seems a little confusing to me.

solarissmoke:
Where an initial value is required, this should be set on the form field itself.

Fixes #3590.

### DIFF
--- a/docs/source/releases/v3.0.rst
+++ b/docs/source/releases/v3.0.rst
@@ -154,6 +154,11 @@ Minor changes
 - The behaviour of ``checkout.views.ThankYouView`` when no order is found in the session has changed,
   to redirect the user to the URL defined by ``settings.OSCAR_HOMEPAGE`` instead of returning a page not found error.
 
+- The Javascript that initialises the widget for ``oscar.forms.widgets.DateTimePickerInput`` in dashboard forms
+  no longer sets a default value on the field if no initial value has been supplied for the form field. The
+  ``oscar.dashboard.options.initialDate`` Javascript utility function which was previously used to set this has been
+  removed. This change only affects date-time inputs in the dashboard.
+
 Dependency changes
 ~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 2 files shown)


docs/source/releases/v3.0.rst
```diff
@@ -154,6 +154,11 @@ Minor changes
 - The behaviour of ``checkout.views.ThankYouView`` when no order is found in the session has changed,
   to redirect the user to the URL defined by ``settings.OSCAR_HOMEPAGE`` instead of returning a page not found error.
 
+- The Javascript that initialises the widget for ``oscar.forms.widgets.DateTimePickerInput`` no longer sets a
+  default value on the field if no initial value has been supplied for the form field. The
+  ``oscar.dashboard.options.initialDate`` Javascript utility function which was previously to set this has been
```
jwayodi, line 159: `previously to set` should read `previously used to set`.

Also, you should probably specify that this was an issue only in the dashboard (and not on the front end).

jwayodi:
@solarissmoke Just a minor issue with the accompanying release notes, otherwise it looks good.

PR #3598, opened 2020-12-10
Fix use of aggregation to calculate order basket totals before discount

jwayodi:
Originally implemented in commit 310f82ea512f9faa353cb1663455761ddc9af59a.


solarissmoke:
As per offline discussion, the aggregation result would only be null if an order has no lines, which seems like it would only happen in tests. #3601 has updated `OrderFactory` to allow creation of lines.

PR #3603, opened 2020-12-23
Fixes slow range query on postgres with lots of categories

specialunderwear:
Postgres does the category subquery multiple times, even when there are no
categories. That makes this query VERY slow when there are a lot of categories.
This PR removes the subqueries and instead constructs an or clause on all the
categories in the range. This means ranges now perform 2 queries instead of 1
but this is still a lot faster.


solarissmoke (changes requested):
Could you please fix the lint errors?

solarissmoke (approved):

PR #3604, opened 2020-12-23
Add MIR support

gormih:
Add system MIR to supported payments cards

### DIFF
--- a/src/oscar/apps/payment/bankcards.py
+++ b/src/oscar/apps/payment/bankcards.py
@@ -7,6 +7,7 @@ JCB = 'JCB'
 LASER = 'Laser'
 SOLO = 'Solo'
 SWITCH = 'Switch'
+MIR = 'MIR'
 
 # List of (type, lengths, prefixes) tuples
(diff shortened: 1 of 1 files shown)


gormih:
MIR - is new Russian payments system.

src/oscar/apps/payment/bankcards.py
```diff
@@ -32,6 +33,7 @@
     (VISA, (13, 16), ('4',)),
     (VISA_ELECTRON, (16,), ('4026', '417500', '4405', '4508',
                             '4844', '4913', '4917')),
+    (MIR, (16), ('2')),
```
solarissmoke, line 36: Both these need to be tuples, for which a trailing comma is required inside each bracket. 

Also from [the Wikipedia entry](https://en.wikipedia.org/wiki/Payment_card_number) for this, the prefixes are a bit more complex than just `2` - I think an approach similar to what has been used for Discover above is required.

solarissmoke (changes requested):

solarissmoke:
Closing due to inactivity - feel free to reopen with the requested changes.

PR #3621, opened 2021-01-25
Fix reference to static file path in exception message for symlink_missing_image.

### LINKED ISSUE #3618: static_src and image_not_found.jpg
In oscar 3.0 (beta & master), when image_not_found.jpg is not found, I am getting

> ImproperlyConfigured("Please copy/symlink the 'missing image' image at /path/to/project/public/media/image_not_found.jpg into your MEDIA_ROOT at /path/to/project/public/media. This exception was raised because Oscar was unable to symlink it for you.")

The message is illogical.

Digging around I find my existing symlink is broken, pointing to `/path/to/oscar_repo/../static/oscar/img/image_not_found.jpg` when it should be pointing to `static_src`.

It would be nice if the `MissingProductImage` was able to detect a broken symlink and deletes it before trying to recreate it.

solarissmoke:
Fixes #3618 - the "copy from" path being supplied to this exception message was wrong - instead of reporting the location of the static asset that should be copied manually (`static_file_path`), we were reporting the path it should be copied into (`media_file_path`).

### DIFF
--- a/src/oscar/apps/catalogue/abstract_models.py
+++ b/src/oscar/apps/catalogue/abstract_models.py
@@ -1292,13 +1292,11 @@ class MissingProductImage(object):
                     "Please copy/symlink the "
                     "'missing image' image at %s into your MEDIA_ROOT at %s. "
                     "This exception was raised because Oscar was unable to "
-                    "symlink it for you.") % (media_file_path,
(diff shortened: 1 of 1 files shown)


jwayodi (approved):
This looks good.

PR #3623, opened 2021-01-25
Refactor order search form to be inclusive of orders placed on 'to' date

v1kku:


src/oscar/apps/customer/forms.py
```diff
@@ -226,14 +227,19 @@ def _orders_description(self, date_from, date_to, order_number):
...
-            kwargs['date_placed__gt'] = date_from
-        elif not date_from and date_to:
+
+        if date_from:
+            kwargs['date_placed__gte'] = date_from
+
+        if date_to:
+            # A day is added because date_to is inclusive so that if you search for orders placed from 1st to 3rd you
+            # will get orders placed on 1st 2nd and 3rd.
+            date_to += timedelta(days=1)
```
solarissmoke, line 240: I think it would be preferable to use the same approach as [we do in the dashboard](https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/dashboard/orders/views.py#L190-L191).

src/oscar/apps/customer/forms.py
```diff
@@ -226,14 +227,19 @@ def _orders_description(self, date_from, date_to, order_number):
...
             kwargs['date_placed__lt'] = date_to
+
```
solarissmoke, line 242: Please avoid non-functional white-space changes.

solarissmoke (changes requested):

solarissmoke (approved):

PR #3625, opened 2021-01-26
Fix surcharge name in blocktrans

bruecksen:
Just a little thing I found in the order_detail template not using the surcharge `name` in `blocktrans` block.


solarissmoke:
Thanks!

PR #3626, opened 2021-01-28
Improve upgrading instructions

maikhoepfel:
I struggled with migrations when upgrading from Oscar 2.0.4 to 2.1.
@solarissmoke kindly helped out, and I want to say thank you by adding
the bits I learned to the docs.


solarissmoke (approved):
Thanks for taking the time to write this up!

PR #3632, opened 2021-01-29
Improve voucher creation in dashboard

### LINKED ISSUE #3614: Refactor voucher creation in the dashboard to allow greater flexibility
### Issue Summary

The voucher app provides a very flexible implementation of voucher functionality, where multiple offers can be associated with a voucher. But the dashboard interface for managing vouchers hides this flexibility and only allows the creation of vouchers with a single offer. Meanwhile, on the offer side, only site-wide offers can be created and edited through the dashboard.

### Proposal

I propose to refactor the voucher functionality as follows:

1. Allow CRUD of all types of offer in the dashboard, making it easy to identify the type of the offer (site-wide, voucher, session, user). We can provide a tabbed interface to see the offers of each type. 

    (By default, session and user offers are not applied anywhere - a future enhancement would be to allow associating user offers with particular users/groups of users, instead of requiring projects to implement this themselves.)

2. Refactor the existing voucher creation functionality so that instead of creating a single offer on the fly, the administrator instead selects one or more existing offers (of type voucher) to add to the voucher. This allows the creation of complex vouchers with a combination of offers.

This provides a much more powerful and flexible interface to the voucher functionality that already exists.

### LINKED ISSUE #3619: TypeError creating voucher without end-date
In oscar 3.0 beta/master, creating a coupon/voucher , not specifying an end date results in

> TypeError: '>' not supported between instances of 'datetime.datetime' and 'NoneType'

### LINKED ISSUE #3652: Add possibility to override voucher.utils.get_offer_name
It would be nice to add the possibility to override the `get_offer_name` method to allow for a custom display name of the conditional offer. I found it not useful to add `Offer for voucher 'Voucher name'` as a default to this and would like to customize it without the need to override all Views which use this function.

jwayodi:
- allow creation of offers of all types
- allow creation of vouchers with multiple offers
- allow creation of voucher sets of all usages, and with multiple offers

Closes #3614, closes #3652, closes #3619.


jwayodi:
This PR makes the following changes:

dashboard index page
--------------------
- display number of `Voucher`, `User`, and `Session` offers in the `Offers, vouchers` block (previously, only `Site`-type offers were displayed)

offers
------
- dashboard offer create and update pages:
  - require specifying the offer's offer type (previously, the `Site` type was selected by default)
  - remove `voucher.VoucherSet.offer` model field, which was previously used to keep track of the `VoucherSet`'s `ConditionalOffer`, for use when updating the `VoucherSet`, but we now just fetch the `ConditionalOffer` from the `VoucherSet`'s `Voucher`s (the first one, since they all have the same `ConditionalOffer`s)
  - only allow changing of an offer's offer type from the `Voucher` type, if it has no vouchers attached to it (TODO)
- dashboard offer-list page:
  - list all types of offers (unlike before, where `Voucher`-type offers were excluded)
  - add `Offer type` and `Num of vouchers` (if `Voucher`-type offers are included in the list) columns
  - allow sorting by `offer_type`
  - allow searching by `offer_type`
  - hide `Edit` action menu item for `Voucher`-type offers
  - hide `Delete` action menu item for offers with attached vouchers
- dashboard offer-delete page:
  - only allow deletion of a `Voucher`-type offer, if it has no vouchers attached to it (TODO)

vouchers
--------
- dashboard voucher create and update pages:
  - require specifying (possibly multiple) pre-created offers that should apply to the voucher (previously, a single offer was automatically generated from details entered in the `benefit_range`, `benefit_type`, `benefit_value`, and `exclusive` fields, which have now been removed)
  - attempts to update vouchers that are part of a set, redirect to the dashboard voucher-set update page (they should not be updated individually)
- dashboard voucher-list page:
  - add `Voucher set` columns, if vouchers that are part of a set, are included in the list
  - rework searching of vouchers, to also allow searching for vouchers that are "only inactive", and "only part of a set" (previously, we could only search for "only active", and "active or inactive" (default); and "only not part of a set" (default), and "part of, or not part of a set")
  - for vouchers that are part of a set, instead of the `Edit` action menu item, display an `Edit set` action menu item, that links to the dashboard voucher-set update page (they should not be updated individually)
- dashboard voucher-detail page:
  - for vouchers that are part of a set, instead of the `Edit` button, display an `Edit set` button that links to the dashboard voucher-set update page (they should not be updated individually)
  - for vouchers that are part of a set, make the `cancel` button link to the dashboard voucher-set detail page
- dashboard voucher-delete page:
  - for vouchers that are part of a set, make the `cancel` button link to the dashboard voucher-set detail page
  - on deletion of vouchers that are part of a set, update the voucher set's `count`, and on success redirect to the dashboard voucher-set detail page

voucher sets
------------
- URLs:
  - change dashboard voucher-set detail URL pattern name from `dashboard:voucher-set` to `dashboard:voucher-set-detail`, for consistency
  - change dashboard voucher-set download URL from `/dashboard/vouchers/sets/{pk}/download/` to `/dashboard/vouchers/sets/download/{pk}/`, for consistency
- dashboard voucher-set create and update pages:
  - require specifying usage and (possibly multiple) pre-created offers that should apply to the vouchers (previously, vouchers of `Single use` usage were created, and were assigned a single offer that was automatically generated from details entered in the `benefit_range`, `benefit_type`, and `benefit_value` fields, which have now been removed)
  - for updates, raise a validation error if the `Number of vouchers` is reduced (they should delete vouchers on the dashboard voucher-set detail page, if they want to do this)
  - rely more on generic create and update view features, including moving model-data creating/updating logic from the view and model, to the model form's `save()` method
- dashboard voucher-set detail page:
  - remove redundant `Status` column from the voucher list, and `Is Active?` field from the voucher search form, as all vouchers have the same status (that of the voucher set)
  - add `Delete` and `cancel` buttons to the bottom of the page, and move the `Update voucher set` button here, from the top right of the page
- add ability to delete voucher sets from within the dashboard

jwayodi:
Implements #3614.

src/oscar/apps/dashboard/views.py
```diff
@@ -189,6 +216,9 @@ def get_stats(self):
         if user.is_staff:
             stats.update(
                 total_site_offers=self.get_active_site_offers().count(),
+                total_voucher_offers=self.get_active_voucher_offers().count(),
```
solarissmoke, line 219: I think it should be possible to fetch all three counts in one query with aggregation, something like

```
ConditionalOffer.objects.filter(end_datetime__gt=now()).values('offer_type').aggregate(count=Count('id'))
```

src/oscar/apps/dashboard/offers/forms.py
```diff
@@ -189,3 +189,11 @@ def save(self, *args, **kwargs):
 class OfferSearchForm(forms.Form):
     name = forms.CharField(required=False, label=_("Offer name"))
     is_active = forms.BooleanField(required=False, label=_("Is active?"))
+    offer_type = forms.ChoiceField(required=False, choices=(("", "---------"),) + ConditionalOffer.TYPE_CHOICES)
+
+    @property
+    def is_voucher_offer_type(self):
+        if self.is_bound and (self.cleaned_data['offer_type'] == ConditionalOffer.VOUCHER):
```
solarissmoke, line 196: I think this can be expressed more succinctly as just

```
return self.is_bound and self.cleaned_data['offer_type'] == ConditionalOffer.VOUCHER
```

src/oscar/templates/oscar/dashboard/offers/offer_list.html
```diff
@@ -95,12 +103,16 @@ <h1>{% trans "Offers" %}</h1>
                                         <a class="dropdown-item" href="{% url 'dashboard:offer-detail' pk=offer.pk %}">
                                             {% trans "Stats" %}
                                         </a>
+                                        {% if not offer.is_voucher_offer_type %}
                                         <a class="dropdown-item" href="{% url 'dashboard:offer-metadata' pk=offer.pk %}">
                                             {% trans "Edit" %}
                                         </a>
+                                        {% endif %}
+                                        {% if not offer.vouchers.exists %}
```
solarissmoke, line 111: Note that while this option is hidden here, it is still accessible from the offer detail view (delete button at the top right), and if you use that you can still delete the offer. I think the delete view needs validation to prevent this happening.

src/oscar/apps/offer/abstract_models.py
```diff
@@ -111,7 +111,7 @@ class AbstractConditionalOffer(models.Model):
                     "a user for the duration of their session")),
     )
     offer_type = models.CharField(
-        _("Type"), choices=TYPE_CHOICES, default=SITE, max_length=128)
```
solarissmoke, line 114: I feel like it may be better to leave this default intact, as projects may be relying on it in ways we haven't anticipated, and it doesn't seem like a necessary change to support the feature we're adding. Is there a reason it is necessary?

src/oscar/apps/dashboard/vouchers/forms.py
```diff
@@ -85,15 +78,30 @@ def clean(self):
 class VoucherSearchForm(forms.Form):
     name = forms.CharField(required=False, label=_("Name"))
     code = forms.CharField(required=False, label=_("Code"))
-    is_active = forms.BooleanField(required=False, label=_("Is Active?"))
-    in_set = forms.BooleanField(
-        required=False, label=_("In Voucherset?"))
+    offer_name = forms.CharField(required=False, label=_("Offer name"))
+    is_active = forms.NullBooleanField(required=False, label=_("Is active?"))
```
solarissmoke, line 82: I find the "Unknown" label generated by this widget a little weird - in the sense that it suggests we're asking to filter on vouchers where we don't know what the status of `is_active` is, rather than that we're asking not to filter on this field at all. Not sure how much this is just me being pedantic though - thoughts?

src/oscar/apps/dashboard/vouchers/views.py
```diff
@@ -95,39 +108,20 @@ def get_context_data(self, **kwargs):
...
-        )
-        name = form.cleaned_data['name']
-        offer = ConditionalOffer.objects.create(
-            name=get_offer_name(name),
-            offer_type=ConditionalOffer.VOUCHER,
-            benefit=benefit,
-            condition=condition,
-            exclusive=form.cleaned_data['exclusive'],
-        )
         voucher = Voucher.objects.create(
```
solarissmoke, line 117: Is it possible to subclass `django.views.generic.edit.CreateView` here now that we're not creating offers on the fly? Feels like that would do the object creation for us, plus it would handle the unique constraint checking that we are currently doing manually in `clean_name`, `clean_code` (we would still need to clean the code to uppercase, but would no longer need to check the uniqueness ourselves).

solarissmoke: Similarly subclassing `UpdateView` for `VoucherUpdateView` may help us cut out a lot of code there as well.

solarissmoke (changes requested):
This is looking really good - think it's a big improvement and exposes the true flexibility of Oscar's offer/voucher framework in the dashboard. 

I've left a few comments on the code. In addition it would be great if you could create stub release notes for Oscar 3.1 (copy from 3.0) and in there put a description of these changes. 

I haven't had a close look but we may also need to update the documentation in other places, if there is any regarding the use of the dashboard offer/voucher views.

I think we also need to document the changes to voucher sets, where the `save()` behaviour of that model has changed (to be less dangerous) as has the way counts are handled/enforced.

src/oscar/apps/dashboard/vouchers/views.py
```diff
@@ -4,13 +4,12 @@
...
 from django.db import transaction
 from django.http import HttpResponse, HttpResponseRedirect
-from django.shortcuts import get_object_or_404
+from django.shortcuts import get_object_or_404, redirect
 from django.urls import reverse
 from django.utils import timezone
 from django.utils.translation import gettext_lazy as _
 from django.views import generic
 
-from oscar.apps.voucher.utils import get_offer_name
```
solarissmoke, line 13: @jwayodi `get_offer_name` is no longer used - I think we can remove the method entirely?

solarissmoke (changes requested):

src/oscar/apps/dashboard/offers/forms.py
```diff
@@ -14,7 +13,14 @@
 class MetaDataForm(forms.ModelForm):
     class Meta:
         model = ConditionalOffer
-        fields = ('name', 'description',)
+        fields = ('name', 'description', 'offer_type')
```
specialunderwear, line 16: Since we don't offer a default implementation for USER and SESSION offers, most likely it is unwanted to offer these as choice to the site administrator, enabling an offer with no effect. Maybe ConditionalOffer.TYPE_CHOICES should only include options that actually have an implementation, or maybe define it based on a setting?

specialunderwear (changes requested):

src/oscar/apps/voucher/abstract_models.py
```diff
@@ -119,7 +89,7 @@ class AbstractVoucher(models.Model):
     Oscar enforces those modes by creating VoucherApplication
     instances when a voucher is used for an order.
     """
-    name = models.CharField(_("Name"), max_length=128,
+    name = models.CharField(_("Name"), max_length=128, unique=True,
```
solarissmoke, line 92: The `Voucher` model currently has a `benefit` property which returns the benefit for the first offer - on the assumption that a voucher will likely contain only one offer. I think we should deprecate that property since that assumption is no longer valid (i.e., apply the `oscar.core.decorators.deprecated` decorator to it.

src/oscar/apps/dashboard/offers/forms.py
jwayodi: @specialunderwear  I have fixed this

src/oscar/apps/voucher/migrations/0009_make_voucher_set_voucher_name_unique.py
```diff
@@ -0,0 +1,22 @@
+# Generated by Django 3.1.2 on 2021-02-24 06:18
+
+from django.db import migrations
+
+
+def make_voucher_set_voucher_name_unique(apps, schema_editor):
+    VoucherSet = apps.get_model('voucher', 'VoucherSet')
+    for voucher_set in VoucherSet.objects.all():
```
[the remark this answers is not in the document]
jwayodi: @specialunderwear I have fixed this.

docs/source/releases/v3.1.rst
```diff
@@ -0,0 +1,53 @@
...
+to the related models, signals, and utility functions in the ``offer`` and ``voucher`` apps.
+
+You should review these changes [we'll insert a link to the commit here once it is merged], especially if you have made customisations that depends on these apps, as some are backwards incompatible. The functional changes are summarised below.
+
+- The dashboard offer list view now displays all types of offer (previously voucher offers were hidden), with the
+  ability to filter the list based on type.
+
+- The dashboard offer creation and update views now allow creation of offers of all types (previously only site-wide
+  offers could be created). This exposes the full flexibility of Oscar's offer types in the dashboard. Note that a new
+  setting (``OSCAR_OFFERS_IMPLEMENTED_TYPES``) has been added, which restricts the selectable types to ones that Oscar
```
solarissmoke, line 33: Please add documentation for this setting to `docs/source/ref/settings.rst`.

solarissmoke (changes requested):
Looks good - just one docs issue spotted.

src/oscar/apps/dashboard/offers/forms.py
```diff
@@ -188,4 +203,30 @@ def save(self, *args, **kwargs):
 
 class OfferSearchForm(forms.Form):
     name = forms.CharField(required=False, label=_("Offer name"))
-    is_active = forms.BooleanField(required=False, label=_("Is active?"))
+    is_active = forms.NullBooleanField(required=False, label=_("Is active?"))
+    offer_type = forms.ChoiceField(required=False, label=_("Offer type"), choices=())
```
specialunderwear, line 207: Hi please just use a function to determine the available offer_types, that way you might not need to override the constructor. Choices can be a callable as described in the documentation: https://docs.djangoproject.com/en/3.1/ref/forms/fields/#django.forms.ChoiceField.choices

src/oscar/apps/dashboard/offers/forms.py
```diff
@@ -188,4 +203,30 @@ def save(self, *args, **kwargs):
...
+                                             + tuple(choice for choice in ConditionalOffer.TYPE_CHOICES
+                                                     if choice[0] in get_implemented_offer_type_values()))
+        # Give the "unknown" choice a more meaningful label than the default of "Unknown"
+        NULL_BOOLEAN_FIELD_CHOICES = (
+            ('unknown', _('---------')),
+            ('true', _('Yes')),
+            ('false', _('No')),
+        )
+        self.fields['is_active'].widget.choices = NULL_BOOLEAN_FIELD_CHOICES
+        self.fields['has_vouchers'].widget.choices = NULL_BOOLEAN_FIELD_CHOICES
```
specialunderwear, line 228: Is it really necessary to do this in the constructor? Can it not be specified in the form widget definition?

src/oscar/apps/dashboard/vouchers/forms.py
```diff
@@ -1,99 +1,85 @@
...
+        super().__init__(*args, **kwargs)
+        # Give the "unknown" choice a more meaningful label than the default of "Unknown"
+        NULL_BOOLEAN_FIELD_CHOICES = (
+            ('unknown', _('---------')),
+            ('true', _('Yes')),
+            ('false', _('No')),
+        )
+        self.fields['is_active'].widget.choices = NULL_BOOLEAN_FIELD_CHOICES
+        self.fields['in_set'].widget.choices = NULL_BOOLEAN_FIELD_CHOICES
+        self.fields['has_offers'].widget.choices = NULL_BOOLEAN_FIELD_CHOICES
```
specialunderwear, line 69: If you could specify this in the form field definition that would be preferrable. Since I'm seeing the need for this so many times, maybe just create a special widget class that does what you need? And specify it to be used in the form field.

specialunderwear (changes requested):

specialunderwear (dismissed):

solarissmoke (approved):

PR #3633, opened 2021-02-01
Added display of the separator in the dashboard menu 

rufatpro:
This is for grouping menu items. See example in the picture:

menu_example


JohnRomanski:
I am new to Django-oscar, but could you describe what this PR is going to accomplish?

rufatpro:
> 
> 
> I am new to Django-oscar, but could you describe what this PR is going to accomplish?

See example in the picture:
menu_example

specialunderwear:
Hi, I think this is a lot of code just to add a separator to the dashboard menu. Can you explain why you would find this worth the extra code?

rufatpro:
Yes, it's worth it, I am using this code in my project.

specialunderwear:
I'm sorry but this implementation changes to many thing in both python code and template code to warrant the change. I think a solution would be possible using only css or only html, so we can not accept it like this, sorry.

PR #3642, opened 2021-02-08
Make child products searchable

ajharry69:
Searching by UPC/title in the dashboard should also search child products and show the parent product in the results.

### DIFF
--- a/src/oscar/apps/dashboard/catalogue/views.py
+++ b/src/oscar/apps/dashboard/catalogue/views.py
@@ -159,8 +159,7 @@ class ProductListView(PartnerProductFilterMixin, SingleTableView):
                 queryset = qs_match
             else:
                 # No direct UPC match. Let's try the same with an icontains search.
-                matches_upc = Product.objects.filter(Q(upc__icontains=upc) | Q(children__upc__icontains=upc))\
-                    .distinct()
+                matches_upc = Product.objects.filter(Q(upc__icontains=upc) | Q(children__upc__icontains=upc))
                 queryset = queryset.filter(
                     Q(id__in=matches_upc.values('id')) | Q(id__in=matches_upc.values('parent_id')))
 
--- a/src/oscar/apps/dashboard/catalogue/views.py
+++ b/src/oscar/apps/dashboard/catalogue/views.py
@@ -159,8 +159,8 @@ class ProductListView(PartnerProductFilterMixin, SingleTableView):
                 queryset = qs_match
             else:
                 # No direct UPC match. Let's try the same with an icontains search.
-                matches_upc = Product.objects.filter(Q(upc__icontains=upc) |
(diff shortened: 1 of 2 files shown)


src/oscar/apps/dashboard/catalogue/views.py
```diff
@@ -146,7 +146,7 @@ def apply_search(self, queryset):
             # that contain the UPC.
 
             # Look up all matches (child products, products not allowed to access) ...
-            matches_upc = Product.objects.filter(upc__iexact=data['upc'])
+            matches_upc = Product.objects.filter(Q(upc__iexact=data['upc']) | Q(children__upc__iexact=data['upc']))
```
solarissmoke, line 149: Need to add similar logic below for partial matches.

src/oscar/apps/dashboard/catalogue/views.py
```diff
@@ -163,7 +163,7 @@ def apply_search(self, queryset):
                     Q(id__in=matches_upc.values('id')) | Q(id__in=matches_upc.values('parent_id')))
 
         if data.get('title'):
-            queryset = queryset.filter(title__icontains=data['title'])
+            queryset = queryset.filter(Q(title__icontains=data['title']) | Q(children__title__icontains=data['title']))
 
         return queryset
```
solarissmoke, line 168: I think we need to return `queryset.distinct()` here since it is now possible to return duplicates (e.g., if a child contains the same title as its parent, or if they share part of an SKU).

tests/integration/dashboard/test_catalogue_views.py
```diff
@@ -0,0 +1,24 @@
...
+from oscar.apps.dashboard.catalogue.views import ProductListView
+from oscar.core.loading import get_model
+from oscar.test.factories import create_product
+from oscar.test.utils import RequestFactory
+
+Product = get_model('catalogue', 'Product')
+
+
+class ProductListViewTestCase(TestCase):
+    def test_searching_child_product_by_title_returns_parent_product(self):
```
solarissmoke, line 12: Please add a test to ensure that we don't return duplicates in the event that parent and child share a title.

solarissmoke (changes requested):

src/oscar/apps/dashboard/catalogue/views.py
```diff
@@ -158,14 +159,16 @@ def apply_search(self, queryset):
                 queryset = qs_match
             else:
                 # No direct UPC match. Let's try the same with an icontains search.
-                matches_upc = Product.objects.filter(upc__icontains=data['upc'])
+                matches_upc = Product.objects.filter(Q(upc__icontains=upc) | Q(children__upc__icontains=upc))\
+                    .distinct()
```
solarissmoke, line 163: This is redundant since distinct is applied before returning below.

PR #3646, opened 2021-02-10
Update Solr docs to 6.x

jamesp9:
- Updates instructions for installing Solr 6.x (which seems to be the default generated by Haystack)
- Updates the sample HAYSTACK_CONNECTIONS to use the 6.x settings.


solarissmoke (approved):
Thanks for taking the time to do this!

PR #3647, opened 2021-02-10
Update several test/development dependencies and add support for Python 3.9

solarissmoke:
Also remove `flake8-blind-except` which causes lint errors for cases where we legitimately catch bare exceptions. I think those are better judged on a case-by-case basis rather than blindly.


jwayodi (approved):
This looks good.

PR #3650, opened 2021-02-14
Adding srcsets to oscar.apps.catalog Products

jmarshall9120:
* created module oscar.apps.catalog.srcset.py to manage the creation of
creation of srcsets using the thumbnailed defined in configs.
* srcset contains logic, a descriptor, a SrcsetImage model and a
SrcsetCollection object, all of which are the very basic, simply there
to allow srcsets to work with the models as usual.
* added basic documentation.
* added tests.unit.catalogue.test_srcsets.py (couldn't find more appropriate place to put it,)

-----------------------------------------------
-----TO DO: ---------------------------------
- Test needs work done.  In my local env I tested using an image I made.  Either an image needs to be used, or more work needs to be done mocking the ImageField for the catalog.Products model and the from oscar.core.thumbnails.get_thumbnailer method.


jmarshall9120:
Tests will fail till either we fix the test image so we have something to target, or we mock out all methods that touch the storage.

src/oscar/apps/catalogue/abstract_models.py
```diff
@@ -1318,6 +1319,7 @@ class AbstractProductImage(models.Model):
         help_text=_("An image with a display order of zero will be the primary"
                     " image for a product"))
     date_created = models.DateTimeField(_("Date created"), auto_now_add=True)
+    srcset = SrcSetDescriptor('original')
```
specialunderwear, line 1322: Why would you implement it with a descriptor? The srcset is somethinig that is useful in the frontend when designing the templates. So it would make sence the desicion of what sizes the srcset should be can be determined in the template by the frontend developer, eg full width, 6 column, or 3 column etc.

If srcsets have to be defined on classes like this using the descriptor, we would NEED to override the backend python classes when we need something changed in the frontend. I think that is too complex.

Am I seeing this correctly?

src/oscar/utils/srcsets.py
```diff
@@ -0,0 +1,114 @@
...
+class SrcsetCollection():
+    
+    def __init__(self, instance, srcset_images):
+        self.instance = instance
+        self._srcset = srcset_images
+        return
+    def __getattr__(self, name):
+        rtn = next((si for si in self._srcset if si.size==name), None) \
+            or [si for si in self._srcset if si.image_type==name]
+        if rtn:
```
specialunderwear, line 81: I think that looping in a list on getattr is not efficient.

src/oscar/utils/srcsets.py
```diff
@@ -0,0 +1,114 @@
...
+        return
+    def __getattr__(self, name):
+        rtn = next((si for si in self._srcset if si.size==name), None) \
+            or [si for si in self._srcset if si.image_type==name]
+        if rtn: 
+            return rtn
+        raise AttributeError()
+    def __getitem__(self,name):
+        return next((si for si in self._srcset if si.width==name), None) \
+            or self.__getattr__(name)
```
specialunderwear, line 86: same here.

src/oscar/utils/srcsets.py
```diff
@@ -0,0 +1,114 @@
...
+        
+        srcset = get_srcset(
+            getattr(instance, self.source_field), 
+                image_type=self.image_type, 
+                image_processor=self.image_processor, 
+                do_upscale=self.do_upscale,
+                settings=self.settings, 
+                **self.options
+        )
+        return SrcsetCollection(instance, list(srcset))
```
specialunderwear, line 114: I don't know why you made get_srcset return a generator to then immediately cast it to a list. You might want to use  list comprehension. I do think an implementation of srcset should be lazy and delay creating any images until the image url is actually accessed, which is not what happens here. Correct me if I'm wrong.

specialunderwear (changes requested):

PR #3653, opened 2021-02-19
Remove deprecated providing_args argument from signals.

solarissmoke:
The `providing_args` argument was [deprecated in Django 3.1](https://docs.djangoproject.com/en/3.1/releases/3.1/), with no replacement. It was purely informational. The `providing_args` were actually wrong in quite a few cases (probably because the signal call site has changed, without updating the signal definition). 

The Django documentation recommends using comments to replace this - I've not done that here because I'm not sure that adding comments to where the signal is defined is very useful - it seems better just to inspect the code that is firing the signal?

### DIFF
--- a/src/oscar/apps/basket/signals.py
+++ b/src/oscar/apps/basket/signals.py
@@ -1,8 +1,5 @@
 import django.dispatch
 
-basket_addition = django.dispatch.Signal(
-    providing_args=["product", "user", "request"])
-voucher_addition = django.dispatch.Signal(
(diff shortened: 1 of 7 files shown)


solarissmoke:
Tests are failing because of an unrelated issue - see https://github.com/django-oscar/django-oscar/pull/3654.

jwayodi (approved):
This looks good.

And I agree that adding comments to replace `providing_args` is not very useful.

PR #3656, opened 2021-02-22
how_to_handle_statics.rst: Update make command

jayvdb:
`make css` has been renamed to `make assets`.


jayvdb:
Note https://github.com/django-oscar/django-oscar/blob/master/docs/source/releases/v1.2.rst also mentions `make css` , but I guess updating relnotes is the road to madness.

PR #3657, opened 2021-02-22
Fix static files source dir in "oscar_fork_statics" management command

### LINKED ISSUE #3651: oscar_fork_statics tries to read from non-existant static_src directory
Using the current django-oscar, and django, when I try to copy in the default static assets into Frobshop using `oscar_fork_statics` I get an error. Looks like this problem is related to issues #3621 and #3618.
```
$ ./manage.py oscar_fork_statics
System check identified some issues:

WARNINGS:
catalogue.ProductAttributeValue.value_boolean: (fields.W903) NullBooleanField is deprecated. Support for it (except in historical migrations) will be removed in Django 4.0.
        HINT: Use BooleanField(null=True) instead.
Copying Oscar's static files to /var/work/django/frobshop/static/
Traceback (most recent call last):
  File "./manage.py", line 22, in <module>
    main()
  File "./manage.py", line 18, in main
    execute_from_command_line(sys.argv)
  File "/home/mslinn/oscar/lib/python3.8/site-packages/django/core/management/__init__.py", line 401, in execute_from_command_line
    utility.execute()
  File "/home/mslinn/oscar/lib/python3.8/site-packages/django/core/management/__init__.py", line 395, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/home/mslinn/oscar/lib/python3.8/site-packages/django/core/management/base.py", line 330, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/home/mslinn/oscar/lib/python3.8/site-packages/django/core/management/base.py", line 371, in execute
    output = self.handle(*args, **options)
  File "/home/mslinn/oscar/lib/python3.8/site-packages/oscar/management/commands/oscar_fork_statics.py", line 35, in handle
    shutil.copytree(source, destination)
  File "/usr/lib/python3.8/shutil.py", line 552, in copytree
    with os.scandir(src) as itr:
FileNotFoundError: [Errno 2] No such file or directory: '/home/mslinn/oscar/lib/python3.8/site-packages/oscar/static_src'
```

The source directory that `oscar_fork_statics` looks for (`/home/mslinn/oscar/lib/python3.8/site-packages/static_src`) does not exist:
```
$ ls /home/mslinn/oscar/lib/python3.8/site-packages/oscar
__init__.py  apps       core         forms   management  static     templatetags  utils
__pycache__  config.py  defaults.py  locale  models      templates  test          views
```
There is a directory called `static` that seems reasonable, however.

jwayodi:
Also:
- change the default value of the "target_path" argument
- improve documentation of the "target_path" argument
- fix old documentation that says frontend and dashboard use different
  Bootstrap versions

Fixes #3651.

### DIFF
--- a/src/oscar/management/commands/oscar_fork_statics.py
+++ b/src/oscar/management/commands/oscar_fork_statics.py
@@ -16,7 +16,12 @@ class Command(BaseCommand):
     help = "Copy Oscar's static files"
 
     def add_arguments(self, parser):
-        parser.add_argument('target_path', nargs='?', default='static_src')
+        parser.add_argument('target_path', nargs='?', default='static',
+                            help='Directory path (relative to current working '
+                                 'directory, or absolute) in local project, to '
+                                 'copy files to. This directory should be '
+                                 'included in STATICFILES_DIRS, in order for '
+                                 'your local overrides to be picked up.')
 
     def handle(self, *args, **options):
         # Determine where to copy to
@@ -30,7 +35,7 @@ class Command(BaseCommand):
                 "The folder %s already exists - aborting!" % destination)
 
         source = os.path.realpath(
-            os.path.join(os.path.dirname(__file__), '../../static_src'))
+            os.path.join(os.path.dirname(__file__), '../../static'))
         print("Copying Oscar's static files to %s" % (destination,))
         shutil.copytree(source, destination)
 
--- a/docs/source/howto/how_to_handle_statics.rst
+++ b/docs/source/howto/how_to_handle_statics.rst
@@ -30,14 +30,14 @@ is often preferable.
 Frontend vs. Dashboard
 ----------------------
 
-The frontend and dashboard are intentionally kept very separate. They
-incidentally both use Bootstrap, but may be updated individually.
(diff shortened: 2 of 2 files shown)


jwayodi:
TODO: tests

src/oscar/management/commands/oscar_fork_statics.py
```diff
@@ -30,7 +35,7 @@ def handle(self, *args, **options):
                 "The folder %s already exists - aborting!" % destination)
 
         source = os.path.realpath(
-            os.path.join(os.path.dirname(__file__), '../../static_src'))
+            os.path.join(os.path.dirname(__file__), '../../static'))
```
solarissmoke, line 38: @jwayodi `static/` contains only the compiled assets - not the original less files etc - and I think the latter is what a project would need? Previously both the original and compiled assets were here, but now they are separate. I think we need to copy `static_src` (and ensure it is in the pypi package).

solarissmoke: We discussed this and agreed that it is in fact the compiled statics that we need to fork here.

solarissmoke (approved):

PR #3668, opened 2021-03-05
Remove some unnecessary styling on checkout pages.

### LINKED ISSUE #3666: Don't set max width of quantity fields
Setting `max-width` on checkout's input fields and its wrapper is causing the insufficient width of the input in Firefox (Linux), where the arrows are overflowing the field.

https://github.com/django-oscar/django-oscar/blob/3f5397f4a64825777585047066ebf90a39955a1c/src/oscar/static_src/oscar/scss/page/checkout.scss#L10

https://github.com/django-oscar/django-oscar/blob/3f5397f4a64825777585047066ebf90a39955a1c/src/oscar/static_src/oscar/scss/page/checkout.scss#L13

Django Oscar basket width set

solarissmoke:
Cleanup of CSS for checkout:

1. Remove `max-width` on quantity inputs. The width of the outer wrapper is already constrained by CSS, so allow the input to fill the space available there. 

2. Remove some width constraints on address fields - these seem arbitrary and don't have any effect on the most popular screen sizes anyway.

3. Remove some redundant margins/font weights that are already applied by BS4.

4. Remove centre-alignment for checkout nav, which looks really weird on mobile. This has no effect on large screen display.

Fixes #3666.

### DIFF
--- a/src/oscar/static_src/oscar/scss/page/checkout.scss
+++ b/src/oscar/static_src/oscar/scss/page/checkout.scss
@@ -1,28 +1,8 @@
 // OSCAR CHECKOUT UNIQUE STYLES
 // ----------------------------
 
-// Set max width of all quantity fields
-input[type="text"][id*="quantity"], input[type="number"][id*="quantity"] {
-  max-width: 55px;
-}
-
 .checkout-quantity .input-group {
   max-width: 160px;
-
-  input[type="number"][id*="quantity"] {
-    max-width: 65px;
-  }
-
-  .input-group-btn {
-    min-width: 95px;
-  }
-}
-
-// Quantity on the basket
-.checkout-quantity {
-  input {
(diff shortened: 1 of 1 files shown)


src/oscar/static_src/oscar/scss/page/checkout.scss
```diff
@@ -125,11 +93,8 @@ input[type="text"][id*="quantity"], input[type="number"][id*="quantity"] {
 
 .shipping-payment {
   .card {
-    margin-bottom: $base-margin-bottom;
```
jwayodi, line 128: I think this needs to remain?

solarissmoke: Yeah, you're right, sorry. I've added it back.

jwayodi (approved):
This looks good.

PR #3670, opened 2021-03-05
Remove redundant JS from ui.js

solarissmoke:
Removes a whole load of old JS:

1. `o.page.init` provided in-page scroll effects for links in the top navigation, from a time when those links linked to other sections on the page. It doesn't do anything now.
2. `o.page.init` also added some tooltip handling JS - I can't find any HTML that uses `rel="tooltip"` anywhere in our code.
3. `o.responsive.initNav` was doing nothing at all - the elements it is looking for no longer exist. We should be relying on BS4 for this sort of thing anyway.
4. `o.responsive.initSlider` was initialising the product detail gallery with an interval of 20 seconds (overriding the default of 5 seconds) - which is so long that you'd probably not notice it anyway. I've removed the auto-sliding altogether.

### DIFF
--- a/src/oscar/static_src/oscar/js/oscar/ui.js
+++ b/src/oscar/static_src/oscar/js/oscar/ui.js
@@ -104,52 +104,6 @@ var oscar = (function(o, $) {
         }
     };
 
-
-    o.page = {
-        init: function() {
-            // Scroll to sections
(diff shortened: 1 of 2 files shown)


jwayodi (approved):
This looks good.

PR #3671, opened 2021-03-05
Drop OSCAR_HIDDEN_FEATURES setting in favour of removing app config from INSTALLED_APPS to achieve the same effect

### LINKED ISSUE #3414: Allowing OSCAR_HIDDEN_FEATURES apps to be excluded from URLConf

ajharry69:
Closes #3414


src/oscar/apps/dashboard/apps.py
```diff
@@ -16,40 +15,31 @@ class DashboardConfig(OscarDashboardConfig):
...
+            'pages_dashboard': 'pages/',
+            'partners_dashboard': 'partners/',
+            'offers_dashboard': 'offers/',
+            'ranges_dashboard': 'ranges/',
+            'reviews_dashboard': 'reviews/',
+            'vouchers_dashboard': 'vouchers/',
+            'communications_dashboard': 'comms/',
+            'shipping_dashboard': 'shipping/',
+        }
+
```
specialunderwear, line 33: While you are reworking how the url prefixes and such are done, would you consider making them translatable?

See https://docs.djangoproject.com/en/3.1/topics/i18n/translation/#translating-url-patterns

It can be done like this:
```
return {
             'catalogue_dashboard': _('catalogue/'),
             'reports_dashboard': _('reports/'),
             'orders_dashboard': _('orders/'),
             'users_dashboard': _('users/'),
             'pages_dashboard': _('pages/'),
             'partners_dashboard': _('partners/'),
             'offers_dashboard': _('offers/'),
             'ranges_dashboard': _('ranges/'),
             'reviews_dashboard': _('reviews/'),
             'vouchers_dashboard': _('vouchers/'),
             'communications_dashboard': _('comms/'),
             'shipping_dashboard': _('shipping/'),
         }
```
When composing the full urls, you can make use of ```django.utils.text.format_lazy``` to create the *translatable* view patterns.

solarissmoke: @specialunderwear this is a good idea - I'm going to request that we handle it in a separate PR though, just so that this PR can focus on the iffeature change which is quite substantial in itself.

solarissmoke: (The translation change looks like it will be quite a simple diff after this).

specialunderwear: I think so yes, also it is not important for the dashboard, but more important for the frontend really.

specialunderwear: It is worthwile to keep the translatable urls in mind while completing this though.

src/oscar/core/application.py
```diff
@@ -1,8 +1,55 @@
...
+            app_config_attribute_name = f'{label}_app'
+            if count == 0 and not hasattr(self, app_config_attribute_name):
+                # this method was probably called before calling `self._create_required_attributes()`
+                self._create_required_attributes()
+            count += 1
+
+            endpoint, regex = value, False
+            if isinstance(value, dict):
+                endpoint = value.get('endpoint') or ''
+                regex = value.get('regex', False)
```
jwayodi, line 41: Can't we just detect if the endpoint is a regex (by checking for regexy characters, e.g. `()^$?\+{}`), instead of having it specified explicitly? @solarissmoke?

src/oscar/core/application.py
```diff
@@ -1,8 +1,55 @@
...
+
+            app_config = getattr(self, app_config_attribute_name)
+            if app_config is None:
+                continue  # app with the label probably wasn't installed
+
+            _urls = app_config.urls
```
jwayodi, line 47: Please rename this to `child_app_urls`.

src/oscar/core/application.py
```diff
@@ -1,8 +1,53 @@
-from django.apps import AppConfig
+from django.apps import AppConfig, apps
 from django.core.exceptions import ImproperlyConfigured
-from django.urls import URLPattern, reverse_lazy
+from django.urls import URLPattern, include, path, re_path, reverse_lazy
 
-from oscar.core.loading import feature_hidden
+
+class AutoLoadURLsConfigMixin:
+    include_urls_in_parent = True
```
jwayodi, line 7: Previously, apps in `OSCAR_HIDDEN_FEATURES` had their URLs [excluded](https://github.com/django-oscar/django-oscar/blob/master/src/oscar/core/application.py#L82) (the referenced issue #3414 claims otherwise, or am I misunderstanding something?), in addition to having template sections hidden using the `iffeature` template tag. To achieve the same, one should now remove the app config from `INSTALLED_APPS`.

With this attribute (which should be in the child app config class), we are adding the option for one to exclude the URLs of installed apps. What is the use case for this? And why are we doing this only for child apps, and not all apps in general?

If one really wants to do this (and is ready to deal with the resulting `django.urls.NoReverseMatch` errors), how about they just override the app's `OscarConfig.get_urls()` to return an empty list?

@solarissmoke?

src/oscar/templatetags/display_tags.py
```diff
@@ -42,29 +41,10 @@ def render(self, context):
 get_parameters = register.tag(get_parameters)
 
 
-@register.tag()
-def iffeature(parser, token):
-    nodelist = parser.parse(('endiffeature',))
+@register.filter
+def is_app_installed(app_label):
```
jwayodi, line 45: I think this is better being a block template tag (e.g. `{% if_app_installed "reviews" %}`), like `iffeature` was, as filters are mostly used to transform the value.

Or at least a simple tag, which would be less code and more flexible, but require a bit more work to use:
```
{% is_app_installed "reviews" as reviews_app_is_installed %}
{% if reviews_app_is_installed %}
...
{% endif %}
```

jwayodi (commented):
Removing the apps (`reviews` and `wishlists`) currently supported for `OSCAR_HIDDEN_FEATURES`, from `INSTALLED_APPS`, raises Django-app-registry errors.

src/oscar/core/application.py
ajharry69: @jwayodi Sorry,`include_urls_in_parent` name is/was a bit misleading. It's meant to create a differentiate between two usecases `path("endpoint/", include(...))` and `path("endpoint/", ...)`.

src/oscar/templatetags/display_tags.py
specialunderwear: I would vote for the block tag, because it would make the templates a lot cleaner.

PR #3672, opened 2021-03-05
Fix logic for cleaning voucher start/end datetimes.

### LINKED ISSUE #3619: TypeError creating voucher without end-date

solarissmoke:
Using `all()` with a list is problematic because all items in the list will be evaluated before passing to `all()`, and we can't do the `>` check unless both the datetimes are non-null.

Fixes #3619.

### DIFF
--- a/src/oscar/apps/voucher/abstract_models.py
+++ b/src/oscar/apps/voucher/abstract_models.py
@@ -170,10 +170,8 @@ class AbstractVoucher(models.Model):
         return self.name
 
     def clean(self):
-        if all([self.start_datetime, self.end_datetime,
-                self.start_datetime > self.end_datetime]):
-            raise exceptions.ValidationError(
-                _('End date should be later than start date'))
+        if self.start_datetime and self.end_datetime and self.start_datetime > self.end_datetime:
+            raise exceptions.ValidationError(_('End date should be later than start date'))
 
     def save(self, *args, **kwargs):
         self.code = self.code.upper()


src/oscar/apps/voucher/abstract_models.py
```diff
@@ -170,10 +170,8 @@ def __str__(self):
         return self.name
 
     def clean(self):
-        if all([self.start_datetime, self.end_datetime,
-                self.start_datetime > self.end_datetime]):
-            raise exceptions.ValidationError(
-                _('End date should be later than start date'))
+        if self.start_datetime and self.end_datetime and self.start_datetime > self.end_datetime:
```
jwayodi, line 173: This is also getting added in [#3632](https://github.com/django-oscar/django-oscar/pull/3632/files#diff-996e1c61247748a36e9dbf3c9847d19ed403a2459a8bfcbab1bb35f3fa180044L173-R143).

jwayodi (approved):
This is also getting added in #3632.

src/oscar/apps/voucher/abstract_models.py
solarissmoke: Oh I'd missed that - let's just wait for that.

solarissmoke:
Fixed in #3632.

PR #3673, opened 2021-03-06
Simplify code for several old template tags.

solarissmoke:
Cleans up several old template tags to considerably simplify the code (at the time they were written Django's template tag framework was not as easy to use). None of these tags had any tests, so I've added tests for each as well.

### DIFF
--- a/src/oscar/templatetags/sorting_tags.py
+++ b/src/oscar/templatetags/sorting_tags.py
@@ -4,6 +4,8 @@
 
 from django import template
 from django.conf import settings
+from django.utils.html import format_html
+from django.utils.safestring import mark_safe
 
 register = template.Library()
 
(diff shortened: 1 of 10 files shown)


src/oscar/templatetags/wishlist_tags.py
```diff
@@ -3,36 +3,10 @@
 register = template.Library()
 
 
-@register.tag(name="wishlists_containing_product")
-def do_basket_form(parse, token):
+@register.simple_tag
+def wishlists_containing_product(wishlists, product):
     """
     Template tag for adding the user's wishlists form to the
     template context so it can be rendered.
```
jwayodi, line 10: Please remove the bit in the docstring suggesting it adds the variable to the template context (which was an implementation detail of the previous tag), and just say it returns the value? Or just remove the whole thing.

src/oscar/templatetags/sorting_tags.py
```diff
@@ -21,45 +22,27 @@
...
+        direction = get_vars.pop('dir', [''])[0]
+        get_vars['dir'] = sort_directions[direction]['inverse']
+        icon = sort_directions[direction]['icon']
+
+    href = '%s?sort=%s' % (request.path, field)
+    if len(get_vars) > 0:
+        href += "&%s" % get_vars.urlencode()
+    if icon:
+        title = "%s %s" % (title, icon)
+    return mark_safe('<a href="%s">%s</a>' % (href, title))
```
jwayodi, line 48: It's hard to imagine, but could we somehow accidentally end up with an unsafe `title`?

solarissmoke: Seems unlikely, but I've added escaping for it (and included a check in the tests).

jwayodi (approved):
This looks good.

PR #3677, opened 2021-03-09
Load customer.history with get_class

### LINKED ISSUE #3675: Load customer.history with get_class
### Issue Summary

All functions of customer.history are loaded by using direct import instead of `get_class`, making it very impractical to override.

### Steps to Reproduce

Use cases are for example if you want to change how the object are loaded when getting recently viewed products (I wanted to use search backend instead of DB), or if you want to change some properties of the oscar_history cookie (I wanted to set `samesite` to Lax). Not all use cases can be anticipated, hence this is a good candidate for a `get_class` encapsulation. I think, contrarily to its name, that `get_class` also works for module import, but history module could also be converted to a class with static methods, what would you prefer? I can do the pull request then.

### Technical details

* Python version: 3.9.1
* Django version: 3.1.6
* Oscar version: 3.0.0

Chadys:
Fixes #3675

### DIFF
--- a/src/oscar/apps/customer/history.py
+++ b/src/oscar/apps/customer/history.py
@@ -8,6 +8,14 @@ Product = get_model('catalogue', 'Product')
 
 
 class CustomerHistoryManager:
+    cookie_name = settings.OSCAR_RECENTLY_VIEWED_COOKIE_NAME
+    cookie_kwargs = {
+        'max_age': settings.OSCAR_RECENTLY_VIEWED_COOKIE_LIFETIME,
+        'secure': settings.OSCAR_RECENTLY_VIEWED_COOKIE_SECURE,
+        'httponly': True,
+    }
+    max_products = settings.OSCAR_RECENTLY_VIEWED_PRODUCTS
+
     @classmethod
     def get(cls, request):
         """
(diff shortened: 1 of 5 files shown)


solarissmoke (changes requested):
Thanks for this! Since we're moving it to a class, I think it would be helpful to move some of the parameters into class variables, so that projects can override them easily without replacing entire methods and (if they wish) reuse the class for other purposes entirely:


```
class CustomerHistoryManager:
    cookie_name = settings.OSCAR_RECENTLY_VIEWED_COOKIE_NAME
    cookie_kwargs = {
        'max_age': settings.OSCAR_RECENTLY_VIEWED_COOKIE_LIFETIME,
        'secure': settings.OSCAR_RECENTLY_VIEWED_COOKIE_SECURE,
        'httponly': True,
    }
    max_products = settings.OSCAR_RECENTLY_VIEWED_PRODUCTS
```

And then use these in the class methods, e.g.;:

```
response.set_cookie(cls.cookie_name, json_dumps(updated_ids), **cls.cookie_kwargs)
```

(This would make your use case of modifying the `samesite` kwarg easier)

Chadys:
You're right, updated it ;)

solarissmoke (approved):
Thanks!

PR #3678, opened 2021-03-10
Use minified Bootstrap JS by default.

### LINKED ISSUE #3676: Please load bootstrap.bundle.min.js instead of the unminified version
https://github.com/django-oscar/django-oscar/blob/ce95a8f742c18ce4ee2c9b2556d969cf654e85d7/src/oscar/templates/oscar/layout.html#L55

The production file is 84kB, the unminified one 236kB in size. Also, the version 4.6.0 of Bootstrap 4 is out, maybe it can be updated here as well?
Thank you!

solarissmoke:
Fixes #3676 . The unminified version is still collected into our static directory in case projects were depending on it (and also in case they want to use it for development).


jwayodi (approved):
This looks good.

PR #3679, opened 2021-03-10
Make basket line form's 'quantity' field required

ajharry69:
If you empty the value in the `quantity` field and attempt to update a basket line, you'll get an error saying that "this field is required", which is not ideal. This PR seeks to address that by making the field required at the form level.


solarissmoke (changes requested):
From observing how people use this field, emptying the value is almost always used to try and remove the product from the basket. I think that it would be better if we can adjust the cleaning of the field so that an empty string is converted to 0.

solarissmoke (approved):

Issue #3690, opened 2021-04-06
Line referense depends on options order


Kudria:
Basket line referense depends on options order. This may lead to creation additioanal line instances while adding products with same options, instead of line.quantity increase.

https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/basket/abstract_models.py#L360

This behaviour may not be an issue while using built-in forms and views, but it resulted in hard to find bug my project.

solarissmoke:
Fixed in #3691.

PR #3691, opened 2021-04-12
Make line reference independent from options ordering.

Kudria:
Issue #3690

### DIFF
--- a/src/oscar/apps/basket/abstract_models.py
+++ b/src/oscar/apps/basket/abstract_models.py
@@ -1,5 +1,6 @@
 import zlib
 from decimal import Decimal as D
+from operator import itemgetter
 
 from django.conf import settings
 from django.core.exceptions import ObjectDoesNotExist, PermissionDenied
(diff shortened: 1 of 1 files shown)


src/oscar/apps/basket/abstract_models.py
```diff
@@ -365,8 +365,9 @@ def _create_line_reference(self, product, stockrecord, options):
         base = '%s_%s' % (product.id, stockrecord.id)
         if not options:
             return base
-        repr_options = [{'option': repr(option['option']),
-                         'value': repr(option['value'])} for option in options]
+        repr_options = [(repr(option['option']), repr(option['value']))
```
solarissmoke, line 368: I'd prefer not to change the content of what is being hashed, to reduce the risk of breaking an existing hash. I think we can make the existing hash values deterministic by doing this instead:

```
repr_options.sort(key=lambda a: a['option'])
```

solarissmoke (changes requested):
Thanks for this! One small change requested to the approach for fixing this.

solarissmoke (approved):
Thanks!

PR #3692, opened 2021-04-19
Add support for django 3.2

robalford:
This updates Oscar's requirements to support Django 3.2 and fixes a test failure that appears related to [this change](https://docs.djangoproject.com/en/3.2/ref/contrib/messages/#django.contrib.messages.storage.cookie.CookieStorage) to the Cookie storage hash format by using the new in 3.2 [`Signer.unsign_object()`](https://docs.djangoproject.com/en/3.2/topics/signing/#signing-complex-data) method to retrieve the human-readable message for the test assertion.

### DIFF
--- a/docs/source/releases/v3.0.rst
+++ b/docs/source/releases/v3.0.rst
@@ -19,7 +19,7 @@ also need to upgrade their frontend and dashboard templates to use Bootstrap 4.
 Compatibility
 ~~~~~~~~~~~~~
 
-Oscar 3.0 is compatible with Django 2.2, Django 3.0 and Django 3.1 as well as Python 3.6, 3.7 and 3.8.
+Oscar 3.0 is compatible with Django 2.2, Django 3.0, 3.1 and 3.2 as well as Python 3.6, 3.7 and 3.8.
 
 Support for Django 1.11 and Django 2.1 has been dropped. Support for Python 3.5 has been dropped.
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -16,7 +16,7 @@ jobs:
       fail-fast: true
       matrix:
         python-version: [3.6, 3.7, 3.8, 3.9]
-        django-version: [2.2, 3.0, 3.1]
+        django-version: [2.2, 3.0, 3.1, 3.2]
     services:
       postgres:
         image: postgres:10
--- a/setup.py
+++ b/setup.py
@@ -18,7 +18,7 @@ sys.path.append(os.path.join(PROJECT_DIR, 'src'))
 from oscar import get_version  # noqa isort:skip
 
 install_requires = [
-    'django>=2.2,<3.2',
+    'django>=2.2,<3.3',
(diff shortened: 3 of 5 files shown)


docs/source/releases/v3.0.rst
```diff
@@ -19,7 +19,7 @@ also need to upgrade their frontend and dashboard templates to use Bootstrap 4.
 Compatibility
 ~~~~~~~~~~~~~
 
-Oscar 3.0 is compatible with Django 2.2, Django 3.0 and Django 3.1 as well as Python 3.6, 3.7 and 3.8.
```
solarissmoke, line 22: Please revert the changes to 3.0 release notes (3.0 has already shipped). This change will ship in Oscar 3.1 and we'll update the notes for 3.1 before release (since we'll also be dropping support for Django 3.0 which is now end of life).

solarissmoke (changes requested):
Thanks for this! Looking at the [test log](https://github.com/django-oscar/django-oscar/pull/3692/checks?check_run_id=2383920103) for Django 3.2, it looks like there are a number of deprecation warnings which we need to address (if we don't projects are going to get a lot of warnings that they can't do anything about):

```
RemovedInDjango40Warning: The DEFAULT_HASHING_ALGORITHM transitional setting is deprecated. Support for it and tokens, cookies, sessions, and signatures that use SHA-1 hashing algorithm will be removed in Django 4.0.
```
This is just a test setting which can be safely removed.

```
DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
```
Looks like we need to switch to importlib.

```
RemovedInDjango40Warning: request.is_ajax() is deprecated. See Django 3.1 release notes for more details about this deprecation.
    if self.request.is_ajax():
```
We're going to have to provide our own implementation of this, as suggested in the Django [release notes](https://docs.djangoproject.com/en/3.2/releases/3.1/#id2). 

```
RemovedInDjango40Warning: Passing None for the middleware get_response argument is deprecated.
    super().__init__(get_response)
```
Should be a simple change.

```
RemovedInDjango41Warning: 'oscar.apps.customer' defines default_app_config = 'oscar.apps.customer.apps.CustomerConfig'. However, Django's automatic detection did not find this configuration. You should move the default config class to the apps submodule of your application and, if this module defines several config classes, mark the default one with default = True.
```
Should also be a simple change.

I don't know if you have the time/capacity to address these - if you do then it would be greatly appreciated.

robalford:
@solarissmoke no prob, I should have the capacity to address the deprecation warnings in the next week or so if that works for your release timeline.

docs/source/releases/v3.0.rst
robalford: Will do

solarissmoke:
@robalford that would be great, thank you.

solarissmoke:
@robalford I've incorporated your commits into #3704, where I've worked on the deprecation issues (we'd like to get a release out soon). Thanks for your help!

Issue #3693, opened 2021-04-20
Missing translation tag in e-mails


Anto59290:
### Issue Summary

When using Oscar in another language, some texts in the e-mails are not translated.

### Steps to Reproduce

- Configure oscar in any other language than English
- Make oscar send the reset password email
- You can see that in the email the CTA is still "Reset Password"

### Technical details

Source of the issue: 

https://github.com/django-oscar/django-oscar/blob/master/src/oscar/templates/oscar/communication/emails/commtype_password_changed_body.html#L20

This code is missing a "trans" tag.

Same issue in a few other places:

https://github.com/django-oscar/django-oscar/blob/master/src/oscar/templates/oscar/communication/emails/commtype_email_changed_body.html#L20

https://github.com/django-oscar/django-oscar/blob/1442cb734ec542459a31cbd9c174bf8817dcfeea/src/oscar/templates/oscar/communication/emails/commtype_password_reset_body.html#L20

solarissmoke:
Thanks - PR to fix this would be appreciated if you have time.

solarissmoke:
Fixed in https://github.com/django-oscar/django-oscar/commit/04cd6a4fc750db9310147d96776f06fe289269bf

PR #3694, opened 2021-04-20
Carousel's navigation vertical placement

krzysztofjeziorny:
Carousel's navigation more optimal placement should be inside of the `div.carousel-inner` to correct its position to the vertical middle of the main image, not of the whole gallery (constructed with the main image and thumbnails).

### DIFF
--- a/src/oscar/templates/oscar/catalogue/partials/gallery.html
+++ b/src/oscar/templates/oscar/catalogue/partials/gallery.html
@@ -6,7 +6,7 @@
 
     {% if all_images|length > 1 %}
 
-        <div id="product_gallery" class="carousel slide">
+        <div id="product_gallery" class="carousel slide" data-ride="carousel">
 
             <div class="img-thumbnail mb-3">
                 <div class="carousel-inner">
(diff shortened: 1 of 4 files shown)


solarissmoke:
> Carousel's navigation more optimal placement should be inside of the `div.carousel-inner`.

Hi @krzysztofjeziorny , can you point to an example in the Boostrap documentation that does this? All the [examples](https://getbootstrap.com/docs/4.6/components/carousel/#with-controls) I can find have the indicators *outside* the `.carousel-inner` block, so I'm fairly sure our existing implementation is correct. It is possible that Oscar has added custom styling that alters how this looks, but in that case we need to fix the styling.

krzysztofjeziorny:
@solarissmoke You are right that Bootstrap puts the navigation outside of the image block, however Oscar introduced the thumbnails below it, and so stretching the gallery vertically.

Moving the image arrows containers into the `.carousel-inner` don't have any side effects I can find, and as they are `flexbox` containers themselves, their height is being stretched to 100% height of the main image box, `align-items` and `justify-content` positioning the arrows in the center of them.

Alternatively one have to restyle the gallery to not include the thumbnails in the carousel layout. I implemented my easy fix without problems, but I understand it need to be tested more.

krzysztofjeziorny:
That went wrong, was intended for #3693

solarissmoke:
Thanks, I see what you mean now. Looking at it I think it may be better to move the thumbnails outside of the carousel element - we're currently abusing the `carousel-indicators` class there, and then having to override all its styling.

PR #3704, opened 2021-04-30
Add support for django 3.2 and drop support for Django 3.0

solarissmoke:
We can't drop `DEFAULT_HASHING_ALGORITHM` from the tests until we drop support for Django 2.2.


jwayodi (approved):
This looks good.

PR #3705, opened 2021-04-30
Fix dashboard filtering of inactive/active offers without start/end dates

solarissmoke:
The dashboard filtering doesn't work as expected:

1. Filtering on "is active" does not include offers without a start or end date - only offers with both start/end dates will appear (assuming the current date is within that range).
2. Filtering on "is inactive" does not include offers whose start date is in the future.


jwayodi (approved):
This looks good.

PR #3706, opened 2021-05-04
Prepare 3.1 beta release:

solarissmoke:
- Remove deprecated features and update deprecation notices for next release.
- Update release notes.
- Bump version


setup.py
```diff
@@ -18,15 +18,15 @@
...
+    'django>=2.2,<3.2',
     # PIL is required for image fields, Pillow is the "friendly" PIL fork
     'pillow>=6.0',
     # We use the ModelFormSetView from django-extra-views for the basket page
     'django-extra-views>=0.13,<0.14',
     # Search support
     'django-haystack>=3.0b1',
     # Treebeard is used for categories
-    'django-treebeard>=4.3,<4.5',
+    'django-treebeard>=4.3',
```
solarissmoke, line 29: This was pinned temporarily because of a breaking change in treebeard, which has since been fixed.

src/oscar/core/loading.py
```diff
@@ -239,7 +239,7 @@ def get_model(app_label, model_name):
             warnings.warn(
                 'Model %s has recently moved from %s to the application %s, '
                 'please update your imports.' % (model_name, original_app_label, app_label),
-                RemovedInOscar31Warning, stacklevel=2)
+                RemovedInOscar32Warning, stacklevel=2)
```
jwayodi, line 242: The whole `if oscar_moved_model` block this is in, was [added](https://github.com/django-oscar/django-oscar/commit/2af6753c6733198d66519becd78e62f8af0b2cf0) with the `communication` app. The affected models were [removed](https://github.com/django-oscar/django-oscar/commit/c30de9d008bb11ae80e230f156111726b4c24d00) from `MOVED_MODELS` with the release of 3.0 alpha, but this code was left behind. Maybe we should remove it completely?

jwayodi (commented):
`oscar.apps.basket.abstract_models.AbstractLine.is_available_for_discount` has been deprecated (in favour of `oscar.apps.basket.abstract_models.AbstractLine.is_available_for_offer_discount`) [for a long time](https://github.com/django-oscar/django-oscar/commit/78823ca0445eccf0db88fa9a0bd0d9e2a89f67e2#diff-c269280e2151233bdc937d009eefb03861b19eb0915aa09193ad77c1b6f25d64R776), but has not been properly marked as such. Could we remove it now, or at least mark it as deprecated?

src/oscar/core/loading.py
solarissmoke: We discussed this at the time and decided to leave this code here in case we need to move other models in future.

solarissmoke:
Yeah, let's remove `is_available_for_discount`

jwayodi (approved):
This looks good.

PR #3729, opened 2021-07-06
Issue/publiccategoriesonly

xplsek03:
Shouldn't be categories for templates selected from those which are marked as public only?


solarissmoke:
Yes, think this makes sense. Grateful if you can rebase the PR onto master, and remove the unrelated commits.

PR #3739, opened 2021-07-23
using public categories only

xplsek03:
Fixed original PR:
https://github.com/django-oscar/django-oscar/pull/3729

### DIFF
--- a/src/oscar/apps/catalogue/abstract_models.py
+++ b/src/oscar/apps/catalogue/abstract_models.py
@@ -659,12 +659,12 @@ class AbstractProduct(models.Model):
 
     def get_categories(self):
         """
-        Return a product's categories or parent's if there is a parent product.
+        Return a product's public categories or parent's if there is a parent product.
         """
         if self.is_child:
-            return self.parent.categories.filter(is_public=True)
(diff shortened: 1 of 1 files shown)


src/oscar/apps/catalogue/abstract_models.py
```diff
@@ -662,9 +662,9 @@ def get_categories(self):
         Return a product's categories or parent's if there is a parent product.
         """
         if self.is_child:
-            return self.parent.categories
+            return self.parent.categories.filter(is_public=True)
```
solarissmoke, line 665: Please use the `browsable()` queryset instead - i.e., `self.parent.categories.browsable()`, which correctly filters out categories with non-public ancestors (`is_public` alone doesn't).

src/oscar/apps/catalogue/abstract_models.py
```diff
@@ -662,9 +662,9 @@ def get_categories(self):
         Return a product's categories or parent's if there is a parent product.
```
solarissmoke, line 662: Please amend this comment to read "Return a product's *public* categories..."

solarissmoke (changes requested):

PR #3741, opened 2021-07-28
[#3740] Add is_ajax helper to utils.py

### LINKED ISSUE #3740: Add is_ajax helper to utils.py
Django 3.1 deprecated the `HttpRequest.is_ajax` method as it relies on a jQuery-specific way of signifying AJAX calls.

For most Oscar projects that's probably ok though, as they only use jQuery.

I see that in the current codebase `self.request.is_ajax()` was replaced by the less readable
```
self.request.headers.get('x-requested-with') == 'XMLHttpRequest'
```

I suggest adding a helper to `core/compat.py` to restore the old behaviour. Apart from readability that would enable oscar projects to check for AJAX requests in a consistent way with the built-in code.

I'm thinking of
```
def is_ajax(request):
    """
    Django 3.1 deprecated the  HttpRequest.is_ajax method as it relies on
    a jQuery-specific way of signifying AJAX calls.

    https://docs.djangoproject.com/en/3.1/releases/3.1/#id2
    """
    return request.META.get('HTTP_X_REQUESTED_WITH') == 'XMLHttpRequest'
```

I'm happy to submit a PR for this ✌️

SaturnFromTitan:
closes #3740

I'd also want to add a release note, but wasn't sure where to put it. the latest release note file is for 3.1, but that one seems to be already out.

### DIFF
--- a/docs/source/releases/v3.2.rst
+++ b/docs/source/releases/v3.2.rst
@@ -34,7 +34,8 @@ Removal of deprecated features
 Minor changes
 ~~~~~~~~~~~~~
 
-
+- Added a new helper ``core.utils.is_ajax`` which replicates the logic of Django's ``HttpRequest.is_ajax``
+  method that was deprecated in Django 3.1.
 
 .. _dependency_changes_in_3.2:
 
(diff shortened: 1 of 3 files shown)


solarissmoke:
@SaturnFromTitan I've added stub release notes for the next release [here](https://github.com/django-oscar/django-oscar/blob/master/docs/source/releases/v3.2.rst) - grateful if you can add a note about this in the "minor changes" section.

SaturnFromTitan:
I am surprised that the test coverage goes down. Is there no test coverage for the `BasketView`?

I could add a unit test, but that feels pretty redundant I'd have to replicate the implementation more or less directly.

jwayodi (approved):
This look good to me.

PR #3743, opened 2021-07-31
Fix output of OfferReportGenerator and generate its data more efficiently

solarissmoke:
The "offer performance" report has been broken since the changes to the `ReportGenerator` introduced in Oscar 3.0 . 

This fixes the `OfferReportGenerator` to work with those changes, and also reimplements the data processing itself to do it via a database query instead of iterating over all the `OrderDiscount` objects. It also fixes the HTML template not to link to deleted offers, and orders the report by discount value.

### DIFF
--- a/src/oscar/apps/offer/reports.py
+++ b/src/oscar/apps/offer/reports.py
@@ -1,5 +1,5 @@
-from decimal import Decimal as D
-
+from django.db.models import OuterRef, Subquery, Sum
+from django.db.models.functions import Coalesce
 from django.utils.translation import gettext_lazy as _
 
 from oscar.core.loading import get_class, get_model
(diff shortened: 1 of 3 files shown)


solarissmoke:
Turns out the conditional lookup only works in Django 3.2. I'll see if I can find a solution that works in <3.1.

src/oscar/apps/offer/reports.py
```diff
@@ -42,20 +41,16 @@ class OfferReportGenerator(ReportGenerator):
...
-                }
-            offer_discounts[discount.offer_id]['total_discount'] \
-                += discount.amount
-
-        return self.formatter.generate_response(list(offer_discounts.values()))
+    def get_queryset(self):
+        offers = ConditionalOffer.objects.filter(pk=OuterRef("offer_id"))
+        return super().get_queryset().order_by().values("offer_id").annotate(
+            total_discount=Sum("amount"),
+            offer=Subquery(offers.values("pk")[:1]),
```
jwayodi, line 48: It would help to have a comment saying this is used in the template, to determine whether to add a link to the offer.

jwayodi (dismissed):
Just a minor comment, otherwise this looks good.

PR #3748, opened 2021-08-03
Make CSV order columns overridable

ajharry69:
This change will help make dependant projects easily add extra columns to the generated CSV.


src/oscar/apps/dashboard/orders/views.py
```diff
@@ -356,43 +367,40 @@ def render_to_response(self, context, **response_kwargs):
     def get_download_filename(self, request):
         return 'orders.csv'
 
+    @cached_property
+    def sorted_csv_columns(self):
+        columns = OrderedDict()
```
solarissmoke, line 372: This method can be simplified to just:

```
return OrderedDict(self.CSV_COLUMNS)
```
which is simple enough that it may be easier just to do that where we need it, directly.

src/oscar/apps/dashboard/orders/views.py
```diff
@@ -356,43 +367,40 @@ def render_to_response(self, context, **response_kwargs):
...
+        for k, v in self.CSV_COLUMNS:
+            columns[k] = v
+        return columns
+
+    def get_row_values(self, order):
+        row = self.sorted_csv_columns.copy()
```
solarissmoke, line 378: I know the old code was doing this, but I think it's confusing to initialise the `row` with incorrect (heading) values. I think this method should return a regular dict - we should then order them before rendering the row (by looping over the columns).

solarissmoke: (This should also mean we only have to read `self.CSV_COLUMNS` once, in the code that generates the CSV.

solarissmoke (changes requested):

solarissmoke (approved):

PR #3750, opened 2021-08-05
Drop support for Django 2.2 and Python 3.6

solarissmoke:
Both Django 2.2 and Python 3.6 reach end of life in December 2021 - which is around the time we will be shipping the next release of Oscar.

Dropping support for Python 3.6 also means we can remove all uses of `collections.OrderedDict`, because dictionaries are ordered in Python 3.7.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
-FROM python:3.7
+FROM python:3.8
 ENV PYTHONUNBUFFERED 1
 
-RUN curl -sL https://deb.nodesource.com/setup_12.x | bash -
+RUN curl -sL https://deb.nodesource.com/setup_14.x | bash -
 RUN apt-get install -y nodejs
 
 COPY ./requirements.txt /requirements.txt
--- a/docs/source/releases/v3.2.rst
+++ b/docs/source/releases/v3.2.rst
@@ -14,6 +14,9 @@ Oscar 3.2 release notes (in development)
 Compatibility
 ~~~~~~~~~~~~~
 
+Oscar 3.2 is compatible with Django 3.1 and Django 3.2 and Python versions 3.7 to 3.9.
+
+Support for Django 2.2 has been dropped. Support for Python 3.6 has been dropped.
 
 .. _new_in_3.2:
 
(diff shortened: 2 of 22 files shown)


docs/source/releases/v3.2.rst
```diff
@@ -14,6 +14,9 @@ Oscar 3.2 release notes (in development)
 Compatibility
 ~~~~~~~~~~~~~
 
+Oscar 3.2 is compatible with Django Django 3.1 and Django 3.2 and Python versions 3.7 to 3.9.
```
jwayodi, line 17: There is a typo ("Django Django") here.

jwayodi:
In `Dockerfile`, `FROM python:3.7` should be `FROM python:3.8`?

src/oscar/templatetags/category_tags.py
```diff
@@ -67,8 +67,8 @@ def __iter__(self):
         yield self
 
 
-@register.simple_tag(name="category_tree")   # noqa: C901 too complex
-def get_annotated_list(depth=None, parent=None):
+@register.simple_tag(name="category_tree")
+def get_annotated_list(depth=None, parent=None):    # noqa: C901 too complex
```
jwayodi, line 71: This is because of switching to Python 3.8, eh? I could not find it documented anywhere.

jwayodi:
@solarissmoke A few minor comments, otherwise this looks good.

jwayodi (approved):

src/oscar/templatetags/category_tags.py
solarissmoke: Yeah - I couldn't find anything either but how isort checks decorated functions seems to change in Python 3.8 .

solarissmoke: flake8, rather

Dockerfile
```diff
@@ -1,7 +1,7 @@
-FROM python:3.7
+FROM python:3.8
 ENV PYTHONUNBUFFERED 1
 
-RUN curl -sL https://deb.nodesource.com/setup_12.x | bash -
+RUN curl -sL https://deb.nodesource.com/setup_14.x | bash -
```
solarissmoke, line 4: Bumping to latest LTS of node.

jwayodi (approved):

PR #3753, opened 2021-08-09
Support Django 3.2 automatic "AppConfig" discovery

jwayodi:
Importing "oscar.core.application.OscarConfig" directly into an app's
"apps.py" results in Django 3.2. finding multiple "AppConfig"
subclasses, and requiring us to specify the default one.

### DIFF
--- a/docs/source/howto/how_to_disable_an_app_or_feature.rst
+++ b/docs/source/howto/how_to_disable_an_app_or_feature.rst
@@ -25,7 +25,6 @@ the dashboard app::
 
     # myproject/config.py
     from oscar.config import Shop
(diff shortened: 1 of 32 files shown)


solarissmoke:
Having discussed this, and raised the underlying issue with the [Django project team](https://code.djangoproject.com/ticket/33013), it seems there isn't going to be an ideal solution to this. I feel like changing the imports in this way is unlikely to "stick" as a solution, because it's not obvious why it's been done, and is also just very non-obvious.

Suggest that we just document the problem, and recommend that when forking an app projects always use explicit paths to their app config in `INSTALLED_APPS`.

PR #3756, opened 2021-08-26
Avoid KeyError in BulkEditMixin if objects matching filters no longer exist

solarissmoke:
This small changes ensure that we don't cause an unhandled `KeyError` in views subclassing `BulkEditMixin` when they try to access objects that no longer match the filters applied in `get_queryset()` . An example of when this happens is when bulk editing orders in the dashboard, when they are filtered on status. If one of those orders had it's status changed independently since the view was loaded, a `KeyError` is raised when you try to submit a bulk status changed.

### DIFF
--- a/src/oscar/views/generic.py
+++ b/src/oscar/views/generic.py
@@ -8,7 +8,7 @@ from django.views.generic.base import View
 from oscar.core.utils import safe_referrer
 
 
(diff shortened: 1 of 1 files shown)


jwayodi (approved):
This looks good.

PR #3757, opened 2021-08-30
Document properly specifying forked app in "INSTALLED_APPS"

jwayodi:


docs/source/topics/customisation.rst
```diff
@@ -95,6 +95,20 @@ with that for your own app.
             'yourappsfolder.dashboard.reports.apps.ReportsDashboardConfig',
         ]
 
+.. note::
+
+    You will need to use the dotted Python path to the app config class, rather
```
solarissmoke, line 100: Suggest we say "it is recommended to use" instead of "you will need to use", because it is still possible to make automatic discovery work by messing around with imports.

jwayodi: @solarissmoke Fixed.

solarissmoke (approved):

PR #3765, opened 2021-09-06
Add pagination to RangeDetailView

### LINKED ISSUE #3731: RangeDetailView should have pagination
The page RangeDetailView (oscar/offer/range.html) should show products in pagination.
Because there might be too many products in a Range

SakshiUppoor:
Closes #3731


SakshiUppoor:
This seems like a small PR
If this solves the issue, should I add any tests in this PR for the same?

solarissmoke (approved):
Thanks for this! I don't think tests are necessary for this change.

PR #3777, opened 2021-09-29
Remove title field from user address form

### LINKED ISSUE #3696: Person's title translations giving confusing choices in other languages
The same translations of the English terms in address book's dropdown list of titles are causing confusion in other languages.

The original choices in English: 

https://github.com/django-oscar/django-oscar/blob/1a3372e1728d7b464693da6c78525ef4f13176a1/src/oscar/apps/address/abstract_models.py#L22

are being translated to a same term:  Miss/Mrs/Ms turns Frau/Frau/Frau in German or Mlle/Mme/Mme in French or Panna/Pani/Pani in Polish. It's impossible to differentiate the title chosen.

https://github.com/django-oscar/django-oscar/blob/1a3372e1728d7b464693da6c78525ef4f13176a1/src/oscar/locale/de/LC_MESSAGES/django.po#L45-L55

Screenshot 2021-04-28 at 11 02 43

One can try it out with different languages in the address book in the [sandbox](https://latest.oscarcommerce.com/en-gb/accounts/addresses/add/).

I'd propose, instead of searching for direct translations, that could be difficult or awkward to use ("panna" in Polish or "Fräulein" in German for Miss are rather old-fashioned and not being used often), we leave just one term for a woman, hopefully it's sensible and inclusive to all in question.

SakshiUppoor:
Fixes #3696
- Passed all tests in [integration\address\test_models.py](https://github.com/django-oscar/django-oscar/blob/master/tests/integration/address/test_models.py) and [integration\address\test_forms.py](https://github.com/django-oscar/django-oscar/blob/master/tests/integration/address/test_forms.py)

### DIFF
--- a/src/oscar/apps/address/abstract_models.py
+++ b/src/oscar/apps/address/abstract_models.py
@@ -323,7 +323,7 @@ class AbstractAddress(models.Model):
         """
         return self.join_fields(
             ('title', 'first_name', 'last_name'),
-            separator=" ")
+            separator=" ").strip()
 
     @property
     def name(self):
--- a/src/oscar/apps/address/forms.py
+++ b/src/oscar/apps/address/forms.py
@@ -25,7 +25,7 @@ class UserAddressForm(PhoneNumberMixin, AbstractAddressForm):
     class Meta:
         model = UserAddress
         fields = [
-            'title', 'first_name', 'last_name',
+            'first_name', 'last_name',
             'line1', 'line2', 'line3', 'line4',
             'state', 'postcode', 'country',
             'phone_number', 'notes',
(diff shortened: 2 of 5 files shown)


solarissmoke (changes requested):
I think we also need to remove the title from `checkout.forms.ShippingAddressForm`, and to adjust the `salutation` property to strip extraneous white space if the `title` is blank.

SakshiUppoor:
On it.

I also found `title` field in `dashboard.orders.forms.ShippingAddressForm` and `payment.forms.BillingAddressForm`. So removing these as well.

solarissmoke (approved):
Thanks!

PR #3786, opened 2021-10-14
Add unique=True to ProductAlert.key 

odunybrad:
This PR makes the following changes:
* Add unique=True to AbstractProductAlert.key  in model. 
* MIgration file.

The `AbstractProductAlert` has a field `key` to be a random string of 40 characters in `get_random_key` when save. 
It's better to enforce the `unique=True` constraints on this field.


specialunderwear (changes requested):
Like solarissmoke explained, the migration should deal with existing non unique values and make sure they are all uniquer before applying the constraint.

PR #3794, opened 2021-10-27
Fix partner product filter queryset in order to allow partner users access their products with variations.

sasha0:


solarissmoke:
@sasha0 my only concern here is that this potentially changes edit permissions - it would now mean that a partner can edit a parent product (that could also affect other partners), where they couldn't previously. Do you think there are any concerns with that? I think we need to be clear about what Oscar's default is in this respect.

sasha0:
If the standalone product has multiple stockrecords, belonging to the different partners - each of the partners can already edit it, so I don't think there's not much difference, allowing partners edit the parent products. From the other side, if the admin user created a product with the stockrecords, belonging to the partner, he/she won't be able to edit it, which one of my clients already faced.

PR #3795, opened 2021-10-27
Predefined product options

specialunderwear:
This looks something like this:

image


src/oscar/apps/basket/forms.py
```diff
@@ -293,9 +350,11 @@ def cleaned_options(self):
         options = []
         for option in self.parent_product.options:
             if option.code in self.cleaned_data:
-                options.append({
-                    'option': option,
-                    'value': self.cleaned_data[option.code]})
+                value = self.cleaned_data[option.code]
+                if option.required or value not in EMPTY_VALUES:
+                    options.append(
+                        {"option": option, "value": option.clean_value(value)}
```
solarissmoke, line 356: This is also not a backward compatible change - previously we would pass a string value here and this is no longer the case. A value `"foo"` now gets JSON encoded to `'"foo"'` - which would break the logic of any project that is processing option values assuming they are simple strings. It also somewhat messes up the display of the line in the basket etc. 

I think it would be better to JSON-encode only values that are not simple. 

We also need to make sure option/multi option values are displayed in a sensible way in the basket, by default.

src/oscar/apps/catalogue/abstract_models.py
```diff
@@ -1228,23 +1230,91 @@ class AbstractOption(models.Model):
...
     )
 
+    empty_label = "------"
+    empty_radio_label = _("Skip this option")
+
     name = models.CharField(_("Name"), max_length=128, db_index=True)
     code = AutoSlugField(_("Code"), max_length=128, unique=True, populate_from='name')
     type = models.CharField(_("Type"), max_length=255, default=TEXT, choices=TYPE_CHOICES)
     required = models.BooleanField(_("Is this option required?"), default=False)
+    option_group = models.ForeignKey(
```
solarissmoke, line 1257: I think it is worth adding validation to ensure this value is not populated unless the type is option or multi-option, and that it is not left empty otherwise.

src/oscar/apps/catalogue/abstract_models.py
```diff
@@ -1228,23 +1230,91 @@ class AbstractOption(models.Model):
...
+        "catalogue.AttributeOptionGroup",
+        blank=True,
+        null=True,
+        on_delete=models.CASCADE,
+        related_name="product_options",
+        verbose_name=_("Option Group"),
+        help_text=_('Select an option group if using type "Option" or "Multi Option"'),
+    )
+    help_text = models.CharField(
```
solarissmoke, line 1266: Please add help text to this field (lol) along the lines of "Help text visible to the customer in the add-to-basket form".

solarissmoke (changes requested):

src/oscar/apps/basket/forms.py
specialunderwear: I'm not too sure about having special cases for simple values is a good idea, especially because they where not handled correctly to begin with. If the option type is int and I receive a string from the database that is not correct behaviour.

So what could be done here, is make the breaking change more apparent by introducing a migration and changing the field type to something that encodes to json before storage. We could then also have a data migration that migrates all existing data. The code that treats string values as simple string would then still work, only when you where treating integers as simple string you would have a problem, which seems correct to me.

How does that sound to you @solarissmoke

src/oscar/apps/catalogue/abstract_models.py
specialunderwear: Will do

src/oscar/apps/catalogue/abstract_models.py
```diff
@@ -1228,23 +1230,91 @@ class AbstractOption(models.Model):
...
+        verbose_name=_("Help text"), blank=True, null=True, max_length=255
+    )
+    order = models.IntegerField(
```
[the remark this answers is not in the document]
specialunderwear: I'm sorry that would be a lot of hassle on our side because we already use this code and I'd have to introduce another migration that doesn't really change anything. So since your request does not really solve a problem and just introduce limitations, I'd like to keep it like this.

src/oscar/apps/basket/forms.py
solarissmoke: Sounds good.

src/oscar/apps/catalogue/abstract_models.py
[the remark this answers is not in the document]
solarissmoke: OK

src/oscar/apps/basket/forms.py
solarissmoke: (Still think we need to improve the default display of these in the basket - rendering a JSON blob as string doesn't feel like a usable default)

src/oscar/apps/basket/forms.py
```diff
@@ -13,24 +14,76 @@
 Product = get_model('catalogue', 'product')
 
 
-def _option_text_field(option):
-    return forms.CharField(label=option.name, required=option.required)
+# pylint: disable=unused-argument
```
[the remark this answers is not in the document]
specialunderwear: we should!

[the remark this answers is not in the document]
specialunderwear: And black!

src/oscar/apps/basket/forms.py
```diff
@@ -213,7 +270,7 @@ def _add_option_field(self, product, option):
         This is designed to be overridden so that specific widgets can be used
         for certain types of options.
         """
-        option_field = self.OPTION_FIELD_FACTORIES.get(option.type, _option_text_field)(option)
+        option_field = self.OPTION_FIELD_FACTORIES.get(option.type, _option_text_field)(self, product, option)
```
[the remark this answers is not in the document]
specialunderwear: done

src/oscar/apps/catalogue/abstract_models.py
specialunderwear: done

src/oscar/apps/basket/forms.py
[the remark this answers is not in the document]
solarissmoke: Happy to add black and pylint, but let's do that as a separate change?

src/oscar/apps/catalogue/abstract_models.py
specialunderwear: done

alexsabb:
Just found this! :D and will be following the state of this PR - as this is --- naturally --- exactly what I was looking to implement on www.simple3dprints.com

Will be rolling to 3.2 when this comes out

src/oscar/apps/catalogue/abstract_models.py
```diff
@@ -1228,23 +1230,104 @@ class AbstractOption(models.Model):
...
+        if not self.required:
+            choices = self.add_empty_choice(choices)
+
+        return choices
+
+    def clean_value(self, value):
+        return json.dumps(value, cls=DjangoJSONEncoder)
+
+    def clean(self):
+        if self.type in [self.DATE, self.SELECT, self.MULTI_SELECT, self.CHECKBOX]:
```
alexsabb, line 1316: I had to add self.RADIO to this array and then the radio option seemed to work correctly when I manually tested it - 

this is my first github 'review' btw - hope I'm not stepping on anyones toes or reading the source incorrectly -

alexsabb (commented):
Th

src/oscar/apps/catalogue/abstract_models.py
specialunderwear: ha yeah and self.DATE should not be there.

src/oscar/apps/basket/forms.py
specialunderwear: So I just implemented the JSONField but I'm not sure if I mind the JSON blob as much as you do 😃 
That will take some template changes and I'm not really up for that right now.

src/oscar/apps/catalogue/abstract_models.py
specialunderwear: good catch, self.DATE does not belong there, I guess I added the wrong value.

src/oscar/apps/basket/forms.py
solarissmoke: You mention above that we'd need a data migration to migrate existing (CharField) data to the JSONField - I don't see that data migration in the PR? 

We also need to include a prominent note about this in the release notes. 

I still feel that the change in how these options are displayed by default is problematic, and that if we are going to improve this functionality then it is worth the effort of sorting out the templates at the same time. Whether or not that should block merging of the functional changes is a matter of opinion and I expect that you will argue it's better to merge and improve later. I just think Oscar needs to to better at shipping usable defaults.

specialunderwear: The data migration is not needed because the existing data in the charfield is valid json. That is why all the tests still pass as well.

We will have an intern mess around with the templates tomorrow @viggo-devries will make that happen.

specialunderwear: The display of the options has been fixed.

viggo-devries (approved):

joeyjurjens (approved):

solarissmoke (approved):
Looks good. @viggo-devries , we will need to clearly document the change to JSONField in the release notes.

viggo-devries:
> Looks good. @viggo-devries , we will need to clearly document the change to JSONField in the release notes.

Yeah definitely!

PR #3808, opened 2021-11-09
Fix browsable recommended products in product detail

xplsek03:
I felt like it might be useful to create separate QuerySet method `browsable_dashboard_recommended` for it.. it might seem little duplicate with `public` but for my scenario it was useful.

### DIFF
--- a/src/oscar/apps/dashboard/catalogue/views.py
+++ b/src/oscar/apps/dashboard/catalogue/views.py
@@ -636,7 +636,7 @@ class ProductLookupView(ObjectLookupView):
     model = Product
 
     def get_queryset(self):
-        return self.model.objects.browsable().all()
+        return self.model.objects.public()
 
     def lookup_filter(self, qs, term):
         return qs.filter(Q(title__icontains=term)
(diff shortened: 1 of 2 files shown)


src/oscar/apps/dashboard/catalogue/views.py
```diff
@@ -636,7 +636,7 @@ class ProductLookupView(ObjectLookupView):
     model = Product
 
     def get_queryset(self):
-        return self.model.objects.browsable().all()
+        return self.model.objects.browsable_dashboard_recommended().all()
```
solarissmoke, line 639: `ProductLookupView` is a generic view that can be used by the `ProductSelect` and `ProductSelectMultiple` widgets in any context. It is not specific to recommended products and so this needs to remain generic. I think it's best just to use `model.objects.public()`.

solarissmoke (changes requested):

src/oscar/apps/dashboard/catalogue/views.py
xplsek03: If its like this, ok. I broke something so I will start new PR with it.

xplsek03:
Something broke, created new PR: https://github.com/django-oscar/django-oscar/pull/3809

PR #3809, opened 2021-11-10
add public products to ProductLookupView queryset

xplsek03:


specialunderwear:
Could you please add 1 test demonstrating that currently this functionality is broken, and you have fixed it?

xplsek03:
> Could you please add 1 test demonstrating that currently this functionality is broken, and you have fixed it?

100%, just gotta take some time, very busy now

specialunderwear (changes requested):
please add tests demonstrating the issue

samar-hassan:
@xplsek03  Can you please explain the reason for adding each variant individually to the recommended products dropdown list? Currently we just show the parent or standalone products. Although for parent products we are redirected to a separate page and there we have to select the variant to add to our basket. I guess you may want to avoid such behavior and display just a specific variant in the recommendations so that it can be added without additional steps.

xplsek03:
> @xplsek03 Can you please explain the reason for adding each variant individually to the recommended products dropdown list? Currently we just show the parent or standalone products. Although for parent products we are redirected to a separate page and there we have to select the variant to add to our basket. I guess you may want to avoid such behavior and display just a specific variant in the recommendations so that it can be added without additional steps.

Overflown with another projects. Will have to get to oscar again for next proj.

PR #3823, opened 2021-12-07
Make attribute codes unique per product class

### LINKED ISSUE #3785: ProductAttribute.code should be a unique field
### Issue Summary

The product attribute should have a unique code. 
You can only assign one attribute with that code in the dashboard-product anyway. 

### Steps to Reproduce

1. In dashboard, go to `product-type` - `product attributes` page: _/dashboard/catalogue/product-type/1/update/_
2. Add two product attributes with the same name. (I use name att1 and type text.)
3. Save the form. Both attributes with the same code are saved and shown on the page. 
4. Go to one `dashboard/products` page for that product class, _/dashboard/catalogue/products/8/?upc=&title=shirt_, only one attribute name is shown on the form. 
5. Note that even I create attributes with different types, only one attribute of that code will be shown on the `dashboard/products` page. 

Screen Shot 2021-10-14 at 11 30 32 AM

Screen Shot 2021-10-14 at 11 30 14 AM

#### Expected result:
* Give Exception when saving the attribute - "Duplicate product attributes code / attributes with this code already exists." 
* Enforce the `uniqueness` of `AbstractProductAttribute.code`. This is a SlugField. 
* Another way is to enforce the unique_together of (code, type), while the data retrieval part needs to be updated as well. 
* I can work on the PR for this issue, please let me know. 

#### Cause
Here have the assumption that `attribute.code` should be unique. 

```python
File: src/oscar/apps/dashboard/catalogue/forms.py
    def set_initial_attribute_values(self, product_class, kwargs):
        ...
        for attribute in product_class.attributes.all():
            try:
                value = instance.attribute_values.get(
                    attribute=attribute).value
            except exceptions.ObjectDoesNotExist:
                pass
            else:
                kwargs['initial']['attr_%s' % attribute.code] = value
```

### Technical details

* Python version: 3.9.7
* Django version: 3.2.7
* Oscar version: master

joeyjurjens:
Fixes #3785


specialunderwear:
Hi @joeyjurjens please add the missing migration ok?

joeyjurjens:
> Hi @joeyjurjens please add the missing migration ok?

I have added the migration!

specialunderwear:
Ok so what happens when sites have multiple attributes with the same code on a productclass?

joeyjurjens:
> Ok so what happens when sites have multiple attributes with the same code on a productclass?

When adding the same code in dashboard: It will show an error message with the error: "Please correct the duplicate data for code."
When migrating this migration: It will raise a UNIQUE constraint error. Should I create a forward function that removes the duplicate attribute and move all attribute values that may be linked to this attribute?

joeyjurjens:
@specialunderwear I added a migration that removes duplicate attributes before creating this unique constraint migration.

viggo-devries:
@joeyjurjens some checks were not successful could you look at those?

joeyjurjens:
@viggo-devries It was because the branch wasn't up to date with master anymore. I've rebased it and now the checks pass again.

viggo-devries:
> @viggo-devries It was because the branch wasn't up to date with master anymore. I've rebased it and now the checks pass again.

Nice, one thing I noticed was that all catalogue id's changed in the migration. Why is that?

@specialunderwear please look at the latests changes and review please so we can add this to the upcoming Oscar release.

joeyjurjens:
> Nice, one thing I noticed was that all catalogue id's changed in the migration. Why is that?

I think it's because of this: https://docs.djangoproject.com/en/4.0/releases/3.2/#customizing-type-of-auto-created-primary-keys

src/oscar/apps/catalogue/migrations/0025_attribute_code_uniquetogether_constraint.py
```diff
@@ -0,0 +1,72 @@
...
+    operations = [
+        migrations.AlterField(
+            model_name='attributeoption',
+            name='id',
+            field=models.BigAutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID'),
+        ),
+        migrations.AlterField(
+            model_name='attributeoptiongroup',
+            name='id',
+            field=models.BigAutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID'),
```
specialunderwear, line 21: @joeyjurjens please remove these alterations that you did not explicitly define. We don't want them right now, you can control the autofield with a setting in django

specialunderwear (changes requested):

src/oscar/apps/catalogue/migrations/0024_remove_duplicate_attributes.py
```diff
@@ -0,0 +1,65 @@
...
+            code=attribute_code,
+            product_class=product_class
+        )
+        used_attributes = attributes.filter(productattributevalue__isnull=False)
+        used_attribute_count = used_attributes.distinct().count()
+
+        # In most cases, the used attributes count will be one or zero as
+        # the dashboard will always show one attribute. If the used attribute
+        # count is one, we exclude this from attributes and remove the others.
+        # If it's zero, we pick the last created and delete others.
```
specialunderwear, line 35: Hi, I don't think you can just pick the last one.
What I think you should do, is use the code that normally picks the attribute.
That means you should use the ProductAttributesContainer to determine which attribute oscar would have chosen and use that as an exclude. The reason is that by default not all database backends will pick the last attribute.

specialunderwear (changes requested):

src/oscar/apps/catalogue/migrations/0024_remove_duplicate_attributes.py
joeyjurjens: I only pick the last one if none of the duplicate attributes have an value relation.
In that case, I don't think it matters which one I keep and which one I delete, as long as there's only one left, right?

If only ONE attribute has a value relation, I keep that one and remove others.

If MORE than ONE attribute has a relation, I try to resolve that as well from line 46 till 53.

src/oscar/apps/catalogue/migrations/0024_remove_duplicate_attributes.py
```diff
@@ -0,0 +1,65 @@
...
+            attributes.exclude(pk=used_attributes.first().pk).delete()
+            continue
+        elif used_attribute_count == 0:
+            attributes.exclude(pk=attributes.last().pk).delete()
+            continue
+
+        # If we found multiple attributes that have values linked to them,
+        # we must move them to one attribute and then delete the others.
+        # We can only do this if the value_types are all the same!
+        assert used_attributes.values("type").distinct().count() == 1, "Types must be same in order to move duplicate attributes."
```
specialunderwear, line 46: Can you add a helpful suggestion on how people can fix their data to make sure the migration can pass?

src/oscar/apps/catalogue/migrations/0024_remove_duplicate_attributes.py
joeyjurjens: The way attributes are being added to the product update/create view is as follows:
https://github.com/django-oscar/django-oscar/blob/34c4a1b3448a80fee321e1533d3d93fdd8edd3a8/src/oscar/apps/dashboard/catalogue/forms.py#L274-L285

So every duplicate code, would overwrite the field with that code before (line 282).
So the last on ends up being used inside the dashboard because it's never replaced.

I tried to use this method, but it used .get which raises an error if multiple objects are found.
https://github.com/django-oscar/django-oscar/blob/34c4a1b3448a80fee321e1533d3d93fdd8edd3a8/src/oscar/apps/catalogue/product_attributes.py#L59-L60

So I think I can keep .last() now, as this one is the one being used in dashboard?

src/oscar/apps/catalogue/migrations/0024_remove_duplicate_attributes.py
joeyjurjens: I have added this.

src/oscar/apps/catalogue/migrations/0024_remove_duplicate_attributes.py
specialunderwear: okok

specialunderwear (approved):
This has been reviewed and approved

viggo-devries (approved):

PR #3831, opened 2021-12-26
A much more efficient approach to ranges.

specialunderwear:
I ~~might~~ did translate this approach to the range itself for selecting products.
The big improvement in this approach is that selecting the categories  and all their parent categories is now much faster.
This used to require a full category table scan and would become very inefficient with many categories.
It is now really really fast, effectively removing any performance penalty in the use of ranges.

### DIFF
--- a/src/oscar/apps/catalogue/expressions.py
+++ b/src/oscar/apps/catalogue/expressions.py
@@ -1,0 +1,34 @@
+from django.db.models.expressions import Subquery
+
+EXPAND_UPWARDS_CATEGORY_QUERY = """
+(SELECT "CATALOGUE_CATEGORY_JOIN"."id" FROM "catalogue_category" AS "CATALOGUE_CATEGORY_BASE"
+LEFT JOIN "catalogue_category" AS "CATALOGUE_CATEGORY_JOIN" ON (
+    "CATALOGUE_CATEGORY_BASE"."path" LIKE "CATALOGUE_CATEGORY_JOIN"."path" || '%%%%'
+    AND "CATALOGUE_CATEGORY_BASE"."depth" >= "CATALOGUE_CATEGORY_JOIN"."depth"
+)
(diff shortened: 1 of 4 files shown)


src/oscar/apps/offer/queryset.py
```diff
@@ -1,57 +1,64 @@
...
+            # child products are excluded from a range if either they are
+            # excluded, or their parent.
+            return ~(
+                models.Q(excluded_products=product)
+                | models.Q(excluded_products__id=product.parent_id)
+            )
+        return ~models.Q(excluded_products=product)
 
+    def _included_products_clause(self, product):
+        if product.structure == product.CHILD:
```
solarissmoke, line 25: We can use `if product.is_child` here which is more succint. Same for all these other methods.

src/oscar/apps/catalogue/expressions.py
```diff
@@ -0,0 +1,34 @@
+from django.db.models.expressions import Subquery
+
+EXPAND_UPWARDS_CATEGORY_QUERY = """
```
solarissmoke, line 3: Dare I even ask whether this works with MySQL? :stuck_out_tongue: I know we don't officially support it - just be prepared for some complaints anyway if it doesnt.

solarissmoke (changes requested):

src/oscar/apps/offer/queryset.py
specialunderwear: There is  1 problem with that, in historic models all non standard stuff like properties is not available. So if I change it to is_child I cannot use this queryset in datamigrations. That would be problematic in many cases.

src/oscar/apps/catalogue/expressions.py
specialunderwear: There is no reason this shouldn't work in MySQL, however I haven't tested it because I don;t use it and we don't support it.

solarissmoke (approved):

PR #3843, opened 2022-01-10
changed keyOrder to form.field_order

rafael-h-ferreira:
changed keyOrder to field_order because it doesn't exist anymore in django.

### DIFF
--- a/src/oscar/apps/customer/forms.py
+++ b/src/oscar/apps/customer/forms.py
@@ -311,7 +311,9 @@ if Profile:  # noqa (too complex (12))
                 self.fields[field_name].initial = getattr(user, field_name)
 
             # Ensure order of fields is email, user fields then profile fields
-            self.fields.keyOrder = user_field_names + profile_field_names
+            self.field_order = user_field_names + profile_field_names
+            self.order_fields(self.field_order)
+
 
         class Meta:
             model = Profile


src/oscar/apps/customer/forms.py
```diff
@@ -311,7 +311,9 @@ def __init__(self, user, *args, **kwargs):
                 self.fields[field_name].initial = getattr(user, field_name)
 
             # Ensure order of fields is email, user fields then profile fields
-            self.fields.keyOrder = user_field_names + profile_field_names
+            self.field_order = user_field_names + profile_field_names
```
solarissmoke, line 314: Could you clarify why we need to set `self.field_order` here? I feel like it's not necessary.

rafael-h-ferreira: Are you suggesting removing the line completely? Well, Django does have the field_order field in the BaseForm class and uses it for the obvious reasons (https://github.com/django/django/blob/456466d932830b096d39806e291fe23ec5ed38d5/django/forms/forms.py#L95). In my case, I was using a Profile class to add a language field along with the AUTH_PROFILE_MODULE setting and it crashed on that line. I think it's nice to keep the Django approach but the truth is , in my case, I don't need it. If anyone needs it, they can subclass UserAndProfileForm and implement it, true.

specialunderwear: I think you should set it to None, just like https://github.com/django/django/blob/456466d932830b096d39806e291fe23ec5ed38d5/django/forms/forms.py#L62 if it causes problems?

PR #3846, opened 2022-01-11
Upgrade to copy-props 2.0.5 in order to fix vulnerability CVE-2020-28503.

sasha0:


package.json
```diff
@@ -7,6 +7,7 @@
   "devDependencies": {
     "@fortawesome/fontawesome-free": "^5.15.4",
     "bootstrap": "^4.6.1",
+    "copy-props": ">=2.0.5",
```
solarissmoke, line 10: I don't think we should be adding this as an explicit dependency - should just be a case of updating the lock file, or updating the dependency that uses this.

PR #3854, opened 2022-01-17
Prevent non-browsable product index update

nakarinh14:


src/oscar/apps/search/search_indexes.py
```diff
@@ -106,3 +106,6 @@ def get_updated_field(self):
         Can be used to filter the query set when updating the index
         """
         return 'date_updated'
+
+    def should_update(self, instance, **kwargs):
+        return instance.is_public and not instance.is_child
```
specialunderwear, line 111: What is the reason you need to exclude these products from the index? I know you don't want these on the frontend, but you can just filter them out? For backend search application you still need them to be indexed.

viggo-devries:
@nakarinh14 thanks for this! I would really like to see a test which shows this didn't work before and proof that this code actually works!

specialunderwear (changes requested):
Please add a test demonstrating the issue

samar-hassan:
@nakarinh14 please add a test to indicate the problem

viggo-devries:
@nakarinh14 the index_queryset: return self.get_model().objects.browsable().order_by('-date_updated') should already take care of this right?

PR #3859, opened 2022-01-27
Extended Product Option:

SummerSeaSun:
- default: value that is shown as form initial data
- sequence: allow user sort of options
- help_text
- choices: new form field type show a select as result of comma separated input

Affect setup: host/dashboard/catalogue/option/ and product detail view.


solarissmoke:
I think the functionality you're aiming for is covered (with greater flexibility) in #3795, so I'm going to close this in favour of that PR. Thanks nonetheless for your efforts.

PR #3860, opened 2022-01-28
Docker file to create Sandbox

guilherme-funchal:
Docker file and docker-compose.yml to create demo project Sandbox without need to compiling source code

To build is necessary : 
#docker-compose up --build

or 

#docker-compose up

### DIFF
--- a/Dockerfile.sandbox
+++ b/Dockerfile.sandbox
@@ -1,0 +1,21 @@
+FROM python:3.8
+ENV PYTHONUNBUFFERED 1
+
+COPY ./requirements.txt /requirements.txt
+
+RUN pip3 install -r /requirements.txt
+
+RUN groupadd -r django && useradd -r -g django django
+
+COPY . /app
+
+RUN chown -R django /app
+
+WORKDIR /app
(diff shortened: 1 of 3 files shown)


guilherme-funchal:
Hello!

I'm from brazil and I'm create some modifications to my project, and I like to contribuite in project.

Docker file and docker-compose.yml to create demo project Sandbox without need to compiling source code

solarissmoke:
@guilherme-funchal it's not clear to me what this change achieves? There is already a working Dockerfile in the project, which can be used to create an instance of the sandbox. I don't think a docker compose configuration is necessary since there is only one service in the cluster. We also can't modify the requirements.txt file in the way you have proposed, because it creates a circular reference to this package itself.

I'm going to close this for now - but please feel free to clarify what the specific issue is that is preventing you from using the existing Dockerfile, and then we can see what to do about it.

guilherme-funchal:
No problem, Maybe because of my inexperience I couldn't run docker and I modified it to make it easier.

So is trying to collaborate, greetings.

PR #3867, opened 2022-02-15
Some tests misusing assertTrue for comparisons fix

### LINKED ISSUE #3866: Some tests misusing assertTrue for comparisons
`assertTrue` is not for comparing arguments, should use `assertEqual` for that.

The developer's intent of the test was to compare argument 1 with argument 2, which is not happening. Really what is happening is the test is passing because first argument is truthy. The correct method to use is assertEqual. [more details](https://codereview.doctor/features/python/best-practice/avoid-misusing-unittest-assert-true)

https://github.com/django-oscar/django-oscar/blob/c05353a76dfbb7ce0df744e97cf0e282c54cae56/tests/integration/catalogue/test_options.py#L35

I found this issue automatically, see other issues [here](https://codereview.doctor/django-oscar/django-oscar)

code-review-doctor:
Fixes #3866

### DIFF
--- a/tests/integration/catalogue/test_options.py
+++ b/tests/integration/catalogue/test_options.py
@@ -32,7 +32,7 @@ class ProductOptionTests(TestCase):
         qs = Product.objects.browsable().base_queryset().filter(id=self.product.id)
         product = qs.first()
         self.assertTrue(product.has_options)
-        self.assertTrue(product.has_product_options, 1)
+        self.assertEqual(product.has_product_options, 1)
 
     def test_queryset_both(self):
         "The options attribute on a product should return a queryset containing "


tests/integration/catalogue/test_options.py
```diff
@@ -32,7 +32,7 @@ def test_queryset_per_product(self):
         qs = Product.objects.browsable().base_queryset().filter(id=self.product.id)
         product = qs.first()
         self.assertTrue(product.has_options)
-        self.assertTrue(product.has_product_options, 1)
+        self.assertEqual(product.has_product_options, 1)
```
specialunderwear, line 35: Unfortunately has_product_options will return a boolean and not an integer. So this does not fix the test, in fact it breaks the test! While you have a point that this test is faulty, this is not a good fix, it is not even a fix at all!

specialunderwear (changes requested):

tests/integration/catalogue/test_options.py
code-review-doctor: apologies for wasting you time then. I scanned a few hundred codebases to find cases of what _looks_ like misuse of `assertTrue`. False positives will happen. Sorry it was your repo that had the false positive :)

PR #3868, opened 2022-02-22
remove null=True for CharFields and TextFields

byshen:
As discussed in #584 and #821, the fields (CharFields and TextFields) should not be allowed to be null, as it reflects two possible states for data.
 

- https://docs.djangoproject.com/en/dev/ref/models/fields/#null

The modified fields are the `meta_title` and `meta_description` of the Product and Category class, in `apps.catalogue.abstract_models`.


jwayodi:
It looks like `catalogue.Option.help_text` also has this issue? Could you please search through the whole codebase, for any others?

Also, please create data migrations to convert the stored `NULL`s for these fields to empty strings (like #821 does).

byshen:
Thanks for your reply.  `catalogue.Option.help_text` is recently added. I changed it now. 

I have added the data migrations as advised.  Please see if it fits.

byshen:
Also #3870 and #3869 are similar issues. Could you please have a look? I can help create a PR if confirmed. Thanks!

solarissmoke:
Setting `null=True` on newly added fields is actually a deliberate design decision - the reason is that it allows for a much faster database migration on large catalogues. If you introduce a new field, whose default is a blank string, most database engines still have to write this blank string into every row of the database, which can be pretty slow if there are a few million records in the database. But if you default to `null` then on the databases we care about (Postgres primarily) this doesn't happen - this is the reason why `null=True` is set on newly added fields. The cost - of having two possible states - is considered an acceptable tradeoff.

Pinging @specialunderwear to weigh in on this - but I don't think it is something we are likely to want to change.

byshen:
Thank you for your reply! `null=False` can help prevent the ambiguity, but it seems to be a valid concern for large tables.

However, these fields are the table for `Category`, `ProductAttributeValue`, `Option`, which are usually small tables comparing to `user` or `product` table, but contains important data. Do you think these tables are small enough to add this constraint? 

Also, do you recommend to add the `not null` constraints for those who want to self-host a new site? Thanks!

solarissmoke:
`ProductAttributeValue` typically has the same number of rows as `Product`. `Category` can also have a very large number of categories depending on the nature of the site. I'm not sure the constraint is necessary (either in Oscar core, or in a project) unless there is a demonstrated case of it causing problems in a typical use case?

specialunderwear:
I agree that there is a valid reason the fields can be null. On sites with heavy traffic, the downtime the migrations can cause because of the need to copy the entire table can be unacceptable. However it is a valid concern that @byshen has, so we would accept a pull request adding a comment explaining the deliberate nature of the null=True

specialunderwear:
Please also note that this pull request by itself will cause the above mentioned downtime for all very large sites that use oscar. Are you sure you want to be the one that caused this @byshen ?

byshen:
Thank you @solarissmoke @specialunderwear for the discussions.

I will close this and submit a PR to improve the comment.

PR #3871, opened 2022-02-23
Use as_stars template filter instead of very long if/else statement

SaturnFromTitan:
A small refactoring

### DIFF
--- a/src/oscar/templates/oscar/catalogue/reviews/review_form.html
+++ b/src/oscar/templates/oscar/catalogue/reviews/review_form.html
@@ -16,7 +16,7 @@
                     </label>
                     <div class="controls">
                         {{ form.score }}
-                        <div class="star-rating {{ form.score.value|as_star }}">
+                        <div class="star-rating {{ form.score.value|as_stars }}">
                             <i class="fas fa-star"></i>
                             <i class="fas fa-star"></i>
                             <i class="fas fa-star"></i>
--- a/src/oscar/templates/oscar/catalogue/reviews/review_form.html
+++ b/src/oscar/templates/oscar/catalogue/reviews/review_form.html
@@ -1,5 +1,6 @@
 {% extends "oscar/catalogue/detail.html" %}
 {% load i18n %}
+{% load reviews_tags %}
 
 {% block product_review %}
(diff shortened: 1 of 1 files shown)


src/oscar/templates/oscar/catalogue/reviews/review_form.html
```diff
@@ -16,7 +16,7 @@
                     </label>
                     <div class="controls">
                         {{ form.score }}
-                        <div class="star-rating {% if form.score.value == '1' %}One{% elif form.score.value == '2' %}Two{% elif form.score.value == '3' %}Three{% elif form.score.value == '4' %}Four{% elif form.score.value == '5' %}Five{% endif %}">
+                        <div class="star-rating {{ form.score.value|as_star }}">
```
jwayodi, line 19: I think this [should be](https://github.com/django-oscar/django-oscar/blob/master/src/oscar/templatetags/reviews_tags.py#L7) `as_stars`?

SaturnFromTitan: Indeed 🙈 Thanks for spotting this!

src/oscar/templates/oscar/catalogue/reviews/review_form.html
```diff
@@ -16,7 +16,7 @@
                     </label>
                     <div class="controls">
                         {{ form.score }}
-                        <div class="star-rating {% if form.score.value == '1' %}One{% elif form.score.value == '2' %}Two{% elif form.score.value == '3' %}Three{% elif form.score.value == '4' %}Four{% elif form.score.value == '5' %}Five{% endif %}">
+                        <div class="star-rating {{ form.score.value|as_stars }}">
```
solarissmoke, line 19: You need to load `reviews_tags` at the top of this file, otherwise the filter cannot be found

solarissmoke (changes requested):

src/oscar/templates/oscar/catalogue/reviews/review_form.html
SaturnFromTitan: indeed - thanks for spotting this as well

PR #3873, opened 2022-02-23
[#3872] OrderFactory doesn't require settings.SITE_ID to be set anymore

### LINKED ISSUE #3872: OrderFactory breaks if settings.SITE_ID isn't set

### Issue Summary

The OrderFactory requires the `SITE_ID` setting to be set. So when I import the factory and don't have this setting, the application crashes.

https://github.com/django-oscar/django-oscar/blob/master/src/oscar/test/factories/order.py#L62

### Proposed Solution

At the bare minimum, this should be a lazy attribute, so I can import the OrderFactory and overwrite the site_id default value. I think it'd be even better to try and use settings.SITE_ID and if it's not available to fall back to `Site.objects.first()`. Wdyt?

I'll prepare a PR ✌️

SaturnFromTitan:
closes #3872

### DIFF
--- a/src/oscar/test/factories/order.py
+++ b/src/oscar/test/factories/order.py
@@ -2,7 +2,6 @@ from decimal import Decimal as D
 
 import factory
 from django.conf import settings
-from django.contrib.sites.models import Site
 
 from oscar.core.loading import get_class, get_model
 from oscar.core.utils import slugify
@@ -60,9 +59,7 @@ class OrderFactory(factory.django.DjangoModelFactory):
     if hasattr(settings, 'OSCAR_INITIAL_ORDER_STATUS'):
         status = settings.OSCAR_INITIAL_ORDER_STATUS
 
-    site_id = factory.LazyAttribute(
-        lambda o: settings.SITE_ID if hasattr(settings, "SITE_ID") else Site.objects.first()
-    )
+    site_id = getattr(settings, "SITE_ID", None)
     number = factory.LazyAttribute(lambda o: '%d' % (100000 + o.basket.pk))
     basket = factory.SubFactory(
         'oscar.test.factories.BasketFactory')
--- a/src/oscar/test/factories/order.py
+++ b/src/oscar/test/factories/order.py
@@ -2,6 +2,7 @@ from decimal import Decimal as D
 
 import factory
 from django.conf import settings
+from django.contrib.sites.models import Site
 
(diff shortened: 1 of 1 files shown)


specialunderwear (changes requested):
Hi unfortunately this is not the correct way to implement multisite in django.
If you want to do that you have to set 
```
SITE_ID = None
```
In your settings file and not leave it out completely :D
That will make django deduce the site from the request instead of reading it from a file.
So if you where settings SITE_ID from code in your settings, that is not the correct way to do it, you should never modify settings after the application has loaded ;)

SaturnFromTitan:
Hi @specialunderwear, thanks for your feedback.

I don't think I use the sites framework in an incorrect way. I have different top-level domains for different languages/countries (e.g. `example.de` for Germany, `example.it` for Italy, etc.). All pages are served from the same server, so I want the site to be determined based on the request, not based on a setting. Also, I'm not modifying any settings at runtime, don't worry.

For now, I use `SITE_ID = None` as a workaround. Looking at the implementation of `CurrentSiteMiddleware` and `Site.objects.get_current`, it's semantically equivalent to set `SITE_ID = None` or to omit the setting at all. I'd prefer the latter and think it'd be great if oscar allows it (especially since it only comes down to a test factory).

src/oscar/test/factories/order.py
```diff
@@ -59,7 +60,9 @@ class Meta:
     if hasattr(settings, 'OSCAR_INITIAL_ORDER_STATUS'):
         status = settings.OSCAR_INITIAL_ORDER_STATUS
 
-    site_id = settings.SITE_ID
+    site_id = factory.LazyAttribute(
```
specialunderwear, line 63: Would everything work properly if you just do:
```
site_id = getattr(settings, "SITE_ID", None)
```
That would be preferable to me because it is much simpler. Would the tests still pass that way?

specialunderwear (changes requested):

src/oscar/test/factories/order.py
SaturnFromTitan: Yes indeed, that would work as well :)

SaturnFromTitan:
@specialunderwear I believe I addressed all your feedback. Is there anything else or is this good to be merged?

joeyjurjens (approved):

specialunderwear (approved):

PR #3876, opened 2022-02-28
Disable html5 validation of attribute form

webomar:


solarissmoke:
@webomar can you explain why you feel we should be disabling validation on this form?

joeyjurjens:
@webomar #3892 fixes the issue you were trying to fix with this PR as I took over the issue from you. So closing this PR.

PR #3884, opened 2022-03-02
Prevent unnecessary database query in UseFirstStockRecord strategy when passed an annotated product.

### LINKED ISSUE #3875: Performance issue in cases where selecting strategy for product

### Issue Summary

In Oscar 2.1, Selecting a strategy for product in [`UseFirstStockRecord`](https://github.com/django-oscar/django-oscar/blob/2.1.0b0/src/oscar/apps/partner/strategy.py#L188) used to (correctly, for performance reasons) index on the related stock records for a product to get the first. This meant, of course, that the `prefetch_related` call to `stockrecords` on `Product` was used and no additional queries generated.

Now, in Oscar 3.1 [this was changed to use the queryset's `first` method](https://github.com/django-oscar/django-oscar/blob/3.1/src/oscar/apps/partner/strategy.py#L188), which unfortunately makes the prefetch invalid and causes a new query each time a product shows price info (for various reasons, this has a huge performance impact on my use case).

This change was introduced in [this commit](https://github.com/django-oscar/django-oscar/commit/3a826796d01e5cba345aa646dbc5c8ae466cbf4a), which also implemented the same pattern over various related queries, and therefore I assume has similar performance implications (but I haven't checked).

### Steps to Reproduce

Try to render a page with various snippets of pricing info for a collection of hundreds of products - use a tool like Django Debug Toolbar to see the query multiplication :)

### Technical details

* Python version: Run `python --version`: 3.10
* Django version: Look in your requirements.txt, or run `pip show django | grep Version` 3.2.12
* Oscar version: Look in your requirements.txt, or run `pip show django-oscar | grep Version`. 3.1

solarissmoke:
Fixes #3875 . 

I checked [all the other changes](https://github.com/django-oscar/django-oscar/commit/3a826796d01e5cba345aa646dbc5c8ae466cbf4a) in the commit that introduced this regression and am not sure that any of the other instances need to be reverted - in all cases I can't see anywhere that we use `prefetch_related` with the associated models, or any likely use case for a project needing to do so - but open to feedback on this.

### DIFF
--- a/src/oscar/apps/partner/strategy.py
+++ b/src/oscar/apps/partner/strategy.py
@@ -185,17 +185,19 @@ class Structured(Base):
 # Mixins - these can be used to construct the appropriate strategy class
 
 
-class UseFirstStockRecord(object):
+class UseFirstStockRecord:
     """
     Stockrecord selection mixin for use with the ``Structured`` base strategy.
     This mixin picks the first (normally only) stockrecord to fulfil a product.
-
-    This is backwards compatible with Oscar<0.6 where only one stockrecord per
-    product was permitted.
     """
 
     def select_stockrecord(self, product):
-        return product.stockrecords.first()
+        # We deliberately fetch by index here, to ensure that no additional database queries are made
+        # when stockrecords have already been prefetched in a queryset annotated using ProductQuerySet.base_queryset
+        try:
+            return product.stockrecords.all()[0]
+        except IndexError:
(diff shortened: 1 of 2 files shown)


pwalsh (approved):
Looks good, and especially good that there is a a test to check for this now :)

specialunderwear (approved):

PR #3886, opened 2022-03-02
Fix that the condition form required a range when choosing a custom condition

### LINKED ISSUE #3885: When choosing a custom condition for your offer, it requires a range while it shouldn't
When you are choosing a custom condition on your offer, it will throw an error if you do not select a range as well.

image

However, the range is saved on the custom condition model so it should not be required.

### Steps to Reproduce
1. Create a custom condition, see [here](https://django-oscar.readthedocs.io/en/3.0.2/howto/how_to_create_a_custom_condition.html) how.
2. Create an offer and choose your custom condition without choosing a range, type and value.
3. Error should occur.

The logic in the the clean method of the ConditionForm seems wrong (line 127):
https://github.com/django-oscar/django-oscar/blob/23013a61a1f84517522c071bbeddb3fd3e841ee6/src/oscar/apps/dashboard/offers/forms.py#L126-L133

I think it should be like the BenefitForm does it:
https://github.com/django-oscar/django-oscar/blob/23013a61a1f84517522c071bbeddb3fd3e841ee6/src/oscar/apps/dashboard/offers/forms.py#L178-L191

Or is there any reason this is a requirement even when you choose a custom condition?

### Technical details

* Python version: Python 3.9.9
* Django version: Version: 3.2.9
* Oscar version: Version: 3.1

joeyjurjens:
Fixes #3885

### DIFF
--- a/src/oscar/apps/dashboard/offers/forms.py
+++ b/src/oscar/apps/dashboard/offers/forms.py
@@ -123,12 +123,17 @@ class ConditionForm(forms.ModelForm):
 
         # Check that either a condition has been entered or a custom condition
         # has been chosen
+        if not any(data.values()):
+            raise forms.ValidationError(
+                _("Please either choose a range, type and value OR "
+                  "select a custom condition"))
+
         if data['custom_condition']:
             if data.get('range') or data.get('type') or data.get('value'):
                 raise forms.ValidationError(
                     _("No other options can be set if you are using a "
                       "custom condition"))
-        elif not all([data.get('range'), data.get('type'), data.get('value')]):
+        elif not data.get('type'):
             raise forms.ValidationError(
                 _("Please either choose a range, type and value OR "
                   "select a custom condition"))
--- a/src/oscar/apps/dashboard/offers/forms.py
+++ b/src/oscar/apps/dashboard/offers/forms.py
@@ -123,10 +123,11 @@ class ConditionForm(forms.ModelForm):
 
         # Check that either a condition has been entered or a custom condition
         # has been chosen
-        if data.get('custom_condition'):
-            if not data.get('range', None):
+        if data['custom_condition']:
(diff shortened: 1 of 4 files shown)


jwayodi:
Just to confirm that this change does not modify any behaviour - if a custom condition is selected, then the other fields are [completely ignored](https://github.com/django-oscar/django-oscar/blob/0a01977e9cef028bd3e8985309aa4741b6a8cb33/src/oscar/apps/dashboard/offers/forms.py#L141).

jwayodi:
I like the approach taken by `BenefitForm.clean`, where it only checks to make sure the `type` has been selected, and then lets the `type`-specific models do more specific validation, according to their requirements.

This allows creating conditions that may make use of a merchant-configurable `range` and `value` (or one without the other). E.g. a "No item in the basket should weigh more than x kilogrammes" condition (which does not require a `range`, but requires a `value`). The "classic" custom conditions are not dynamic - they use the range and value they are initially created with.

This is probably overdesigning (since I am not sure anyone has asked for this extra flexibility). But if you agree that doing it this way is a good idea, then you can copy the [implementation](https://github.com/django-oscar/django-oscar/blob/0a01977e9cef028bd3e8985309aa4741b6a8cb33/src/oscar/apps/offer/abstract_models.py#L543) of `offer.Benefit.clean`, and for the current types, have `type`-specific clean methods that ensure that a range and value has been selected. I do not know if this can be DRYed.

joeyjurjens:
@jwayodi Could you take a look at it whenever you got the time? We would like to have this merged before the new release.

jwayodi:
@joeyjurjens Yeah, sure. Let me look at it in the next couple of days.

jwayodi (approved):
This looks good to me. Thanks!

solarissmoke (approved):

PR #3887, opened 2022-03-02
Added catalogue merge migration

joeyjurjens:
There were conflicting migrations due to these two merged PR's:
#3795 & #3823


jwayodi:
The changes in PR #3795 (which caused the conflict) are not yet in a release, and it was only just merged yesterday, so I think we can just edit the migration to fix the conflict?
- rename `src/oscar/apps/catalogue/migrations/0024_predefined_product_options.py` to `src/oscar/apps/catalogue/migrations/0026_predefined_product_options.py`, and change its dependancy from `0023_auto_20210824_1414` to `0025_attribute_code_uniquetogether_constraint`
- rename `tests/_site/apps/catalogue/migrations/0024_predefined_product_options.py` to `tests/_site/apps/catalogue/migrations/0026_predefined_product_options.py`, and change its dependancy from `0023_auto_20210824_1414` to `0025_attribute_code_uniquetogether_constraint`

Also, please copy the migrations from PR #3823 (`src/oscar/apps/catalogue/migrations/0024_remove_duplicate_attributes.py` and `src/oscar/apps/catalogue/migrations/0025_attribute_code_uniquetogether_constraint.py`) into `tests/_site/apps/catalogue/migrations/`, to keep them in sync.

@solarissmoke?

joeyjurjens:
@jwayodi Ah yes, that might be better indeed. I'll wait for @solarissmoke his reply and if he agrees I'll adjust my PR.

solarissmoke:
I agree with @jwayodi suggestion :+1:

jwayodi (commented):
I do not think the changes in `tests/integration/catalogue/test_attributes.py` are related to this PR?

joeyjurjens:
> I do not think the changes in `tests/integration/catalogue/test_attributes.py` are related to this PR?

Not directly.

But because I forgot to add the migrations from #3823 to the tests/_site/apps/catalogue/migrations/ folder, the test I wrote in that PR didn't have the unique constraint migration applied to the database so it did not raise the IntegrityError.

However, now that I have added it to the test migrations, it does raise that error so I thought I'd fix that in this PR as well, as this PR also copies those migrations to the test migrations folder.

jwayodi:
> > I do not think the changes in `tests/integration/catalogue/test_attributes.py` are related to this PR?
> 
> Not directly.
> 
> But because I forgot to add the migrations from #3823 to the tests/_site/apps/catalogue/migrations/ folder, the test I wrote in that PR didn't have the unique constraint migration applied to the database so it did not raise the IntegrityError.
> 
> However, now that I have added it to the test migrations, it does raise that error so I thought I'd fix that in this PR as well, as this PR also copies those migrations to the test migrations folder.

Okay, I see.

In that test, instead of using `ProductAttribute.objects.create()` and then testing for an `IntegrityError`, and having to roll back the transaction, you could instead have used `ProductAttribute().full_clean()` and tested for a `ValidationError`?

I would also use `ProductAttribute().full_clean()` for `attribute3`, as `full_clean` [runs](https://docs.djangoproject.com/en/3.2/ref/models/instances/#django.db.models.Model.full_clean) the uniqueness validation. And just `ProductAttribute.objects.create()` (without the `.full_clean()`) for `attribute1`, as we are not running the validation here.

Also, the test's name has an extra `s`.

jwayodi (approved):
This looks good to me.

Issue #3890, opened 2022-03-10
ProductImage is re-saved after deletion (solved)


snake-soft:
### Issue Summary

After deleting a ProductImage with the dashboard Form, the Image is existing without a File.
That causes the Product list to throw an error as well as the detail page.

It's the same error as in this issue #3735 and maybe this issue #3762

### Steps to Reproduce

1. Open the dashboard product form
2. Click checkbox to delete image
3. Save

It doesn't happen when using the sandbox and therefore i guess the error is somehow domain specific although i couldn't find any customized line of code in any of the traces.

### Problem source

When deleting an Image, the Image is deleted **and recreated** in AbstractProductImage.delete().
This happens because the product of this object is not re-evaluated from the db before saving again.
So it saves exactly the same image with all its Metadata (and a new created-date).

### Solution

Re-evaluate the product from db before reordering the images in AbstractProductImage.delete().
No matter if this error occurs only under some circumstances, i'd suggest to include this single line to explicitly re-evaluate.
It should have no side-effects but could save some time searching this error.

### Technical details

* Python version: 3.9.2
* Django version: 3.2.12
* Oscar version: 3.1

solarissmoke:
It's hard to be sure how to fix this unless you're able to provide specific steps to reproduce it on the sandbox (or on a modified sandbox). It does seem like it's an issue specific to your project. #3735 has been fixed, which mitigates the problem in the event that an image file is deleted without the associated model being deleted. I'm going to close this for now - feel free to reopen if you do find out how to reproduce it in the sandbox.

snake-soft:
> ng deleted

Now we found the circumstance causing this error:
In the customized Product Manager (inherited from the original ProductQuerySet) there is a prefetch_related(images).
Because of that self.product.images.all() is evaluated **before** one of its elements is deleted (self).
In the next three lines after deletion, it will save all pre-evaluated images and thus save the metadata of the already deleted image without a file. That causes the error like in the linked issues.

It is definitely a design error to not re-evaluate the images before saving them again because it is not ensured to be refreshed after deletion. Lazy evaluation is part of Djangos core design and should explicitly be considered when using it in such a error prone way. On the other hand it doesn't have any side-effects when using a refresh_from_db() at this place because the method doesn't have any returns.

PR #3891, opened 2022-03-10
Fixed ProductImage is saved after deletion

snake-soft:
Re-evaluate the product from db before reordering the images in
AbstractProductImage.delete().

This avoids accidentaly recreation of the image.
#3890

### DIFF
--- a/src/oscar/apps/catalogue/abstract_models.py
+++ b/src/oscar/apps/catalogue/abstract_models.py
@@ -1414,8 +1414,10 @@ class AbstractProductImage(models.Model):
         """
         Always keep the display_order as consecutive integers. This avoids
         issue #855.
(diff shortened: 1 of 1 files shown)


solarissmoke:
It's not clear to me how the issue you raise in #3890 arises - as you indicate in the issue itself it seems to be something specific to your setup? I would prefer to know the underlying cause of the issue before we apply a "best guess" fix for a problem that we haven't understood.

solarissmoke:
Closing for now, as per the reasoning at https://github.com/django-oscar/django-oscar/issues/3890#issuecomment-1068733537

snake-soft:
https://github.com/django-oscar/django-oscar/issues/3890#issuecomment-1072237945

PR #3892, opened 2022-03-11
Fix the issue where html5 validation errors were not obvious when editing a product

### LINKED ISSUE #3791: HTML5 form validation errors not obvious to user in tabbed dashboard edit views
### Issue Summary

HTML5 form validation errors are not obvious to a user if the fields that failed to validated are not in a visible tab.

### Steps to Reproduce

1. Create a product class that has a required attribute.
2. Edit a product of this class, but leave the attribute value empty, and tab back to the "Product details" part of the edit form.
3. Attempt to save the product. Nothing will happen - but it's not at all obvious to the user why.

We should do better here and tab to the effected field and/or add a warning icon next to the affected tab. I am not 100% sure but I feel like this may have broken after upgrading to Bootstrap 4.

joeyjurjens:
This fixes #3791.

The error icon next to the tab appears on page reload and this was also the case before the bootstrap 4 port.

### DIFF
--- a/src/oscar/static_src/oscar/js/oscar/dashboard.js
+++ b/src/oscar/static_src/oscar/js/oscar/dashboard.js
@@ -264,11 +264,16 @@ var oscar = (function(o, $) {
                 }
             });
 
+            // Add href to url, so when the page is reloaded this tab will be displayed.
+            $('.nav-tabs a').on('shown.bs.tab', function (e) {
+                window.location.hash = e.target.hash;
+            });
+
             // Display tabs that have invalid input fields
             $('input').on('invalid', function(){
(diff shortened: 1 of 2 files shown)


src/oscar/static_src/oscar/js/oscar/dashboard.js
```diff
@@ -264,11 +264,16 @@ var oscar = (function(o, $) {
                 }
             });
 
+            // Add href to url, so when the page is reloaded this tab will be displayed.
```
joeyjurjens, line 267: Just a quick comment here; In the past jquery would do this, but jquery deprecated this.

solarissmoke (approved):

PR #3895, opened 2022-03-21
Made order basket read only

### LINKED ISSUE #3888: Large number of baskets makes Order Admin page slow to load

### Issue Summary

A summary of the issue.

Our shop has a large number of baskets now.  This mean that the basket dropdown in the order admin page takes about a minute to load.

### Steps to Reproduce

Create 2000+ baskets, submit one order, and then view that order in Django Admin

Is there a way we should be purging baskets, can we add Select2 (or otherwise) or make this read only?

### Technical details
Python 3.9, Django 3.2, Oscar 3.1

st8st8:
Fixes #3888


specialunderwear (approved):
I think it makes sense that the basket is readonly, I can't think of any use for changing the basket of an order.

joeyjurjens (approved):

Issue #3509, opened 2020-09-24
Add ability to make wishlists public/shareable


solarissmoke:
The existing wishlist functionality in Oscar is incomplete - the models allow for wishlists to be "public" or "shared", but the UI does not allow for making a wishlist public/shared. 

1. *Public wishlists* - we should llow users to make wishlists "public" which means that they can share links to this with others. My feeling is that "public" in this case should still require the viewer to be logged-in, so that these views are not exposed to abuse and SEO implications. This should be relatvely straighforward to implement, and has a clear use case.

2. *Shared wishlists* - it seems the original intention behind this was to have lists that can collaboratively be managed by multiple users. The use case for this isn't clear to me, and I think needs elaboration before we can implement.

Andruten:
I've implemented the public wishlist in a project and one of the requirements is public wishlists can be shared by link and email with people without account (some kind of influencers' wishlists). So public wishlists is a really interesting feature.

joeyjurjens:
This has been implemented in #3905

PR #3905, opened 2022-04-01
Implement shareable wishlists

joeyjurjens:
Partly implements #3509 .

This PR makes it possible to share your wish list link with other people.
There are three types of wish lists:
1. Private
  - Only available for the owner of the wish list.
 2. Shared
   - Available for the owner and emails the owner has set for access. This requires people to create an account with this email in order to view it (if they didn't yet). The reason for a email field instead of a user field is that I don't think customers should be able to select users.
 3. Public
   - Available for everyone, also non logged in users. We could add no-index and no-follow tags so robots don't index the page, but I think a public wish list should not require a login but we can discuss about it.

The shared wish list does **not** allow editing the wish list (except owner of course).

As the issue also states: "it seems the original intention behind this was to have lists that can collaboratively be managed by multiple users. The use case for this isn't clear to me, and I think needs elaboration before we can implement."

So I haven't implemented any of that. We could add a new field "allow_edit" on the newly created model and then implement from there.


src/oscar/templates/oscar/customer/wishlists/wishlists_form.html
```diff
@@ -38,6 +40,16 @@
                 {% endblocktrans %}
             </p>
         {% endif %}
+
+        <div class="wishlist-shared-emails mt-5">
+            <h3>{% trans "Share wishlist" %}</h3>
+            <small>
+                {% trans "Add the emails from the people you want to share your wishlist with. The wishlist visibility must be set to shared if you want to share it with certain people only." %}
```
viggo-devries, line 47: I think this should be a different text

viggo-devries (changes requested):

src/oscar/apps/wishlists/views.py
```diff
@@ -0,0 +1,46 @@
...
+    def dispatch(self, request, *args, **kwargs):
+        wishlist = self.get_object()
+
+        if wishlist.is_allowed_to_see(request.user):
+            return super().dispatch(request, *args, **kwargs)
+        elif wishlist.visibility == WishList.PRIVATE:
+            raise Http404
+        else:
+            messages.info(request, _("Please login with the email address that is shared to this wish list."))
+            print(settings.LOGIN_REDIRECT_URL)
```
viggo-devries, line 28: print

src/oscar/apps/wishlists/views.py
```diff
@@ -0,0 +1,46 @@
...
+            return redirect(settings.LOGIN_REDIRECT_URL)
```
viggo-devries, line 29: is it maybe possible to add a ?next parameter to redirect the user to the wishlist after logging in?

viggo-devries, line 29: is it maybe possible to add a ?next parameter to redirect the user to the wishlist after logging in?

viggo-devries (changes requested):

viggo-devries (approved):

src/oscar/apps/wishlists/abstract_models.py
```diff
@@ -69,10 +69,13 @@ def random_key(cls, length=6):
...
     def is_allowed_to_see(self, user):
-        if self.visibility in (self.PUBLIC, self.SHARED):
+        if user == self.owner:
             return True
-        else:
-            return user == self.owner
+        elif self.visibility == self.PUBLIC:
+            return True
+        elif self.visibility == self.SHARED and user.is_authenticated:
+            return self.shared_emails.filter(models.Q(email=user.email)).exists()
```
specialunderwear, line 77: why models.Q?

src/oscar/apps/wishlists/views.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def dispatch(self, request, *args, **kwargs):
+        wishlist = self.get_object()
+
+        if wishlist.is_allowed_to_see(request.user):
+            return super().dispatch(request, *args, **kwargs)
+        elif wishlist.visibility == WishList.PRIVATE:
+            raise Http404
+        else:
+            messages.info(request, _("Please login with the email address that is shared to this wish list."))
```
specialunderwear, line 28: I don't think this message is a clear way of handling not allowed to see the message. I think you should only redirect to login if the user is anonymoes, and simply state "You are not allowed to see this wishlist", 403 if sombody is logged in but not allowed to view the wishlist.

specialunderwear (changes requested):
I think we need a native english speaker to review the help messages @solarissmoke

specialunderwear (dismissed):

src/oscar/templates/oscar/customer/wishlists/wishlists_form.html
```diff
@@ -38,6 +40,16 @@
                 {% endblocktrans %}
             </p>
         {% endif %}
+
+        <div class="wishlist-shared-emails mt-5">
+            <h3>{% trans "Share wishlist" %}</h3>
+            <small>
+                {% trans "If you want to share your wishlist with other accounts, you can add their emails below. The visibility must be set to shared in order for this to work." %}
```
solarissmoke, line 47: Please amend the first sentence to: "If you want to share your wishlist with others, you can add their email addresses below."

src/oscar/templates/oscar/customer/wishlists/wishlists_list.html
```diff
@@ -22,6 +22,11 @@
                                     <a href="{% url 'customer:wishlists-update' key=wishlist.key %}" class="dropdown-item">
                                         {% trans 'Wish list settings' %}
                                     </a>
+                                    {% if wishlist.is_shareable %}
+                                    <a href="{{ wishlist.get_shared_url }}" class="dropdown-item clipboard-item">
+                                        {% trans 'Copy shared link' %}
```
solarissmoke, line 27: I think "copy sharing link" makes more sense.

solarissmoke (changes requested):
Couple of minor string changes requested, otherwise looks good.

joeyjurjens:
> Couple of minor string changes requested, otherwise looks good.

I've applied your change requests

solarissmoke (approved):

PR #3911, opened 2022-04-11
Add basket post data to the form when invalid

viggo-devries:
So when you have a required checkbox option and you submit the add to basket form the selected / filled out data is not kept so you have to fill out everything over again.

For normal use this isn't that annoying but when adding lots of options (and making custom option widgets which are more complicated and need backend validation) this will become an issue.

### DIFF
--- a/src/oscar/apps/basket/views.py
+++ b/src/oscar/apps/basket/views.py
@@ -1,5 +1,6 @@
 from django import shortcuts
 from django.contrib import messages
+from django.contrib.sessions.serializers import JSONSerializer
 from django.core.exceptions import ObjectDoesNotExist
 from django.http import JsonResponse, QueryDict
 from django.shortcuts import redirect
@@ -331,6 +332,13 @@ class BasketAddView(FormView):
             msgs.append(error.as_text())
         clean_msgs = [m.replace('* ', '') for m in msgs if m.startswith('* ')]
         messages.error(self.request, ",".join(clean_msgs))
+
+        # We serialize the POST data with JSONSerializer before adding it to the session.
+        # Without this, we could expose the site to a security vulnerability
+        # if the SESSION_SERIALIZER has been configured to 'django.contrib.sessions.serializers.PickleSerializer'.
+        # see: https://docs.djangoproject.com/en/3.2/topics/http/sessions/#cookie-session-backend
+        serialized_data = JSONSerializer().dumps(self.request.POST)
+        self.request.session["add_to_basket_form_post_data_%s" % self.product.pk] = serialized_data.decode("latin-1")
 
         return redirect_to_referrer(self.request, 'basket:summary')
 
--- a/src/oscar/templatetags/basket_tags.py
+++ b/src/oscar/templatetags/basket_tags.py
@@ -1,4 +1,5 @@
 from django import template
+from django.contrib.sessions.serializers import JSONSerializer
 
 from oscar.core.loading import get_class, get_model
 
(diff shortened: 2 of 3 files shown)


specialunderwear (changes requested):
This needs tests for the view, showing that the form is populated with the posted data while displaying an error/

src/oscar/apps/basket/views.py
```diff
@@ -326,12 +326,7 @@ def get_form_kwargs(self):
         return kwargs
 
     def form_invalid(self, form):
-        msgs = []
-        for error in form.errors.values():
-            msgs.append(error.as_text())
-        clean_msgs = [m.replace('* ', '') for m in msgs if m.startswith('* ')]
-        messages.error(self.request, ",".join(clean_msgs))
-
+        self.request.session["add_to_basket_form_post_data"] = self.request.POST
```
specialunderwear, line 329: Since you are now storing arbitrary posted data in the session, you are open for a remote execution attack:

https://docs.djangoproject.com/en/3.2/topics/http/sessions/#cookie-session-backend

Please add a check that settings.SESSION_SERIALIZER has been set to ``django.contrib.sessions.serializers.JSONSerializer`` and refuse to store the POST data in the session if not so. Also add a warning.

specialunderwear (changes requested):

src/oscar/apps/basket/views.py
specialunderwear: maybe use the existing functionality if an unsafe session serializer has been set

viggo-devries: okay added the check

specialunderwear (dismissed):

src/oscar/apps/basket/views.py
```diff
@@ -332,6 +333,12 @@ def form_invalid(self, form):
         clean_msgs = [m.replace('* ', '') for m in msgs if m.startswith('* ')]
         messages.error(self.request, ",".join(clean_msgs))
 
+        # We explicitly serialize the post data into the session with the django's session JSONSerializer
+        # in case a different serializer is configured globally, this will create a security vulnerability
+        # https://docs.djangoproject.com/en/3.2/topics/http/sessions/#cookie-session-backend
```
specialunderwear, line 338: We serialize the POST data with JSONSerializer before adding it to the session.
Without this, we could expose us to a security vulnerability if the SESSION_SERIALIZER has been configured to  https://docs.djangoproject.com/en/3.2/topics/http/sessions/#cookie-session-backend

specialunderwear (approved):

src/oscar/apps/basket/views.py
```diff
@@ -332,6 +333,13 @@ def form_invalid(self, form):
         clean_msgs = [m.replace('* ', '') for m in msgs if m.startswith('* ')]
         messages.error(self.request, ",".join(clean_msgs))
 
+        # We serialize the POST data with JSONSerializer before adding it to the session.
+        # Without this, we could expose the site to a security vulnerability
+        # if the SESSION_SERIALIZER has been configured to 'django.contrib.sessions.serializers.PickleSerializer'.
+        # see: https://docs.djangoproject.com/en/3.2/topics/http/sessions/#cookie-session-backend
+        serialized_data = JSONSerializer().dumps(self.request.POST)
+        self.request.session["add_to_basket_form_post_data"] = serialized_data.decode("latin-1")
```
solarissmoke, line 341: I think this may be prone to race conditions because the session key is global. Imagine if two items are added to the basket at the same time (multiple tabs, multiple forms on a page) - it's possible for this response to end up going to the wrong form. Maybe we should suffix it with the product ID?

solarissmoke (changes requested):

src/oscar/apps/basket/views.py
viggo-devries: Yes I think this is a good idea. I will change it.

specialunderwear (approved):

joeyjurjens (approved):

solarissmoke (approved):

PR #3948, opened 2022-07-08
Only change order line status when status is available

viggo-devries:
It's possible for lines to have a completely different status than the order.
When a line is at an "end status" where it cannot change anymore this will result in the line being changed to something you might not want.

### DIFF
--- a/src/oscar/apps/order/abstract_models.py
+++ b/src/oscar/apps/order/abstract_models.py
@@ -135,8 +135,9 @@ class AbstractOrder(models.Model):
         self.status = new_status
         if new_status in self.cascade:
             for line in self.lines.all():
-                line.status = self.cascade[self.status]
-                line.save()
+                if new_status in line.available_statuses():
+                    line.status = self.cascade[self.status]
+                    line.save()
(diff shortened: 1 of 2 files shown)


specialunderwear (changes requested):
please add a test

sasha0:
Needs test.

viggo-devries:
hahah definitely tests, was curious if the current tests would still pass.

src/oscar/apps/order/abstract_models.py
```diff
@@ -135,8 +135,9 @@ def set_status(self, new_status):
         self.status = new_status
         if new_status in self.cascade:
             for line in self.lines.all():
-                line.status = self.cascade[self.status]
-                line.save()
+                if new_status in line.available_statuses():
```
specialunderwear, line 138: ```if self.cascade[new_status] in line.available_statuses():```

specialunderwear (changes requested):

joeyjurjens (approved):

specialunderwear (approved):

PR #3954, opened 2022-07-28
Bug AbsoluteDiscountBenefit

ebottos94:
Hi, according with the description in the class AbstractBenefit in file src/oscar/apps/offer/abstract_models.py in line 480, for the absolute discount the discount's logic is : "Discount is a fixed amount off of the product's value". 

In the file src/oscar/apps/offer/benefits.py instead the logic in the class AbsoluteDiscountBenefit is developed as if the discount were "Discount is a fixed amount off of the total amount of the shopping".

I assumed that the description "Discount is a fixed amount off of the product's value" is correct and i've change the AbsoluteDiscountBenefit's code so that it works as the description would like.

Thanks.


specialunderwear:
Hi, I think you better improve the description, because what you are doing now is change exsiting offers on sites that use this functionality. I think most customers would not enjoy it when suddenly their offers give a much larger discount then before after an update!

ebottos94:
@specialunderwear yes I've thought that but the description of the offer is enough clear now, so i don't think that the customers use this offer now. But you are rigth. Anyway, now the description does not reflect the logic, so if you don't want change the logic I think that the description must be changed!

specialunderwear:
Yes, hence is suggest improving the description.

specialunderwear (changes requested):
We can't change the logic, because that would change working of the offer in existing sites. So please change the description. If you do need the different functionality, I suggest adding a new type of offer that implements that.

viggo-devries:
@samar-hassan I think you need to edit a migration because the make migration test failed now.

viggo-devries:
@samar-hassan please edit the existing migration, a new migration for choices is not needed!

viggo-devries (approved):

PR #3966, opened 2022-08-19
Fixes a potential redirect loop in ProductDetailView if URL includes …

hutchison:
…certain characters

If the URL includes a colon ':' then the ProductDetailView is getting
trapped into a redirect loop.


joeyjurjens:
Hello, thank you for this PR. Could you possibly add a test that "proves" this works now.

specialunderwear (changes requested):
please add a test demonstrating the issue

specialunderwear:
superceded by #4106

PR #3995, opened 2022-10-19
Make sure to only save attributes when they have been saved.

specialunderwear:
This makes sure attributes on the parent are not needlessly saved to the child.

### DIFF
--- a/src/oscar/apps/catalogue/product_attributes.py
+++ b/src/oscar/apps/catalogue/product_attributes.py
@@ -1,4 +1,4 @@
-from django.core.exceptions import ValidationError
+from django.core.exceptions import ObjectDoesNotExist, ValidationError
 from django.utils.translation import gettext_lazy as _
 
 
@@ -66,4 +66,13 @@ class ProductAttributesContainer:
         for attribute in self.get_all_attributes():
             if hasattr(self, attribute.code):
                 value = getattr(self, attribute.code)
+
+                # only go and save values that have changed, don't do anything useless
+                try:
+                    attribute_value_current = self.get_value_by_attribute(attribute)
+                    if attribute_value_current.value == value:
+                        continue  # no new value needs to be saved
+                except ObjectDoesNotExist:
+                    pass  # there is no existing value, so a value needs to be saved.
+
                 attribute.save_value(self.product, value)


src/oscar/apps/catalogue/product_attributes.py
```diff
@@ -66,4 +66,13 @@ def save(self):
         for attribute in self.get_all_attributes():
             if hasattr(self, attribute.code):
                 value = getattr(self, attribute.code)
+
+                # only go and save values that have changed, don't do anything useless
```
joeyjurjens, line 70: This check already exists in the ``` _save_value``` method as well:
https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/catalogue/abstract_models.py#L927

So the comment doesn't really describe what it's actually fixing, so I think it's better to explain the issue here what you've put in the description of this PR: "This makes sure attributes on the parent are not needlessly saved to the child."

The solution seems good to me.

joeyjurjens (changes requested):

src/oscar/apps/catalogue/product_attributes.py
specialunderwear: You forgot to mention I should add a test too ;)

src/oscar/apps/catalogue/product_attributes.py
```diff
@@ -19,14 +19,35 @@ def __setstate__(self, state):
...
+        for v in self.get_values().select_related('attribute'):
+            attrs.setdefault(v.attribute.code, v.value)
 
     def refresh(self):
-        values = self.get_values().select_related('attribute')
-        for v in values:
+        for v in self.get_values().select_related('attribute'):
             setattr(self, v.attribute.code, v.value)
 
+    def __getattribute__(self, name):
```
joeyjurjens, line 40: Wouldn't it be better to call the `initialize` method on the top here if it isn't yet initialized `(self._initialized)` instead of doing this in in the attribute error?

eg
```python
def __getattribute__(self, name):
    if not self._initialized:
        self.initialize()
    return super().__getattribute__(name)
```

specialunderwear: Most definately not. __getattribute__ is called on all attribute accesses, even methods and even on ``self._initalized``. The point is we only want to ``initialize`` only when we are trying to access a product attribute, and that is only when the ``AttributeError`` happens. In all other cases it is not a product attribute that is being accessed. We arre trying to delay the initialization as much as possible to avoid making useless queries.

joeyjurjens (approved):

PR #3997, opened 2022-10-26
Show boolean not required error

specialunderwear:
It makes no sense to have a boolean attribute that is required, because it makes people have to check that box for every product they edit. Now let's also show people the error that will be thrown when that happens, for clarity.

### DIFF
--- a/src/oscar/templates/oscar/dashboard/catalogue/product_class_form.html
+++ b/src/oscar/templates/oscar/dashboard/catalogue/product_class_form.html
@@ -117,7 +117,7 @@
 
                                                         {% include 'oscar/dashboard/partials/form_field.html' with field=attribute_form.option_group nolabel=True %}
                                                     </td>
-                                                    <td>{% include 'oscar/dashboard/partials/form_field.html' with field=attribute_form.required nolabel=True %}</td>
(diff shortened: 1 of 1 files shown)


src/oscar/templates/oscar/dashboard/catalogue/product_class_form.html
```diff
@@ -117,7 +117,7 @@ <h3>{% trans "Product attributes" %}</h3>
 
                                                         {% include 'oscar/dashboard/partials/form_field.html' with field=attribute_form.option_group nolabel=True %}
                                                     </td>
-                                                    <td>{% include 'oscar/dashboard/partials/form_field.html' with field=attribute_form.required nolabel=True %}</td>
+                                                    <td>{% include 'oscar/dashboard/partials/form_field.html' with field=attribute_form.required nolabel=True %}{{ attribute_form.non_field_errors }}</td>
```
joeyjurjens, line 120: What's the error message that will be shown? Doesn't it make more sense to show non field errors at the top of the form, as it might also be an error not related to the "required" field.

specialunderwear: This might not seem to make much sense, but the only error that is a non field error that can happen here is the problem with a boolean field which should not be made required. At the moment, nothing is shown at all.

joeyjurjens (approved):

PR #4002, opened 2022-11-03
Update catalogue search_handlers.py

gwaidacher:
Fixed import order at line 8


gwaidacher:
Haystack search did not work, because SearchHandler was imported as SearchResultsPaginationMixin, which has no super class:
Error: ` object.__init__() takes exactly one argument (the instance to initialize)`

joeyjurjens (approved):

specialunderwear (approved):

Issue #3913, opened 2022-04-20
Product Class disable track stock level but stock record price should be required


eternaltot:
### Issue Summary

When product_class.track_stock is False. On Product Detail page price in StockRecord form is not required and can fill with blank but currency is still required. It can't buy that product also

### Steps to Reproduce
1. Uncheck Track stock level on Product Type Detail page.
2. Go to Product Detail > Stock and pricing.
3. Delete price in form.
4. Go to Product page on store.

It's still available but can't buy.
![image](https://user-images.githubusercontent.com/9781462/164200188-b45d44d0-35f2-4761-8571-75801d448e45.png)

specialunderwear:
I think the reason that price is not required, is that the strategy determines what the price should be. Most of the time that price will be taken from the stockrecord, but the source of the price could be anything, like:

```
from random import randint
randint(0, 4)
```

It might be you lucky day!

joeyjurjens:
Fixed in #4013

PR #4004, opened 2022-11-11
When trackstock is off make price required

Wessellix:
fixes issue: https://github.com/django-oscar/django-oscar/issues/3913

### DIFF
--- a/src/oscar/apps/dashboard/catalogue/forms.py
+++ b/src/oscar/apps/dashboard/catalogue/forms.py
@@ -97,6 +97,12 @@ class StockRecordForm(forms.ModelForm):
         if not self.user.is_staff:
             self.fields['partner'].queryset = self.user.partners.all()
 
+        # If not tracking stock, make price required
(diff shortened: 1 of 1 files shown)


joeyjurjens:
I don't think this is the right solution. A better solution in my opinion would be showing "Unavailable" when there's no price entered. It doesn't really make sense to require a price when the product doesn't use "track stock", as it has nothing to do with each other, in this case price should also be required when the product does track stock.

@viggo-devries @specialunderwear your thoughts?

specialunderwear:
I think the reason that price is not required, is that the strategy determines what the price should be. Most of the time that price will be taken from the stockrecord, but the source of the price could be anything, like:

from random import randint
randint(0, 4)
It might be you lucky day!

viggo-devries:
I agree with @specialunderwear, maybe if there's not price it should say "Free"?

specialunderwear:
> I agree with @specialunderwear, maybe if there's not price it should say "Free"?

no

PR #4005, opened 2022-11-15
show boolean option value to yes/no in stead of true/false

wlaw01:


joeyjurjens:
Can you also add unit tests here?
https://github.com/django-oscar/django-oscar/blob/1f84c57e0675dd3adef1f116cbf36c507017893a/tests/integration/catalogue/test_attributes.py#L10

src/oscar/apps/catalogue/abstract_models.py
```diff
@@ -1142,6 +1142,12 @@ def _entity_as_text(self):
         """
         return str(self.value)
 
+    @property
+    def _boolean_as_text(self):
+        if self.value:
+            return (_("Yes"))
```
specialunderwear, line 1148: why not ```_("Yes")```?

specialunderwear (changes requested):

src/oscar/apps/catalogue/abstract_models.py
wlaw01: i forgot to remove it. its fixed now

specialunderwear (approved):

PR #4010, opened 2022-11-22
Optimize range products reordering

samar-hassan:
If there are 20 products in the range, it takes 40 queries to change the order of a single product. With this PR it just takes 2 queries regardless of the number of products in the range.

### DIFF
--- a/tests/settings.py
+++ b/tests/settings.py
@@ -173,3 +173,9 @@ SECRET_KEY = 'notverysecret'
 DEFAULT_HASHING_ALGORITHM = 'sha1'
 TEST_RUNNER = 'django.test.runner.DiscoverRunner'
 FIXTURE_DIRS = [location('unit/fixtures')]
+
+# Try and import local settings which can be used to override any of the above.
+try:
+    from tests.settings_local import *
+except ImportError:
+    pass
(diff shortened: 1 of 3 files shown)


joeyjurjens (approved):

specialunderwear (changes requested):
It seems a test needs to be added because this code is currently untested.
Would you be so kind to add a test that proves your code works correctly?

tests/settings.py
```diff
@@ -173,3 +173,9 @@
 DEFAULT_HASHING_ALGORITHM = 'sha1'
 TEST_RUNNER = 'django.test.runner.DiscoverRunner'
 FIXTURE_DIRS = [location('unit/fixtures')]
+
+# Try and import local settings which can be used to override any of the above.
+try:
+    from tests.settings_local import *
```
specialunderwear, line 179: I think you  committed this by accident?

specialunderwear (changes requested):

tests/settings.py
samar-hassan: I added this on purpose. How do you update the test database credentials?

viggo-devries: This is not something we would want in the test settings, when testing with different databases please add your own settings and don't push them to your branch.

viggo-devries (changes requested):

specialunderwear (approved):

viggo-devries (approved):

PR #4013, opened 2022-11-23
Make products without price unable to be added to basket

Wessellix:
fixes issue: #3913

### DIFF
--- a/src/oscar/templates/oscar/catalogue/partials/add_to_basket_form.html
+++ b/src/oscar/templates/oscar/catalogue/partials/add_to_basket_form.html
@@ -5,7 +5,7 @@
 
 {% purchase_info_for_product request product as session %}
 
-{% if session.availability.is_available_to_buy %}
+{% if session.availability.is_available_to_buy and session.price.exists %}
     {% basket_form request product as basket_form %}
     <form id="add_to_basket_form" action="{% url 'basket:add' pk=product.pk  %}" method="post" class="add-to-basket mb-0">
         {% csrf_token %}
--- a/src/oscar/templates/oscar/catalogue/partials/add_to_basket_form_compact.html
+++ b/src/oscar/templates/oscar/catalogue/partials/add_to_basket_form_compact.html
@@ -4,7 +4,7 @@
 
 {% purchase_info_for_product request product as session %}
 
-{% if session.availability.is_available_to_buy %}
+{% if session.availability.is_available_to_buy and session.price.exists %}
     {% basket_form request product 'single' as basket_form %}
     <form action="{% url 'basket:add' pk=product.pk %}" method="post">
         {% csrf_token %}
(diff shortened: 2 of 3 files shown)


src/oscar/apps/partner/strategy.py
```diff
@@ -210,6 +210,12 @@ class StockRequired(object):
     def availability_policy(self, product, stockrecord):
         if not stockrecord:
             return Unavailable()
+        if not stockrecord.price:
```
specialunderwear, line 213: Volgens mij ging het issue over ``track_stock``, wat hieronder staat. Dan was ie altijd available.

specialunderwear (changes requested):

src/oscar/apps/partner/strategy.py
```diff
@@ -210,14 +210,11 @@ class StockRequired(object):
...
             return Unavailable()
-        if not stockrecord.price:
-            try:
-                if stockrecord.price == None:
-                    return Unavailable()
-            except:
-                return Available()
         if not product.get_product_class().track_stock:
-            return Available()
+            if stockrecord.price == None:
```
specialunderwear, line 214: We check for None in python like this:
```
if stockrecord.price is None:
```
That is because None is a singleton object there exists only 1 instance of NoneType. You can not check for equality using ``is`` with other types!

specialunderwear (changes requested):

src/oscar/apps/partner/strategy.py
```diff
@@ -211,7 +211,10 @@ def availability_policy(self, product, stockrecord):
         if not stockrecord:
             return Unavailable()
         if not product.get_product_class().track_stock:
-            return Available()
+            if stockrecord.price is None:
```
joeyjurjens, line 214: I don't think this is the right solution. Even when the price is empty on a stockrecord, the product could still have a price through the strategy.

The purchaseinfo "object" has a price, availability and a stockrecord
https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/partner/strategy.py#L14

The price is any of those (assuming a project has no custom ones)
https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/partner/prices.py

So whenever a product has no price (even through the strategy) it will have the Unavailable price object
https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/partner/prices.py#L40

So I see two options

1. Change the template, to make sure it checks whether a price exists
https://github.com/django-oscar/django-oscar/blob/master/src/oscar/templates/oscar/catalogue/partials/add_to_basket_form.html#L8
... and session.price.exists

2. Pass the price as argument in this method
https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/partner/availability.py#L36

Which you would have to do here: https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/partner/strategy.py#L126

And check session.price.exists here instead.

I think solution 2 would be cleaner, as it doesn't require a template change but it does require a change in the strategy. 

@specialunderwear what do you think?

specialunderwear: I think that the strategy should not be changed to pass the price because if needed we can just acces the price via ``self.pricing_policy`` so I think that the template must be changed to have better defaults. @joeyjurjens

Issue #2637, opened 2018-03-06
Products with options can be oversold


ad-65:
When a product is configured with class or product options (not attributes) and tracks stock it possible to oversell the product. This behaviour is visible in the sandbox and in Oscar 1.5.

To reproduce:

1. Create a product option, for example, a message to print on a t-shirt.
2. Create a product type that uses this option in the dashboard.
3. Buy the product multiple times with different messages.
4. Edit the basket, changing the quantities to more than are available. 
5. Stock validation will allow product to be oversold.

![basket_screenshot](https://user-images.githubusercontent.com/3200517/37008289-9782c45a-209e-11e8-8bc7-3b31f49c3359.png)

If I am not mistaken this is because the availability validation happens at the basket line level and does not take the overall basket state into account. The `is_purchase_permitted()` check passes for each line item but should not pass as a whole because there are only 10 t-shirts available for purchase but 13 in the basket.

The options force the same product to be a separate line item (as it should be) but this means that line item validation will not work because there can be multiple lines with the same product. I am not sure the best way to fix this as it requires basket level validation and I had a quick look at the code but did not see a hook for adding this. Any suggestions?

ad-65:
Thanks for the response.

Yeah, that could work. The basket would need to be checked in a number of places to prevent basket errors:

1. When adding a new product to the basket from the catalogue pages. This is already covered because form validation uses `Basket.product_quantity()` which sums up all line quantities with the same product.
2. When editing the basket line items. The user can be making changes to multiple lines with the same product, so we have to get the basket quantities, the form quantities, and the availability. For example, a product may have 10 items available and the user currently has 5 with "Message one" and 5 with "Message two". The basket could be edited to contain 4 with "Message one" and 6 with "Message two" and should validate properly.
3. At all checkout stages to make sure the product is still available in case any stock was consumed since the product was added to the basket.

`Basket.is_quantity_allowed()` isn't currently called anywhere except during the re-order view and during the individual basket form `clean_quantity()`. I think adding a `line` argument to `Basket.is_quantity_allowed()` and then making sure it is called during form/formset cleaning would work for items 1 & 2.

For item 3 it gets a bit tricky as you said. Would the validation have to rely on the availability policy, i.e., `StockRequired.is_purchase_permitted()` or is there a good place to call `Basket.is_quantity_allowed()`? Maybe in`CheckoutSessionMixin.check_basket_is_valid()`?

ad-65:
[...]

PR #4015, opened 2022-11-28
if a product has a option it cannot be oversold

Wessellix:
fixes issue: #2637

### DIFF
--- a/src/oscar/apps/basket/abstract_models.py
+++ b/src/oscar/apps/basket/abstract_models.py
@@ -148,13 +148,18 @@ class AbstractBasket(models.Model):
             return max_allowed, basket_threshold
         return None, None
 
-    def is_quantity_allowed(self, qty):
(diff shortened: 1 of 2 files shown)


src/oscar/apps/basket/abstract_models.py
```diff
@@ -148,20 +148,34 @@ def max_allowed_quantity(self):
...
 
-    def is_quantity_allowed(self, qty):
+    def is_quantity_allowed(self, qty, line=None):
         """
         Test whether the passed quantity of items can be added to the basket
         """
         # We enforce a max threshold to prevent a DOS attack via the offers
         # system.
         max_allowed, basket_threshold = self.max_allowed_quantity()
+        if line:
```
specialunderwear, line 158:
```
If line is not None:
```

specialunderwear (changes requested):
A small change needed

PR #4041, opened 2023-01-25
Update abstract_models.py

gwaidacher:
We have a database with > 3,000.000 products and a simple offer with 40 products in it. The original query with Q - joins took over 10 seconds in time. By checking, if there are only included_products in the range (no parents in products, no excluded_products and no classes in range) we can simplify the query to only return included_products without joins.

Speedup factor is > 1000, request time from > 11.000 ms down below 10 ms

### DIFF
--- a/src/oscar/apps/offer/abstract_models.py
+++ b/src/oscar/apps/offer/abstract_models.py
@@ -964,6 +964,25 @@ class AbstractRange(models.Model):
 
         return self.product_queryset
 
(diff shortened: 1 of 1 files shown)


gwaidacher:
https://github.com/django-oscar/django-oscar/pull/4042 fixes https://github.com/django-oscar/django-oscar/issues/4043 so I close this pull request.

PR #4042, opened 2023-01-27
Update abstract_models.py: speeding up Range.product_queryset()

gwaidacher:
Reduce joins in Range.product_queryset() if there are no classes in the range.
Reduce even more joins, if the included products have no children.
Speedup: 1000 times faster if there are no children!

see https://github.com/django-oscar/django-oscar/issues/4043 for more details, this patch fixes the issue

### DIFF
--- a/src/oscar/apps/offer/abstract_models.py
+++ b/src/oscar/apps/offer/abstract_models.py
@@ -963,11 +963,12 @@ class AbstractRange(models.Model):
             return self.proxy.all_products()
 
         return self.product_queryset
-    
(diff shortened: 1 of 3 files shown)


gwaidacher:
Fixes https://github.com/django-oscar/django-oscar/issues/4043

gwaidacher:
> @gwaidacher please rebase this branch against the main branch since you have changes in this PR that are already merged in you other PR

Hi!
 I rebased it in #4066, closing this PR.

PR #4044, opened 2023-01-27
[FIX] Show address update button only if there exists one

samar-hassan:
If a shipping address button is not associated with the order, update button should not be visible. Currently clicking on the button shows 404 page if an address is not assigned to the order.


viggo-devries:
@samar-hassan Thanks for contributing!
In what case would there not be a shipping address? When it's deleted manually maybe?

samar-hassan:
> @samar-hassan Thanks for contributing! In what case would there not be a shipping address? When it's deleted manually maybe?

Some products do not require shipping. In that case, if we place an order and go to the order detail page, the "update shipping address" should not be displayed.

viggo-devries (approved):

Issue #4049, opened 2023-02-03
Range Model throws ProgrammingError at /dashboard/ranges/


mghiglieri:
### Issue Summary

Opening dashboard/ranges/
ProgrammingError at /shop/dashboard/ranges/
(1064, 'You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near \'."id" FROM "catalogue_category" AS "CATALOGUE_CATEGORY_BASE"\nLEFT JOIN "catal...\' at line 2')

Adding a voucher to a cart
ProgrammingError at /shop/basket/vouchers/add/
(1064, 'You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near \'."id" FROM "catalogue_category" AS "CATALOGUE_CATEGORY_BASE"\nLEFT JOIN "catal...\' at line 2')

There seems to be some problems with the JOINS

The error is caused in site-packages/oscar/apps/offer/abstract_models.py, line 952, in num_products

```
>>> Product.objects.filter(Q(categories__in=expanded_range_categories))
Traceback (most recent call last):
  File "/var/envs/test.sicher3.de/lib/python3.8/site-packages/django/db/backends/utils.py", line 89, in _execute
    return self.cursor.execute(sql, params)
  File "/var/envs/test.sicher3.de/lib/python3.8/site-packages/django/db/backends/mysql/base.py", line 75, in execute
    return self.cursor.execute(query, args)
  File "/var/envs/test.sicher3.de/lib/python3.8/site-packages/MySQLdb/cursors.py", line 206, in execute
    res = self._query(query)
  File "/var/envs/test.sicher3.de/lib/python3.8/site-packages/MySQLdb/cursors.py", line 319, in _query
    db.query(q)
  File "/var/envs/test.sicher3.de/lib/python3.8/site-packages/MySQLdb/connections.py", line 259, in query
    if self._binary_prefix:
MySQLdb._exceptions.ProgrammingError: (1064, 'You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near \'."id" FROM "catalogue_category" AS "CATALOGUE_CATEGORY_BASE"\nLEFT JOIN "catal...\' at line 2')

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/usr/lib/python3.8/code.py", line 90, in runcode
    exec(code, self.locals)
  File "<console>", line 1, in <module>
[...]

Issue #4032, opened 2022-12-23
MySQL error on profile page


st8st8:
Found a bug? Please fill out the sections below.

### Issue Summary

Viewing the account profile page (/account/profile) brings about this error from Django:
ProgrammingError at /accounts/profile/
(1064, 'You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near \'."id" FROM "catalogue_category" AS "CATALOGUE_CATEGORY_BASE"\nLEFT JOIN "catalogu\' at line 2')

### Steps to Reproduce

Log in to Oscar
View the account profile page 

Any other relevant information. For example, why do you consider this a bug and what did you expect to happen instead?

It seems like a query template isn't being resolved here.  Here's the full query from the MySQL query log:  The template seems to be defined in apps/catalogue/expressions.py 

[...]

PR #4050, opened 2023-02-03
fix for #4049

mghiglieri:
Fix for #4049 and probably for #4032


viggo-devries:
@mghiglieri you have lint errors, probably the removed newline.

mghiglieri:
@viggo-devries I added a newline at the end.

src/oscar/apps/catalogue/expressions.py
```diff
@@ -32,3 +32,4 @@ class ExpandDownwardsCategoryQueryset(Subquery):
 
     def as_sqlite(self, compiler, connection):
         return super().as_sql(compiler, connection, self.template[1:-1])
+
```
viggo-devries, line 35: Still a change here, please remove this change.

viggo-devries (changes requested):

src/oscar/apps/catalogue/expressions.py
mghiglieri: Sorry, just changed it. No changes after line 19.

viggo-devries (approved):

viggo-devries:
@mghiglieri unfortunately this breaks gives a syntax error in postgres.

specialunderwear:
you would have to make this work in both postgres and mysql @mghiglieri  this was reverted.

PR #4053, opened 2023-02-20
Exclude non-public Products from Offers

gwaidacher:
AbstractRange.product_queryset() provides non-public Products.

added `.filter(is_public=True)` in AbstractRange.product_queryset()

### DIFF
--- a/src/oscar/apps/offer/abstract_models.py
+++ b/src/oscar/apps/offer/abstract_models.py
@@ -998,7 +998,6 @@ class AbstractRange(models.Model):
                 ~Q(excludes=self),
                 ~Q(parent__excludes=self)
             )
-        # Filter out non-public Products
         return selected_products.distinct()
 
     @property
--- a/src/oscar/apps/offer/abstract_models.py
+++ b/src/oscar/apps/offer/abstract_models.py
@@ -998,7 +998,7 @@ class AbstractRange(models.Model):
                 ~Q(excludes=self),
                 ~Q(parent__excludes=self)
             )
-        # exclude non-public Products
+        # Filter out non-public Products
         return selected_products.filter(is_public=True).distinct()
 
     @property
--- a/src/oscar/apps/offer/abstract_models.py
+++ b/src/oscar/apps/offer/abstract_models.py
@@ -452,7 +452,7 @@ class AbstractConditionalOffer(models.Model):
 
         queryset = self.condition.range.all_products()
         return queryset.filter(is_discountable=True).exclude(
-            structure=Product.CHILD)
+            structure=Product.CHILD).browsable()
 
     @cached_property
     def combined_offers(self):
--- a/src/oscar/apps/offer/abstract_models.py
+++ b/src/oscar/apps/offer/abstract_models.py
@@ -971,7 +971,7 @@ class AbstractRange(models.Model):
 
         if self.includes_all_products:
             # Filter out blacklisted and non-public products
-            return Product.objects.filter(is_public=True).exclude(
+            return Product.objects.browsable().exclude(
                 id__in=self.excluded_products.values("id")
             )
 
@@ -999,7 +999,7 @@ class AbstractRange(models.Model):
                 ~Q(parent__excludes=self)
             )
         # Filter out non-public Products
-        return selected_products.filter(is_public=True).distinct()
+        return selected_products.browsable().distinct()
 
     @property
     def is_editable(self):
--- a/src/oscar/apps/offer/abstract_models.py
+++ b/src/oscar/apps/offer/abstract_models.py
@@ -451,8 +451,7 @@ class AbstractConditionalOffer(models.Model):
             return Product.objects.none()
 
         queryset = self.condition.range.all_products()
-        return queryset.filter(is_discountable=True).exclude(
-            structure=Product.CHILD).browsable()
+        return queryset.filter(is_discountable=True).browsable()
 
     @cached_property
     def combined_offers(self):
--- a/src/oscar/apps/offer/abstract_models.py
+++ b/src/oscar/apps/offer/abstract_models.py
@@ -962,7 +962,7 @@ class AbstractRange(models.Model):
         if self.proxy:
             return self.proxy.all_products()
 
-        return self.product_queryset()
+        return self.product_queryset
 
     @cached_property
     def product_queryset(self):
--- a/src/oscar/apps/offer/abstract_models.py
+++ b/src/oscar/apps/offer/abstract_models.py
@@ -971,7 +971,7 @@ class AbstractRange(models.Model):
 
         if self.includes_all_products:
             # Filter out blacklisted products
-            return Product.objects.all().exclude(
+            return Product.objects.filter(is_public=True).exclude(
                 id__in=self.excluded_products.values("id")
             )
 
@@ -999,7 +999,7 @@ class AbstractRange(models.Model):
                 ~Q(parent__excludes=self)
             )
 
-        return selected_products.distinct()
+        return selected_products.filter(is_public=True).distinct()
 
     @property
     def is_editable(self):
--- a/src/oscar/apps/offer/abstract_models.py
+++ b/src/oscar/apps/offer/abstract_models.py
@@ -970,7 +970,7 @@ class AbstractRange(models.Model):
         Product = self.included_products.model
 
         if self.includes_all_products:
-            # Filter out blacklisted products
+            # Filter out blacklisted and non-public products
             return Product.objects.filter(is_public=True).exclude(
                 id__in=self.excluded_products.values("id")
             )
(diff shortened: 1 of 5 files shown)


joeyjurjens (approved):

viggo-devries:
@gwaidacher could you please provide a unit test that shows this is working?

viggo-devries:
@joeyjurjens maybe this should use the .browsable method ? It could result in some weird results when customising the browsable method though.

gwaidacher:
> @joeyjurjens maybe this should use the .browsable method ? It could result in some weird results when customising the browsable method though.

This is exactly what I thought about this morning, I will do so

joeyjurjens:
> @joeyjurjens maybe this should use the .browsable method ? It could result in some weird results when customising the browsable method though.

Yeah I thought about that as well, but since products without categories could technically still be ordered, I thought I'd let it be. But it's probably best to use .browsable() to keep consistency, people can always override the browsable() implementation.

gwaidacher:
> @gwaidacher could you please provide a unit test that shows this is working?

where should this test be located?

viggo-devries:
> > @gwaidacher could you please provide a unit test that shows this is working?
> 
> where should this test be located?

Probably somewhere in here? https://github.com/django-oscar/django-oscar/blob/master/tests/integration/offer/test_range.py#L259

gwaidacher:
> > @joeyjurjens maybe this should use the .browsable method ? It could result in some weird results when customising the browsable method though.
> 
> Yeah I thought about that as well, but since products without categories could technically still be ordered, I thought I'd let it be. But it's probably best to use .browsable() to keep consistency, people can always override the browsable() implementation.

Tests failed @tests/functional/dashboard/test_range.py in test_adding_child_does_not_add_parent() using browsable()

I worried about that before, since browsable() excludes child products... should I return to the previous Version using `filter(is_public=True)`

{{{
    def browsable(self):
        """
        Excludes non-canonical products and non-public products
        """
        return self.filter(parent=None, is_public=True)
}}}

tests/integration/offer/test_range.py
```diff
@@ -280,6 +280,14 @@ def test_contains_product(self):
         ranges = models.Range.objects.contains_product(self.prod)
         self.assertEqual(ranges.count(), 2, "Both ranges should contain the product")
 
+    def test_not_contains_non_public_product(self):
+        product = create_product()
```
viggo-devries, line 284: create_product(is_public=False)

tests/integration/offer/test_range.py
```diff
@@ -280,6 +280,14 @@ def test_contains_product(self):
...
+        self.range.add_product(product)
+        product.is_public = False
+        product.save()
+        self.assertFalse(product in self.range.all_products())
+        self.range.remove_product(product)
```
viggo-devries, line 289: you don't need to remove the product here, it's just test data

viggo-devries (changes requested):

gwaidacher:
> > @joeyjurjens maybe this should use the .browsable method ? It could result in some weird results when customising the browsable method though.
> 
> Yeah I thought about that as well, but since products without categories could technically still be ordered, I thought I'd let it be. But it's probably best to use .browsable() to keep consistency, people can always override the browsable() implementation.

Because of the failure in dashboard-integration test I think it would be much better to exclude them later in apps/offer/views.py, OfferDetailView.get_queryset(): return self.offer.products().browsable()

viggo-devries:
> > > @joeyjurjens maybe this should use the .browsable method ? It could result in some weird results when customising the browsable method though.
> > 
> > 
> > Yeah I thought about that as well, but since products without categories could technically still be ordered, I thought I'd let it be. But it's probably best to use .browsable() to keep consistency, people can always override the browsable() implementation.
> 
> Because of the failure in dashboard-integration test I think it would be much better to exclude them later in apps/offer/views.py, OfferDetailView.get_queryset(): return self.offer.products().browsable()

I discussed this with Joey really quickly and I think we're approaching this the wrong way, you are trying to remove non public products from the Range detail view right? I didn't see the original issue you were basing this on.
That being said I think you should add .browsable right here:
https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/offer/abstract_models.py#L454

And then add a test that check if the conditional offer products method returns only browsable products.

viggo-devries:
> > > @joeyjurjens maybe this should use the .browsable method ? It could result in some weird results when customising the browsable method though.
> > 
> > 
> > Yeah I thought about that as well, but since products without categories could technically still be ordered, I thought I'd let it be. But it's probably best to use .browsable() to keep consistency, people can always override the browsable() implementation.
> 
> Because of the failure in dashboard-integration test I think it would be much better to exclude them later in apps/offer/views.py, OfferDetailView.get_queryset(): return self.offer.products().browsable()

HA just what I thought.

src/oscar/apps/offer/views.py
```diff
@@ -49,7 +49,7 @@ def get_queryset(self):
         Return a queryset of all :py:class:`Product <oscar.apps.catalogue.abstract_models.AbstractProduct>`
         instances related to the :py:class:`ConditionalOffer <oscar.apps.offer.abstract_models.AbstractConditionalOffer>`.
         """  # noqa
-        return self.offer.products()
+        return self.offer.products().browsable()
```
viggo-devries, line 52: please do this on the model method

src/oscar/templates/oscar/offer/list.html
```diff
@@ -34,7 +34,7 @@ <h2>{{ offer.name }}</h2>
         {% endif %}
         {% if offer.has_products %}
             <ul class="row list-unstyled mx-0">
-                {% for product in offer.products|slice:":12" %}
+                {% for product in offer.products.browsable|slice:":12" %}
```
viggo-devries, line 37: if you do .browsable on the model method this is not needed.

src/oscar/apps/offer/abstract_models.py
```diff
@@ -998,7 +998,7 @@ def product_queryset(self):
                 ~Q(excludes=self),
                 ~Q(parent__excludes=self)
             )
-
+        # Filter out non-public Products
```
viggo-devries, line 1001: remove comment.

viggo-devries (changes requested):

src/oscar/apps/offer/abstract_models.py
gwaidacher: done

src/oscar/apps/offer/views.py
gwaidacher: done

src/oscar/apps/offer/abstract_models.py
```diff
@@ -452,7 +452,7 @@ def products(self):
 
         queryset = self.condition.range.all_products()
         return queryset.filter(is_discountable=True).exclude(
```
viggo-devries, line 454: En then you can remove the .exclude since this is also done in browsable :)

src/oscar/apps/offer/abstract_models.py
```diff
@@ -998,7 +998,6 @@ def product_queryset(self):
                 ~Q(excludes=self),
                 ~Q(parent__excludes=self)
             )
-
```
viggo-devries, line 1001: no newline

viggo-devries (changes requested):

src/oscar/apps/offer/abstract_models.py
viggo-devries: https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/catalogue/managers.py#L108

see parent=None is the same as .exclude(structure=CHILD)

gwaidacher:
I think we got it the right way round now, sometimes I think too complicated while there exists an easier way to solve a problem...

viggo-devries:
> I think we got it the right way round now, sometimes I think too complicated while there exists an easier way to solve a problem...

almost, you now removed .browsable from the range detail view, this does not use the offer.products method.
https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/offer/views.py#L71

I do think this should be in place here otherwise the range detail will stil show non public products.

viggo-devries:
please also provide test for the range view, cannot be merged because the hit target was not met...

tests/integration/offer/test_offer.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+class TestOffer(TestCase):
+
+    def test_non_public_product_not_in_offer(self):
+        offer = ConditionalOfferFactory()
+        product = create_product(is_public=False)
+        offer.condition.range.add_product(product)
+        self.assertFalse(product in offer.products())
+        self.assertFalse(
```
viggo-devries: This is not sufficient unfortunately, you need to test the RangeDetail view. this is just executing the code you added to the view.

viggo-devries (changes requested):

viggo-devries (approved):

viggo-devries (approved):

Issue #4043, opened 2023-01-27
ConditionalOffer.products() queries Product children and exludes them again.


gwaidacher:
Hello!

### Issue Summary

I had a problem with requests concerning Offers taking a lot of time:

We have a database of > 3,000.000 products and a simple ConditionalOffer with 40 products without children added and no included classes and categories.

OfferList and OfferDetail request time was over 11 seconds, which is unacceptable. 

Digging into this issue, I found out that in Range.product_queryset() Q-joins to collect Products' children are passed into the query:

offer.abstract_models.py, Lines 988 & 997 in AbstractRange.product_queryset():

```
        | Q(parent__includes=self),
```
 (note: this will later be called "children-join")

At last, in  ConditionalOffer.products() these children are excluded again, so the included children-joins seem to be unnecessary for me:

offer.abstract_models.py, Line 454 in AbstractConditionalOffer.products():

```
    return queryset.filter(is_discountable=True).exclude(
        structure=Product.CHILD)
```

Caching the products with `@cached_property` works with a short list of products, but it`s useless if the OfferDetailView is paginated. Getting back from page 2 to page 1 the formerly cached page 1 is requested again with > 11 seconds.

I committed a patch https://github.com/django-oscar/django-oscar/pull/4042, which reduces the query time significant if there are no classes in the range; if the included products have no children, the remaining `LEFT OUTER JOIN`s are omitted, so the time for gathering the merge joins (~ 8 seconds in the original query) is saved. The remaining INNER JOINS are insignificant in time.

My Tests worked, reducing query time from ~ 11 seconds to far below 1 second.

Speedup factor is > 1000 !

### Steps to Reproduce

* Create a Database with 3,000.000 products without children
* Create a simple ConditionalOffer with 40 products
* Query OfferListView and OfferDetailView

### Technical details

* Python version: 3.9.2
* Django version: 3.2.16
* Oscar version: 3.2.0

gwaidacher:
see #4066

PR #4066, opened 2023-03-03
Speed up Range.product_queryset()

gwaidacher:
Reduce joins in Range.product_queryset() if there are no classes in the range and / or included products have no children. 

see #4043 for more details, this patch fixes the issue

rebased #4042

### DIFF
--- a/src/oscar/apps/offer/abstract_models.py
+++ b/src/oscar/apps/offer/abstract_models.py
@@ -984,7 +984,8 @@ class AbstractRange(models.Model):
                 ~Q(parent__excludes=self)
             )
         # check if products in included categories have children
-        if self.included_categories.exclude(product__parent_id=None).exists():
+        if self.included_categories.filter(
+                product__structure=Product.CHILD).exists():
             return Product.objects.filter(
                 Q(categories__in=expanded_range_categories)
                 | Q(includes=self)
@@ -1017,7 +1018,7 @@ class AbstractRange(models.Model):
                 ~Q(parent__excludes=self)
             )
         # check if included products have children
-        if self.included_products.exclude(parent_id=None).exists():
+        if self.included_products.filter(structure=Product.CHILD).exists():
             return Product.objects.filter(
                 Q(includes=self)
                 | Q(parent__includes=self),
--- a/src/oscar/apps/offer/abstract_models.py
+++ b/src/oscar/apps/offer/abstract_models.py
@@ -984,7 +984,7 @@ class AbstractRange(models.Model):
                 ~Q(parent__excludes=self)
             )
         # check if products in included categories have children
-        if self.included_categories.exclude(product__children=None).exists():
+        if self.included_categories.filter(product__structure='child').exists():
             return Product.objects.filter(
                 Q(categories__in=expanded_range_categories)
                 | Q(includes=self)
@@ -1017,7 +1017,7 @@ class AbstractRange(models.Model):
                 ~Q(parent__excludes=self)
             )
         # check if included products have children
-        if self.included_products.exclude(children=None).exists():
+        if self.included_products.filter(structure='child').exists():
             return Product.objects.filter(
                 Q(includes=self)
                 | Q(parent__includes=self),
--- a/src/oscar/apps/offer/abstract_models.py
+++ b/src/oscar/apps/offer/abstract_models.py
@@ -1051,8 +1051,8 @@ class AbstractRange(models.Model):
                 id__in=self.excluded_products.values("id")
             )
         # start with empty filter (if included_products do not exist)
-        _filter = Q(True)
-        _exclude_filter = Q(True)
+        _filter = Q()
+        _exclude_filter = Q()
         # extend filter for included products
         if self.included_products.exists():
             _filter |= Q(includes=self)
--- a/src/oscar/apps/offer/abstract_models.py
+++ b/src/oscar/apps/offer/abstract_models.py
@@ -984,7 +984,7 @@ class AbstractRange(models.Model):
                 ~Q(parent__excludes=self)
             )
         # check if products in included categories have children
-        if self.included_categories.filter(product__structure='child').exists():
+        if self.included_categories.exclude(product__parent_id=None).exists():
             return Product.objects.filter(
                 Q(categories__in=expanded_range_categories)
                 | Q(includes=self)
@@ -1017,7 +1017,7 @@ class AbstractRange(models.Model):
                 ~Q(parent__excludes=self)
             )
         # check if included products have children
-        if self.included_products.filter(structure='child').exists():
+        if self.included_products.exclude(parent_id=None).exists():
             return Product.objects.filter(
                 Q(includes=self)
                 | Q(parent__includes=self),
--- a/src/oscar/apps/offer/abstract_models.py
+++ b/src/oscar/apps/offer/abstract_models.py
@@ -963,6 +963,89 @@ class AbstractRange(models.Model):
 
         return self.product_queryset
 
+    def included_categories_queryset(self):
+        """
+        reduce joins if there are no classes in the range
+        or if included products have no children
+        """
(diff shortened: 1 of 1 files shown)


gwaidacher:
Could you please explain the joins over parent_ids (see #4043) - is this used when a child product is addet to the range, so it's parent is automatically included and shown in the offer, the child is excluded in the offer via browsable()?

gwaidacher:
Thank you, specialunderwear, for your detailed answer.

> I think it would be much easier to just offer the browsable method on range and alter the offers.
The browsable() method in RangeDetailView was already merged from my PR into master a week ago, see https://github.com/django-oscar/django-oscar/pull/4053 "Exclude non-public Products from Offer - https://github.com/django-oscar/django-oscar/commit/75ed8827bd9a1c2a7613e41f7f08974e3c747530

We have a demo-shop running at https://shop.acat.cc, where my patch is running; the speeded up offer is at https://shop22/en/offers/oster-2023/

Variants are not planned in our shop in the near future; we have a very fast designated database server with SSD disks, and we will keep the changes in the forked range queryset to avoid response times over 10 seconds in our offers.

Concerning a worst case scenario, we would not copy 3000000 into included_products, but picking up all products via range.includes_all_products having unwanted in excluded_products - this would be the first response in range.product_queryset(), not touching the queries with tests if classes or children exist.

Adding classes to the range would not worsen the performance by the query  `if self.classes.exists()`.

The worst case is - like before - if self.included_categories.exists() and self.classes.exists() leading to the original query with the response time of > 10 seconds.  This scenario is not worsened significant by this two queries, which are very fast.

All other scenarios have reduced joins and the single check for children is now much faster - from my opinion all scenarios will now gain performance:

I just optimized the query for children:
`if self.included_products.exclude(children=None).exists()` takes a view seconds, so i replaced it with:

`if Product.objects.filter(parent__includes=self).exists()` which is much faster by using indexed field parent_id - but there was a failure in codecov/project, see above, please check.

cheers!

gwaidacher:
I had another test with a category in the range containing > 1000000 products and added a child to the last product (oldest date_updated):
The original query took over 11 seconds:
Time: 11346.651 ms (00:11.347)

I also added a check for included_products, so the optimized query for the range with category, without classes and included_products could be reduced to 7 seconds:
Time: 7012.940 ms (00:07.013)

All three checks for classes, included_categories and included_products are insignifcant in time.

gwaidacher:
Hi!

I added another check if included_categories have children and performed some tests measuring 
response times with two ranges including > 1000000 products, please have a look at my comments in the PR

https://github.com/django-oscar/django-oscar/pull/4066


thank you,

-- 
Gunther Waidacher
Acat GmbH
Linke Wienzeile 106/12, A-1060 Wien · Austria


Am 09.03.23 um 09:54 schrieb Voxin Muyli:

specialunderwear:
> Thank you, specialunderwear, for your detailed answer.
> 
> > I think it would be much easier to just offer the browsable method on range and alter the offers.
> 
> The browsable() method in RangeDetailView was already merged from my PR into master a week ago, see #4053 "Exclude non-public Products from Offer - [75ed882](https://github.com/django-oscar/django-oscar/commit/75ed8827bd9a1c2a7613e41f7f08974e3c747530)

Ok I think we need to roll that back, because browsable does not only filter out non public products but also children @viggo-devries @gwaidacher I think you need to put ``browsable`` in the view so it will not display children and not in the core logic because offers should still apply to children that are in the basket.

Note that I am proposing to add ``browsable`` as a method to ``Range`` - not ``Offer`` - which would return the queryset without children, as you are proposing. So you have a sure way to get a queryset without any children and only public products.

If you still think your current solution with the check is good enough and even better than adding a ``browsable`` method to get a filtered queryset from ``Range``, we will have to first test that in a couple of our own sites, where there are A LOT of children. Because in that case the check might be bad for performance. Some sites we have each product has about 40 children and there are about 300k parents.

specialunderwear:
One thing that would also be preferable, can you do the check in the database instead of in python, maybe it's possible to not add the extra query that way?

gwaidacher:
> If you still think your current solution with the check is good enough and even better than adding a `browsable` method to get a filtered queryset from `Range`, we will have to first test that in a couple of our own sites, where there are A LOT of children. Because in that case the check might be bad for performance. Some sites we have each product has about 40 children and there are about 300k parents.


My first efforts by excluding the children earlier in the range queryset did not work, a test checking children in the range failed:
https://github.com/gwaidacher/django-oscar/blob/patch-2/tests/integration/offer/test_range.py#L53

So the browsable() method had to be added in the view:
https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/offer/views.py#L71

My checks for children use exists() - so they return True with the first product with parent_id matching range products found, this is independent of how many children a product has.

I had the original query which took 11 sec run in an analyze tool, which said just collecing merge joins over classes, included_categories and included_products takes over 5 seconds - knowing that we have no classes and categories lead me to the solution to reduce the joins.

I had my tests run in the shell, just fetching the range as 'self' - so I could paste the different queries into the shell, print the query and paste it into database shell to obtain execution time
eg:
self = Range.objects.latest('id')
qs = Product.objects.filter(Q(includes=self) | Q(parent__includes=self) ...)
print(qs.query)

in this way it's also easy to check my switches do not consume significant execution time:
Product.objects.filter(parent__includes=self).exists()
Product.objects.filter(parent__categories__in=expanded_range_categories).exists()

src/oscar/apps/offer/abstract_models.py
```diff
@@ -963,22 +963,17 @@ def all_products(self):
...
-                self.included_categories.values("id")
-            )
-            selected_products = Product.objects.filter(
+        expanded_range_categories = ExpandDownwardsCategoryQueryset(
+            self.included_categories.values("id")
+        )
+        if self.classes.exists():
+            return Product.objects.filter(
                 Q(categories__in=expanded_range_categories)
                 | Q(product_class__classes=self)
```
specialunderwear, line 978: If classes exist, you are skipping the optimization.

src/oscar/apps/offer/abstract_models.py
```diff
@@ -988,15 +983,76 @@ def product_queryset(self):
...
+        if self.classes.exists():
+            return Product.objects.filter(
                 Q(product_class__classes=self)
                 | Q(includes=self)
                 | Q(parent__product_class__classes=self)
                 | Q(parent__includes=self),
                 ~Q(excludes=self),
                 ~Q(parent__excludes=self)
             )
+        # check if included_products have children
```
specialunderwear, line 1027: If classes exist, you are skipping the optimization

specialunderwear (commented):
I think you can simplify your code a lot if you would not duplicate the entire query in you if statements, but instead build up the filter clause based on the if statements.
something like done here as well: https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/catalogue/managers.py#L39

The code would look something like this then:
```
_filter =  models.Q(includes=self) # start with filter clause that always applies
_exclude_filter = models.Q(excludes=self) # also build and exclude
 if self.classes.exists():
    _filter |= Q(product_class__classes=self)
if Product.objects.filter(parent__includes=self).exists():
    _filter |= Q(parent__includes=self)
    _exclude_filter != Q(parent__excludes=self)

# etc ...

return Product.objects.filter(_filter).exclude(_exclude_filter)
```
Something like that. You can make the code a LOT shorter if you do it like that and perhaps keep it in the single method?

We are reviewing pull requests every week from now on.

specialunderwear (changes requested):
Please see the above comments on how to simplify these changes.

gwaidacher:
> Please see the above comments on how to simplify these changes.

Hi! I had a problem with editing the file in my patch-3, so I created a new pull request #4100 for this issue.

gwaidacher:
Simplified code in #4100, closing this PR

PR #4067, opened 2023-03-03
Removing unnecessary permissions from static and documentation f…

dd:
I found that one file in the documentation and one picture have permissions to execute. It seems to be absolutely unnecessary permissions for these files, so I removed them.


specialunderwear (approved):
thanks!

PR #4068, opened 2023-03-03
Fixed import of payment and order exclusions in dashboards

dd:
In a dashboard, unlike a checkout, PaymentError and order errors are imported [directly from packages](https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/dashboard/orders/views.py#L15). I replaced the import in the dashboard with a _get_class_ call similar to [checkout](https://github.com/django-oscar/django-oscar/blob/master/src/oscar/apps/checkout/views.py#L21)

### DIFF
--- a/src/oscar/apps/dashboard/orders/views.py
+++ b/src/oscar/apps/dashboard/orders/views.py
@@ -13,7 +13,6 @@ from django.utils.translation import gettext_lazy as _
 from django.views.generic import DetailView, FormView, ListView, UpdateView
 
 from oscar.apps.order import exceptions as order_exceptions
-from oscar.apps.payment.exceptions import PaymentError
 from oscar.core.compat import UnicodeCSVWriter
 from oscar.core.loading import get_class, get_model
 from oscar.core.utils import datetime_combine, format_datetime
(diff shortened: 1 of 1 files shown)


joeyjurjens (approved):

viggo-devries:
@dd Please write unit test so the code coverage doesn't drop. Thanks!

src/oscar/apps/dashboard/orders/views.py
```diff
@@ -36,6 +34,11 @@
 ShippingAddressForm = get_class(
     'dashboard.orders.forms', 'ShippingAddressForm')
 OrderStatusForm = get_class('dashboard.orders.forms', 'OrderStatusForm')
+InvalidOrderStatus = get_class('order.exceptions', 'InvalidOrderStatus')
+InvalidPaymentEvent = get_class('order.exceptions', 'InvalidPaymentEvent')
+InvalidShippingEvent = get_class('order.exceptions', 'InvalidShippingEvent')
+InvalidStatus = get_class('order.exceptions', 'InvalidStatus')
```
specialunderwear, line 40: Please use get_classes to import multiple classes from the same file

specialunderwear (changes requested):

specialunderwear:
I don't think anyone will ever need to change the exceptions by subclassing.
The get_class mechanism is only needed for this scenario. If you can demonstrate a case where this is needed, feel free to reopen the issue, thank you.

PR #4071, opened 2023-03-13
fix use get_is_discountable instead is_discountable attribute from product

kaewjai:
When there are product discountable true and variant discountable false in cart in the same time, discount is calculated for variant which is incorrect.

### DIFF
--- a/src/oscar/apps/offer/abstract_models.py
+++ b/src/oscar/apps/offer/abstract_models.py
@@ -655,7 +655,7 @@ class AbstractBenefit(BaseOfferMixin, models.Model):
         """
         Determines whether the benefit can be applied to a given basket line
         """
-        return line.stockrecord and line.product.is_discountable
+        return line.stockrecord and line.product.get_is_discountable()
 
     def get_applicable_lines(self, offer, basket, range=None):
         """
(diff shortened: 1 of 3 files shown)


src/oscar/apps/offer/abstract_models.py
```diff
@@ -655,7 +655,7 @@ def can_apply_benefit(self, line):
         """
         Determines whether the benefit can be applied to a given basket line
         """
-        return line.stockrecord and line.product.is_discountable
+        return line.stockrecord and line.product.get_is_discountable()
```
specialunderwear, line 658: Could you please add a test demonstrating the issue you are describing?

specialunderwear (changes requested):

src/oscar/apps/offer/abstract_models.py
kaewjai: @specialunderwear 

image
image

specialunderwear:
Hi there is a failing test:
```
django-oscar/tests/integration/offer/test_absolute_benefit.py:409

ValueError: Strategy hasn't found a price for product Dùｍϻϒ title
```
I think you need to update the fixtures to fix this. If you fix this we will review again. We are going to review pull requests weekly from now!

specialunderwear:
Failed to push changes to this branch so reopened in #4102 4102

Issue #4055, opened 2023-02-23
Configuring ranges does not allow excluding products through the UI


joeyjurjens:
The range model allows exclusion of specific products:
https://github.com/django-oscar/django-oscar/blob/151dc05395865ec0f51cdd45be9970a1bbea2f0c/src/oscar/apps/offer/abstract_models.py#L854

However, you can not configure those through the dashboard UI, while it's a pretty useful feature.

You can configure single included products, with the "Save and edit products" button;
<img width="349" alt="image" src="https://user-images.githubusercontent.com/7702200/220860699-86ca45f9-8e73-4648-b978-8a4b242c7708.png">

We could also allow users to exclude products in this view, though we would probably need to make the view UI a bit clearer here, as to me it already looks a bit clumsy.

A separate view would do as well, but it should inherit the logic from the edit products view, as the only difference should be that it saves the products in the `excluded_products` field rather than the `included_products` field

gwaidacher:
Hi!

I added a table with the excluded_products, please check. If my changes are ok for you, I will try to add the missing tests.

gwaidacher:
Hi, I have a working solution with tabs, with a second textfield for upcs and a separate file upload field
https://github.com/django-oscar/django-oscar/pull/4073

unfortunately the tests failed because the migrations (add 2nd fileuploadfield for excluded products) are not performed automatically - how can this be accomplished?

2nd question: how are translations for the new strings added?

gwaidacher:
![image](https://user-images.githubusercontent.com/117359695/225294963-588ff622-58eb-4bd4-9435-648443f2a223.png)

gwaidacher:
> However, you can not configure those through the dashboard UI, while it's a pretty useful feature.

Hi! Please have a a look at my PR, https://github.com/django-oscar/django-oscar/pull/4073 if it fits your requirements, I'll add the missing tests for the new functionality.

gwaidacher:
added missing tests in #4073

PR #4072, opened 2023-03-13
Added Table for excluded_products in ranges-dashboard

gwaidacher:
Added Table with range.excluded_products including functionality to re-add excluded_product(s) to range again.

see issue https://github.com/django-oscar/django-oscar/issues/4055

### DIFF
--- a/src/oscar/apps/dashboard/ranges/views.py
+++ b/src/oscar/apps/dashboard/ranges/views.py
@@ -129,7 +129,6 @@ class RangeProductListView(BulkEditMixin, ListView):
     def add_selected_products(self, request):
         product_ids = request.POST.getlist('selected_product', None)
         products = self.model.objects.filter(id__in=product_ids)
-        log.info(f'add_selected_products: {products}')
         range = self.get_range()
         for product in products:
             range.add_product(product)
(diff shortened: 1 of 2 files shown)


joeyjurjens:
Hi @gwaidacher, thanks for this PR.

Though, I think you might have misunderstood the issue. The issue is more that you can't add products to the excluded products field. So there should be a way to add products to your excluded products, rather than removing them from your excluded products.

What you've added now, is also useful. So maybe we should create a view where you can just manage the excluding products entirely, as in you can add new products but also remove existing ones.

gwaidacher:
> Hi @gwaidacher, thanks for this PR.
> 
> Though, I think you might have misunderstood the issue. The issue is more that you can't add products to the excluded products field. So there should be a way to add products to your excluded products, rather than removing them from your excluded products.
> 
> What you've added now, is also useful. So maybe we should create a view where you can just manage the excluding products entirely, as in you can add new products but also remove existing ones.

Hi!

I realized that with my PR, one cannot add excluded_products, if all products are included in range. I will close my PR due it's useless in that case. I'd rather solve it with 2 tabs, one for the existing view, another for managing the excluded_products with form and upload_field analog the included_products page

PR #4096, opened 2023-05-05
If a product has a option it cannot be oversold

joeyjurjens:
Replaces #4015 with a change in the reason message.


specialunderwear (changes requested):
plx add tests

Issue #4058, opened 2023-02-24
Upgrade to Django 4.0


viggo-devries:

Guthers:
Potential duplicate of https://github.com/django-oscar/django-oscar/issues/3908?

Issue #3908, opened 2022-04-06
Compatible Django/Python versions are outdated


jsma:
Oscar maintains compatibility with versions of Python and Django that have already reached EOL but does not support the most recent versions. Apologies in advance if I've missed an issue or other PR where this is already under discussion or active development.

### Issue Summary

According to the release notes for Oscar 3.1:

https://github.com/django-oscar/django-oscar/blob/fe37a51c040303b6f0251d1973d39694c3eaba88/docs/source/releases/v3.1.rst?plain=1#L14-L19

According to the release notes for the still in-development Oscar 3.2:

https://github.com/django-oscar/django-oscar/blob/fe37a51c040303b6f0251d1973d39694c3eaba88/docs/source/releases/v3.2.rst?plain=1#L14-L19

I see that there is an [in progress PR](https://github.com/django-oscar/django-oscar/pull/3864) to update the release notes but none of the proposed changes modifies the Django/Python support section.

The currently supported upstream versions as of April 2022 are:
- Python 3.7 to 3.10
- Django 3.2 to 4.0

Details on EOL versions and latest releases:
- Extended support for Django 2.2 ends this month, April 2022
- Extended support for Django 3.1 ended on December 7, 2021
- Django 4.0 was released on December 7, 2021
- Python 3.6 was EOL as of December 2021
- Python 3.10 was released on October 4, 2021 and will be included in the upcoming Ubuntu 22.04 LTS release later this month.

Oscar 3.1 pins Django at:

https://github.com/django-oscar/django-oscar/blob/04dd391c900537a61dbb0ae5250ca5c2df6bbc4b/setup.py#L21

The in-progress Oscar 3.2 pins Django at:

https://github.com/django-oscar/django-oscar/blob/fe37a51c040303b6f0251d1973d39694c3eaba88/setup.py#L21

At the moment, it's not possible to upgrade a project to Django 4.0.x given Oscar's constraints.
I do not know yet if there are any actual incompatibilities with Python 3.10.x.

Are the maintainers of this project open to pull requests to:

1. Remove references to EOL versions in Oscar 3.2
2. Add Django 4.0/Python 3.10 compatibility to Oscar 3.2
 
Is there a target date of when Oscar 3.2 will be released? If not or if it is still many months in the future, would the maintainers be open to a patch release of Oscar 3.1 that adds support for these versions and removes compatibility with EOL versions, which could then be merged into the in-development version?

Thanks!

solarissmoke:
[...]

PR #4098, opened 2023-05-05
Django 4.2, Python 3.11

viggo-devries:
For : #3908 and #4058

### DIFF
--- a/src/oscar/apps/catalogue/abstract_models.py
+++ b/src/oscar/apps/catalogue/abstract_models.py
@@ -220,8 +220,8 @@ class AbstractCategory(MP_Node):
             is_public=False, path__rstartswith=OuterRef("path"), depth__lt=OuterRef("depth")
         )
         self.get_descendants_and_self().update(
-            ancestors_are_public=Exists(
-                included_in_non_public_subtree.values("id"), negated=True))
+            ancestors_are_public=~Exists(
+                included_in_non_public_subtree.values("id")))
 
         # Correctly populate ancestors_are_public
         self.refresh_from_db()
@@ -599,7 +599,9 @@ class AbstractProduct(models.Model):
         """
         Test if this product has any stockrecords
         """
-        return self.stockrecords.exists()
+        if self.id:
+            return self.stockrecords.exists()
+        return False
 
     @property
     def num_stockrecords(self):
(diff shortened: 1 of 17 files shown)


src/oscar/core/filestorage.py
```diff
@@ -0,0 +1,46 @@
...
+class NoSourceMapsStorage(ManifestStaticFilesStorage):
+    """
+    ManifestStaticFilesStorage with additional features:
+    1. If you have a manifest files error, the page to render the error probably
+    has the same error, and then you can't debug it. This returns the error
+    string so one can see in the rendered html there was a problem
+    2. Remove the sourcemap patterns, add javascript module support.
+    """
+
+    # ------ Temp until: https://code.djangoproject.com/ticket/33353#comment:11
```
joeyjurjens, line 30: I don't think this is the right solution for the problem, especially not as a default setting in (`oscar/defaults.py`) because most people will use a different storage backend such has WhiteNoise.

I see two fixes (as the django ticket also points out):
1. Add the .map files to the static folder (bootstrap has them)
2. Remove the `//# sourceMappingURL=bootstrap.bundle.min.js.map` (at the bottom of the file) from the JS file, the collectstatic command will just work fine then.

It doesn't seem like they're gonna change this behaviour.

src/oscar/apps/catalogue/abstract_models.py
```diff
@@ -215,12 +215,17 @@ def set_ancestors_are_public(self):
...
+            self.get_descendants_and_self().update(
+                ancestors_are_public=self.is_public
+            )
+        else:
+            included_in_public_subtree = self.__class__.objects.filter(
+                is_public=True, path__rstartswith=OuterRef("path"), depth__lt=OuterRef("depth")
+            )
+            self.get_descendants_and_self().update(
+                ancestors_are_public=Exists(
+                    included_in_public_subtree.values("id")))
```
specialunderwear, line 228: I don't think this is the correct approach.
In the django release notes we can find:

The undocumented negated parameter of the [Exists](https://docs.djangoproject.com/en/4.2/ref/models/expressions/#django.db.models.Exists) expression is removed.

Maybe you can try ~Exists() ?

specialunderwear (changes requested):

specialunderwear (approved):

specialunderwear (approved):

specialunderwear (approved):

joeyjurjens (approved):

Issue #4115, opened 2023-06-01
`make sandbox` fails, trying to build psycopg2-binary on M1 Mac


TibsAtWork:
### Issue Summary

This is using an M1 Mac with Ventura (13.3.1) and Python 3.9.16

If I run `make sandbox` it fails because it cannot build `psycopg2-binary` from source on my machine.

Summary: `requirements.txt` allows psycopg2 below version 2.10, but `setup.py` (test specification) only allows below 2.9. The current version of psycopg2-binary is 2.9.6. The attempt to satisfy `setup.py` tries to build an earlier version of psycopg2 from source, which doesn't work out-of-the-box.

### Steps to Reproduce

On an M1 Mac without PostgreSQL or any of the support libraries installed, do `make sandbox`

The `install-python` phase of this will successfully install psycopg2-binary, choosing v2.9.6 - this is following the rule `psycopg2-binary>=2.8,<2.10` in `requirements.txt`.

The `install-test` phase then does `pip install -e .[test]`, which tries to install the `test_requires` elements from `setup.py`.
Line 63 of `setup.py` specifies `'psycopg2-binary>=2.8,<2.9',`, so it tries to install an earlier version, 2.8.6. That's old enough that there isn't a binary wheel for Mac M1, so it tries to build it from source.

As I don't have the packages installed to allow that to happen, it fails.

#### Proposed fix

Updating line 63 of `setup.py` to have the same constraint as `requirements.txt` - that is
```
     'psycopg2-binary>=2.8,<2.10',
```
allows `make sandbox` to work.

Note: after that fix, `make test` still works.

### Technical details

* Python version: Run `python --version`.

  `Python 3.9.16`

* Django version: Look in your requirements.txt, or run `pip show django | grep Version`.

  `Version: 3.2.19`

* Oscar version: Look in your requirements.txt, or run `pip show django-oscar | grep Version`.

  `Version: 3.2`

TibsAtWork:
Withdrawing this, as it's expected to be fixed when https://github.com/django-oscar/django-oscar/pull/4098 is merged (as indicated on the linked PR)

PR #4116, opened 2023-06-01
Update "test support" version of psycopg2-binary to match `requirements.txt`

TibsAtWork:
`requirements.txt` allows `psycopg2-binary>=2.8,<2.10`, which chooses the current version, 2.9.6

`setup.py` in the `test_requires` clause, specifies `psycopg2-binary>=2.8,<2.9` (see line 63). This does not allow 2.9.6.

As a result, `make sandbox` (specifically, `make install-test`) tries to install an earlier version of psycopg2-binary. It chooses 2.8.6, but that's old enough that there isn't a binary wheel for Mac M1, so it tries to build it from source. Since I don't have the environment set up to support that build, it fails.

Updating `setup.py` to the same version specification as `requirements.txt`, as in this PR, allows `make install-test` (and thus `make sandbox`) to succeed.

Note: `make test` also succeeds.

Fixes https://github.com/django-oscar/django-oscar/issues/4115


viggo-devries:
Hi, This will be fixed when #4098 is merged, should be soon.

Issue #4117, opened 2023-06-10
0010_convert_to_valid_json crashes


jribbens:
Found a bug? Please fill out the sections below.

### Issue Summary

When I run `./manage.py migrate` it fails with:
```
  File "env/lib/python3.9/site-packages/oscar/apps/basket/migrations/0010_convert_to_valid_json.py", line 19, in forward
    val = literal_eval(at.value)
  File "/usr/local/lib/python3.9/ast.py", line 62, in literal_eval
    node_or_string = parse(node_or_string, mode='eval')
  File "/usr/local/lib/python3.9/ast.py", line 50, in parse
    return compile(source, filename, mode, flags,
  File "<unknown>", line 1
    https://www.example.com/
         ^
```

The same thing happens with `order/migrations/0012_convert_to_valid_json.py` .

### Steps to Fix

In the code which tries to parse the value as python, change the `except ValueError:` to `except (ValueError, SyntaxError):`

### Technical details

* Python version: 3.9
* Django version: 3.2.19
* Oscar version: 3.2

PR #4118, opened 2023-06-10
Fix convert_to_valid_json migrations

jribbens:
Closes django-oscar/django-oscar#4117


specialunderwear:
Thank you for fixing our oversight

PR #4138, opened 2023-07-19
Register discounts per line

viggo-devries:


src/oscar/apps/basket/abstract_models.py
```diff
@@ -924,35 +950,46 @@ def line_price_excl_tax(self):
...
+
+        if excl_tax_discounts and self.line_price_excl_tax is not None:
+            return max(0, self.line_price_excl_tax - excl_tax_discounts)
+
+        if incl_tax_discounts and self.line_price_incl_tax is not None:
             return max(
                 0,
                 self.line_price_excl_tax
-                - round_half_up(self._tax_ratio * self._discount_incl_tax),
+                - round_half_up(self._tax_ratio * incl_tax_discounts),
```
specialunderwear, line 966: This calculation is incorrect if all other prices are calculated from prices ex tax.
This must be checked.

src/oscar/apps/basket/abstract_models.py
```diff
@@ -924,35 +950,46 @@ def line_price_excl_tax(self):
...
-        elif self.line_price_excl_tax is not None and self._discount_excl_tax:
-            return max(
-                0,
-                round_half_up(
-                    (self.line_price_excl_tax - self._discount_excl_tax)
-                    / self._tax_ratio
-                ),
+        if self.line_price_incl_tax is not None and incl_tax_discounts:
+            return max(0, self.line_price_incl_tax - incl_tax_discounts)
+        elif self.line_price_excl_tax is not None and excl_tax_discounts:
```
specialunderwear, line 981: This is unfinshed, it must be checked if the rounding is done at the correct point.

specialunderwear (changes requested):

src/oscar/apps/basket/abstract_models.py
specialunderwear: Checked and concluded rounding is not done at the correct point. To fix this issue, oscar should use 4 decimals for all prices internally. Added hint for developers to implement _tax_ratio to get rid of rounding errors.

specialunderwear (dismissed):

src/oscar/apps/basket/abstract_models.py
```diff
@@ -928,34 +954,55 @@ def line_price_excl_tax(self):
...
-                0,
-                self.line_price_excl_tax
-                - round_half_up(self._tax_ratio * self._discount_incl_tax),
-            )
+        if self.line_price_excl_tax is None:
+            return None
+
+        excl_tax_discounts = self.discounts.excl_tax
+        if excl_tax_discounts:
+            # there are discounts that return a value excluding tax, we can simply
```
samar-hassan, line 962:
```suggestion
            # these are discounts that return a value excluding tax, we can simply
```

samar-hassan (changes requested):

samar-hassan (approved):

PR #4139, opened 2023-07-30
Fix error when there is no "basket" defined

TopDevPros:
This is a simple change that removes an error if someone starts a django-oscar project and goes to the front page without adding any data.


TopDevPros:
Duplicate

PR #4140, opened 2023-07-30
Fix error when there is no "basket" defined

TopDevPros:
When you start a new django-oscar project without any data and go to the front page, you see an error. This change resolves that issue.

### DIFF
--- a/src/oscar/templates/oscar/partials/mini_basket.html
+++ b/src/oscar/templates/oscar/partials/mini_basket.html
@@ -3,10 +3,12 @@
 
 <div class="basket-mini col-sm-5 text-right d-none d-md-block">
     <strong>{% trans "Basket total:" %}</strong>
-    {% if request.basket.is_tax_known %}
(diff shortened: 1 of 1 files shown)


specialunderwear (changes requested):
Hi I think it would be better to change the is_tax_known method to not generate an error for empty backets. Could you change this to work like that?

TopDevPros:
Great suggestion, @specialunderware!

Added fix to is_tax_known() in oscar.apps.basket.abstract_models.AbstractBasket and a test to verify it works with an empty basket. 

Also, modified the 2 tests in tests.integration.order.TestShippingOfferForOrder which relied on is_tax_known() incorrectly returning True. Fixed. Abstracted some duplicated code into TestShippingOfferForOrder.setUp().

All tests pass locally.

Pushed changes to topdevpros/django-oscar/master branch because the original changes to the templates are no longer needed.

Let us know if you need any other changes or info.

TopDevPros:
Replaced by PR 4166

PR #4150, opened 2023-08-23
Speedup Range.product_queryset() even more

gwaidacher:
Do not join over parent__categories if they are None

Remember, we have a huge Database with > 3,000.000 products, see our demo shop: https://shop.acat.cc/

I have a Range with all level - 2 Cateories except 1 for an offer reserved for special users, giving a discount of 10 percent for all included products.

Due all catalogue pages calculate the basket total, we had again a total execution time of ~ 12 seconds.

The check for `parent__categories__in=expanded_range_categories` took 7 seconds and returned 'False', including all parent__categories with value = null in the check. By excluding `parent__categories=None` we have now an acceptable execution time again.

Speedup: from 12 Seconds down to 513 ms

### DIFF
--- a/src/oscar/apps/offer/abstract_models.py
+++ b/src/oscar/apps/offer/abstract_models.py
@@ -1124,11 +1124,9 @@ class AbstractRange(models.Model):
             _filter |= Q(categories__in=expanded_range_categories)
             # extend filter for parent categories, exclude parent = None
             if (
-                Product.objects.exclude(
-                    parent=None
-                ).filter(
-                    parent__categories__in=expanded_range_categories
-                ).exists()
+                Product.objects.exclude(parent=None)
+                .filter(parent__categories__in=expanded_range_categories)
+                .exists()
             ):
                 _filter |= Q(parent__categories__in=expanded_range_categories)
 
--- a/src/oscar/apps/offer/abstract_models.py
+++ b/src/oscar/apps/offer/abstract_models.py
@@ -1119,8 +1119,8 @@ class AbstractRange(models.Model):
                 self.included_categories.values("id")
             )
             _filter |= Q(categories__in=expanded_range_categories)
-            # extend filter for parent categories
-            if Product.objects.filter(
+            # extend filter for parent categories, exclude nulls
+            if Product.objects.exclude(parent__categories=None).filter(
                 parent__categories__in=expanded_range_categories
             ).exists():
                 _filter |= Q(parent__categories__in=expanded_range_categories)
--- a/src/oscar/apps/offer/abstract_models.py
+++ b/src/oscar/apps/offer/abstract_models.py
@@ -1122,11 +1122,13 @@ class AbstractRange(models.Model):
                 self.included_categories.values("id")
             )
             _filter |= Q(categories__in=expanded_range_categories)
-            # extend filter for parent categories, exclude nulls
+            # extend filter for parent categories, exclude parent = None
             if (
-                Product.objects.exclude(parent__categories=None)
-                .filter(parent__categories__in=expanded_range_categories)
-                .exists()
+                Product.objects.exclude(
+                    parent=None
+                ).filter(
+                    parent__categories__in=expanded_range_categories
+                ).exists()
             ):
                 _filter |= Q(parent__categories__in=expanded_range_categories)
 
--- a/src/oscar/apps/offer/abstract_models.py
+++ b/src/oscar/apps/offer/abstract_models.py
@@ -427,6 +427,36 @@ class AbstractConditionalOffer(models.Model):
         descriptions = [r["description"] for r in restrictions]
         return "<br/>".join(descriptions)
 
+    def hide_time_if_zero(self, dt):
+        # Only show hours/minutes if they have been specified
+        if dt.tzinfo:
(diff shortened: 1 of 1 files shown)


gwaidacher:
Here the executed query before my changes:

SELECT (1) AS "a" FROM "catalogue_product" INNER JOIN "catalogue_product" T2 ON ("catalogue_product"."parent_id" = T2."id") INNER JOIN "catalogue_productcategory" ON (T2."id" = "catalogue_productcategory"."product_id") WHERE "catalogue_productcategory"."category_id" IN (SELECT "CATALOGUE_CATEGORY_JOIN"."id" FROM "catalogue_category" AS "CATALOGUE_CATEGORY_BASE" LEFT JOIN "catalogue_category" AS "CATALOGUE_CATEGORY_JOIN" ON ( "CATALOGUE_CATEGORY_JOIN"."path" LIKE "CATALOGUE_CATEGORY_BASE"."path" || '%' AND "CATALOGUE_CATEGORY_BASE"."depth" <= "CATALOGUE_CATEGORY_JOIN"."depth" ) WHERE "CATALOGUE_CATEGORY_BASE"."id" IN (SELECT U0."id" FROM "catalogue_category" U0 INNER JOIN "offer_range_included_categories" U1 ON (U0."id" = U1."category_id") WHERE U1."range_id" = 4)) LIMIT 1;
 a 
═══
(0 rows)

explain analyze:
 Planning Time: 3.045 ms
 Execution Time: 6996.660 ms

gwaidacher:
execution time for the modiefied check is now 5ms (!)

`Product.objects.exclude(parent__categories=None)
 .filter(parent__categories__in=expanded_range_categories)
 .exists()`

specialunderwear:
I'm not sure how the parent__categories are None?
Isn;t the category mandatory?

gwaidacher:
> I'm not sure how the parent__categories are None? Isn;t the category mandatory?

Hi! It's the parent's categories, I suppose. If the product has no parent, parent__categories are None

specialunderwear:
> > I'm not sure how the parent__categories are None? Isn;t the category mandatory?
> 
> Hi! It's the parent's categories, I suppose. If the product has no parent, parent__categories are None

Would it make more sence to exclude parent__isnull then?

gwaidacher:
> > > I'm not sure how the parent__categories are None? Isn;t the category mandatory?
> > 
> > 
> > Hi! It's the parent's categories, I suppose. If the product has no parent, parent__categories are None
> 
> Would it make more sence to exclude parent__isnull then?

it does not affect execution time, but I modified the exclude constraint.

gwaidacher:
The modified check takes below 5 ms execution time, so the query is over 1500 times faster (!) than before:

SELECT (1) AS "a" FROM "catalogue_product" INNER JOIN "catalogue_product" T2 ON ("catalogue_product"."parent_id" = T2."id") INNER JOIN "catalogue_productcategory" ON (T2."id" = "catalogue_productcategory"."product_id") WHERE (NOT ((EXISTS(SELECT (1) AS "a" FROM "catalogue_product" U1 LEFT OUTER JOIN "catalogue_productcategory" U2 ON (U1."id" = U2."product_id") WHERE (U2."category_id" IS NULL AND U1."id" = "catalogue_product"."parent_id") LIMIT 1) OR "catalogue_product"."parent_id" IS NULL)) AND "catalogue_productcategory"."category_id" IN (SELECT "CATALOGUE_CATEGORY_JOIN"."id" FROM "catalogue_category" AS "CATALOGUE_CATEGORY_BASE" LEFT JOIN "catalogue_category" AS "CATALOGUE_CATEGORY_JOIN" ON ( "CATALOGUE_CATEGORY_JOIN"."path" LIKE "CATALOGUE_CATEGORY_BASE"."path" || '%' AND "CATALOGUE_CATEGORY_BASE"."depth" <= "CATALOGUE_CATEGORY_JOIN"."depth" ) WHERE "CATALOGUE_CATEGORY_BASE"."id" IN (SELECT U0."id" FROM "catalogue_category" U0 INNER JOIN "offer_range_included_categories" U1 ON (U0."id" = U1."category_id") WHERE U1."range_id" = 4)) ) LIMIT 1;
 a 
═══
(0 rows)

Time: 4.489 ms

specialunderwear (changes requested):
Hi, are the changes to hide_time_if_zero by accident or are they supposed to be included in this PR? If so can you explain what this is for?

gwaidacher:
> Hi, are the changes to hide_time_if_zero by accident or are they supposed to be included in this PR? If so can you explain what this is for?

Hi! There was a `too complex` warning @ `hide_time_if_zero()`, so I moved lines 473 - 490 into a separate function ` datetime_range_availability_restriction()`

specialunderwear:
> > Hi, are the changes to hide_time_if_zero by accident or are they supposed to be included in this PR? If so can you explain what this is for?
> 
> Hi! There was a `too complex` warning @ `hide_time_if_zero()`, so I moved lines 473 - 490 into a separate function ` datetime_range_availability_restriction()`
> 
> Functionality of hide_time_if_zero() did not change.

Hi could you undo the changes to hide_time_if_zero in this PR, that makes it more clear to see what changed. Sorry for leaving this open so long.

PR #4166, opened 2023-09-11
Return True in is_tax_known() when the basket is empty

TopDevPros:
Added fix to is_tax_known() in oscar.apps.basket.abstract_models.AbstractBasket and a test to verify it works with an empty basket.

Also, modified the 2 tests in tests.integration.order.TestShippingOfferForOrder which relied on is_tax_known() incorrectly returning True when the basket was empty. 

Replaces PR 4140


viggo-devries (approved):
Looks good to me,

PR #4168, opened 2023-09-15
[FEAT] Absolute discount benefit applied on each product

samar-hassan:
I have added an absolute discount benefit that is added to each of the applicable lines. The value in the benefit defines the maximum discount amount regardless of the product quantity. If price * quantity is less than the value than the product is free.


src/oscar/apps/offer/benefits.py
```diff
@@ -203,6 +204,50 @@ def apply(
...
+
+    def apply(self, basket, condition, offer, **kwargs):
+        # Fetch basket lines that are in the range and available to be used in an offer.
+        line_tuples = self.get_applicable_lines(offer, basket)
+        lines_to_discount = self.get_lines_to_discount(offer, line_tuples)
+
+        applied_discount = D("0.00")
+        for (line, price, qty) in lines_to_discount:
+            # If price is less than the fixed discount, then it will be free.
+            line_discount = min(price * qty, self.value)
```
specialunderwear, line 252: Shouldn't self.value also be multiplied with qty?

src/oscar/apps/offer/benefits.py
```diff
@@ -203,6 +204,50 @@ def apply(
...
+    regardless of the quantity.
+    """
+
+    class Meta:
+        app_label = "offer"
+        proxy = True
+        verbose_name = _("Absolute product discount benefit")
+        verbose_name_plural = _("Absolute product discount benefits")
+
+    def get_lines_to_discount(self, offer, line_tuples):
```
specialunderwear, line 220: Is there no default implementation of ``get_lines_to_discount`` on Benefit that works appropriately? Is this implementation really different from other discounts?

tests/integration/offer/test_absolute_product_benefit.py
```diff
@@ -0,0 +1,368 @@
...
+    def test_applies_correctly_to_empty_basket(self):
+        result = self.benefit.apply(self.basket, self.condition, self.offer)
+        self.assertEqual(D("0.00"), result.discount)
+        self.assertEqual(0, self.basket.num_items_with_discount)
+        self.assertEqual(0, self.basket.num_items_without_discount)
+
+    def test_applies_correctly_to_basket_which_matches_condition_with_one_line(self):
+        add_product(self.basket, price=D("12.00"), quantity=2)
+        result = self.benefit.apply(self.basket, self.condition, self.offer)
+        self.assertEqual(D("3.00"), result.discount)
```
specialunderwear, line 81: I think this should be 6?

specialunderwear (changes requested):

src/oscar/apps/offer/abstract_models.py
```diff
@@ -551,6 +552,7 @@ class AbstractBenefit(BaseOfferMixin, models.Model):
     TYPE_CHOICES = (
         (PERCENTAGE, _("Discount is a percentage off of the product's value")),
         (FIXED, _("Discount is a fixed amount off of the basket's total")),
+        (FIXED_PRODUCT, _("Discount is a fixed amount off of the product's value")),
```
viggo-devries, line 555: Discount is a fixed amount off of every product's value in the entire basket.

viggo-devries (changes requested):

src/oscar/apps/offer/abstract_models.py
specialunderwear: Don't listen to viggo, this is fine.

src/oscar/apps/offer/benefits.py
samar-hassan: If we multiply self.value with qty then user can order unlimited products for free. Using the current configuration the maximum discount a user can avail per product would be self.value regardless of the quantity.

src/oscar/apps/offer/benefits.py
samar-hassan: There is no default implementation of `get_lines_to_discount`. Each benefit uses slightly different version of this code.

src/oscar/apps/offer/abstract_models.py
```diff
@@ -537,9 +537,10 @@ class AbstractBenefit(BaseOfferMixin, models.Model):
     )
 
     # Benefit types
-    PERCENTAGE, FIXED, MULTIBUY, FIXED_PRICE = (
+    PERCENTAGE, FIXED, FIXED_UNIT, MULTIBUY, FIXED_PRICE = (
         "Percentage",
         "Absolute",
+        "Fixed unit",
```
specialunderwear, line 543: Just call this "Fixed"

src/oscar/apps/offer/benefits.py
specialunderwear: I don't understand, this value is the maximum discount for the LINE right, not per product.
But right now, if there are 2 products and the discount value is 1, I will only get discount 1 and not 2, which is 1 for each product.

specialunderwear (changes requested):

tests/integration/offer/test_fixed_unit_benefit.py
```diff
@@ -0,0 +1,368 @@
...
+    def test_applies_correctly_to_empty_basket(self):
+        result = self.benefit.apply(self.basket, self.condition, self.offer)
+        self.assertEqual(D("0.00"), result.discount)
+        self.assertEqual(0, self.basket.num_items_with_discount)
+        self.assertEqual(0, self.basket.num_items_without_discount)
+
+    def test_applies_correctly_to_basket_which_matches_condition_with_one_line(self):
+        add_product(self.basket, price=D("12.00"), quantity=2)
+        result = self.benefit.apply(self.basket, self.condition, self.offer)
+        self.assertEqual(D("3.00"), result.discount)
```
specialunderwear, line 81: I still believe this should be 6, 3 dollar discount per product, because there are 2 products and each gets 3 dollar discount.

specialunderwear (changes requested):

src/oscar/apps/offer/benefits.py
specialunderwear: I also don't think that it is a problem that unlimited ordering products for free is a problem. Each benefot has a max applications property the use can use to limit that, and also if the user does not choose a discount value greater than the value of the product there is no free product. With the percentage discount you can also get unlimited products for free if you put the percentage to 100. Making sure the product isn't free is the responsibility of the user.

specialunderwear (approved):

Issue #4152, opened 2023-08-31
Offer's slug doesn't support unicode


alnuaimi94:
I gotten error when I visited offers page.
I'm use Arabic language.![image](https://github.com/django-oscar/django-oscar/assets/96056894/ab15cff6-536e-4bfb-82fd-7752abf06cef)

alnuaimi94:
I fixed that by edit offer detail path from:

'python
path("<slug:slug>/", self.detail_view.as_view(), name="detail"),
'

To:

'python
path("<str:slug>/", self.detail_view.as_view(), name="detail"),
'

specialunderwear:
You can make this work by configuring the following setting:
```
OSCAR_SLUG_ALLOW_UNICODE = True
```

alnuaimi94:
> You can make this work by configuring the following setting:
> 
> ```
> 
> OSCAR_SLUG_ALLOW_UNICODE = True
> 
> ```
> 
> 
> 
> That fixed here: https://github.com/django-oscar/django-oscar/pull/1995
> 
> 
> 
> Hmm I think your issue is a bit different.

It's work only with product slug

PR #4172, opened 2023-09-18
[FIX] Offer's slug doesn't support unicode

samar-hassan:
Resolves https://github.com/django-oscar/django-oscar/issues/4152


specialunderwear (changes requested):
Can you add a small test for each url that demonstrates unicode slugs are now handled properly?

specialunderwear (approved):

PR #4174, opened 2023-09-22
[Refactor] Move OfferWizardStepView to separate file

samar-hassan:
This PR moves `OfferWizardStepView` to a separate file and dynamically loaded in `dashboard.offers.views`. This removes the burden of changing the same thing in 4 different step classes.


specialunderwear (changes requested):
It seems there is still some untested code left.

specialunderwear (approved):

Issue #4164, opened 2023-09-04
Logging-in with basket products may cause order lines to have incorrect partner / stock record


st8st8:
Found a bug? Please fill out the sections below.

### Issue Summary

When logging in to Oscar with products in your basket, if that causes the Partner Strategy to change then the basket price updates, but the partner and stock record does not.  This is turn is reflected in the resulting Order record.

### Steps to Reproduce

1. Set up a shop where certain users have a different strategy / partner to the anonymous user for a given product
2. Visit the shop without logging in
3. Add the product to your basket
4. Log in to oscar
5. Check out and complete the order

Now the lines in the frozen basket and resultant order will have the correct price, but the wrong partner and stock record.  It would be helpful if this was correct as I'm now auditing past purchases based on the partner and stock record.

### Technical details

Any python/Djange version, Oscar 3.2

viggo-devries:
Hi,

I understand.

Could you please create a test for this on a branch maybe that shows this unwanted behaviour?

And would you be willing to work on this issue yourself? I can point you to the right way of going about fixing it.

st8st8:
Yeah, i can look at this.  Leave it with me for now.

st8st8:
OK, I've added a couple of tests in a branch, along with a potential fix:

https://github.com/st8st8/django-oscar/tree/partner_basket

I'll give this a run out in my shops for a bit and see what happens.

PR #4180, opened 2023-10-03
Update basket stock records when strategy changes

st8st8:


st8st8:
References issue:

https://github.com/django-oscar/django-oscar/issues/4164

st8st8:
Actually - this is quite broken because it causes attempted modifications of frozen baskets, I'll have to rethink...

specialunderwear:
Ok, looking forward to your solution, thnx!

st8st8:
In the end, I went with updating stock records at time of merge baskets, when users log in.

Take a look....

PR #4192, opened 2023-10-30
Add Bootstrap responsive image class

jamesp9:
Found when viewing product images on the mobile. That the image wasn't centered properly and stretched off to the right a little. Setting Bootstrap's class for responsive images seems to do the trick.


specialunderwear (approved):
thank you!

PR #4201, opened 2023-11-04
Fix the collapsed table header

patrikniebur:
When all content is floated (like in stock alerts), the height of the table header collapses. Changing its display value to `flow-root` prevents this from happening. I haven't checked all the uses of `.table-header` but I can't imagine a scenario where this change would break the layout.

### BEFORE
Screenshot 2023-11-04 at 10 31 42

### AFTER
Screenshot 2023-11-04 at 10 31 51

### DIFF
--- a/src/oscar/static_src/oscar/scss/dashboard/tables.scss
+++ b/src/oscar/static_src/oscar/scss/dashboard/tables.scss
@@ -21,6 +21,10 @@
   content: "\f078";
 }
 
+.table-header {
+  display: flow-root;
+}
+
 caption,
 .table-header {
   h2, h3 {


samar-hassan:
Thanks @patrikniebur for highlighting the styling issue. However, the proposed solution appears to affect other pages using the .table-header class. A more targeted approach might be beneficial. Consider applying the adjustment exclusively to the table on the stock alerts page to avoid interference with other tables.

samar-hassan (changes requested):
Thanks @patrikniebur for highlighting the styling issue. However, the proposed solution appears to affect other pages using the .table-header class. A more targeted approach might be beneficial. Consider applying the adjustment exclusively to the table on the stock alerts page to avoid interference with other tables.

patrikniebur:
Hello @samar-hassan. I did consider more targeted approach but it would just reduce the benefits of the change. Currently as it is, it has no effect on tables with at least one non floated child and if all child elements are floated it will prevent from the same issue happening again.

patrikniebur:
@samar-hassan any further thoughts on this?

samar-hassan:
@patrikniebur this affects all dashboard pages with table. The layout is distorted. Please check the order and product listing pages for reference.

patrikniebur:
Thank you @samar-hassan. Sorry I missed these pages the project is still new for me. 

I missed the fact I was changing display property of `caption` element too so I have now separated this style. I've reviewed the pages you suggested and these are now displaying correctly, also all the uses of `table-header` class and made sure this won't cause any other layout problems.

samar-hassan (approved):

PR #4221, opened 2024-01-07
Round prices to two decimal places in price comparison check in basket

fortelll:
Fixed price comparison by rounding to two decimal places, preventing misleading cart notifications when prices are effectively the same. Previously, the comparison between these two prices was done directly, which caused issues when the number of decimal places differed.


aekespong (approved):
Minor change that fixes this bug.

viggo-devries:
@fortelll please use round_half_up like everywhere else in the basket.

Could you please add a test that shows this failed before?

PR #4228, opened 2024-01-20
Update pre-commit URLs

cclauss:
These two URLs no longer point to usable pre-commit repositories so fix them to be valid.


specialunderwear:
Hi Claus, we switched to using black and pylint. So I think the precommit hook should reflect that. Here is an example of a precommit file that uses the correct versions installed in the virtualenv by pip, putting dependency management in 1 place, I think that this is the way you should approach it here as well.

https://github.com/django-oscar/django-oscar-odin/blob/master/.pre-commit-config.yaml

PR #4231, opened 2024-01-22
django.utils.timezone.utc was removed in Django 5.0

cclauss:
https://docs.djangoproject.com/en/stable/releases/5.0/#features-removed-in-5-0
> The django.utils.timezone.utc alias to datetime.timezone.utc is removed.

https://docs.djangoproject.com/en/stable/releases/4.1/#id2
> The django.utils.timezone.utc alias to datetime.timezone.utc is deprecated. Use datetime.timezone.utc directly.

### DIFF
--- a/tests/functional/dashboard/test_review.py
+++ b/tests/functional/dashboard/test_review.py
@@ -1,4 +1,4 @@
-from datetime import timedelta
+from datetime import timedelta, timezone as datetime_timezone
 
 from django.urls import reverse
 from django.utils import timezone
@@ -84,7 +84,7 @@ class ReviewsDashboardTests(WebTestCase):
             ProductReviewSearchForm doesn't recognize the timezone notation.
             """
             return timezone.make_naive(
-                now - timedelta(days=days), timezone=timezone.utc
+                now - timedelta(days=days), timezone=datetime_timezone.utc
             )
 
         now = timezone.now()


tests/functional/dashboard/test_review.py
```diff
@@ -1,4 +1,4 @@
-from datetime import timedelta
+from datetime import timedelta, timezone as datetime_timezone
```
viggo-devries, line 1: please do
```from datetime.timezone import utc```

**Review by viggo-devries — CHANGES_REQUESTED** [review]

File: tests/functional/dashboard/test_review.py (hunk 1 above)
Comment (cclauss) [reply tests/functional/dashboard/test_review.py:1]:
```suggestion
from datetime import timedelta
from datetime.timezone import utc
```

tests/functional/dashboard/test_review.py
```diff
@@ -84,7 +84,7 @@ def n_days_ago(days):
             ProductReviewSearchForm doesn't recognize the timezone notation.
             """
             return timezone.make_naive(
-                now - timedelta(days=days), timezone=timezone.utc
+                now - timedelta(days=days), timezone=datetime_timezone.utc
```
cclauss, line 87:
```suggestion
                now - timedelta(days=days), timezone=utc
```

tests/functional/dashboard/test_review.py
```diff
@@ -1,4 +1,5 @@
 from datetime import timedelta
+from datetime.timezone import utc
```
cclauss, line 2: Does not work.  `datetime` is weird that way.

tests/functional/dashboard/test_review.py
viggo-devries: okay whatever this is in tests anyway so it does not really matter.

viggo-devries (approved):

PR #4213, opened 2023-12-08
DRAFT: GitHub Actions: Test on Python v3.12 and Django v5.0

cclauss:
Blocked by:
* django-haystack/django-haystack#1863
    * https://pypi.org/project/django-haystack -- Need a version later than v3.2.1

Many of the GitHub Actions changes will be fixed automatically by:
* #4230

PR #4236, opened 2024-01-26
Django 4.2, Python 3.12

cclauss:
Like #4098 but with Python 3.12
* #4098 
* https://docs.djangoproject.com/en/5.0/faq/install/#what-python-version-can-i-use-with-django


cclauss:
Closing in favor of #4213.
* #4213

PR #4257, opened 2024-03-08
Dashboard: Add a link to the attribute option group list

### LINKED ISSUE #4256: There is no link from the dashboard to Attribute Option Group list page
I found the page `/dashboard/catalogue/attribute-option-group/` usefull to manage option groups (insert / delete options from groups), but I see no link from the dashboard to this page, is it missing?

JulienPalard:
Closes #4256.

It looks like this:

Screenshot 2024-03-08 at 09-00-31 Dashboard Oscar -

and links to the existing list of attribute option group.

The text was already known by gettext (and it's already translated in french at least).


viggo-devries (approved):
I've tested this and all seems to work well!

PR #4261, opened 2024-03-22
[FEAT] Add excluded categories in range

samar-hassan:
When Includes all products is checked in the range, adding categories in the excluded categories field with remove the related products of those categories from the range.


src/oscar/apps/offer/abstract_models.py
```diff
@@ -1112,8 +1122,23 @@ def product_queryset(self):
...
+            # extend filter if excluded_categories exist
+            if self.excluded_categories.exists():
+                expanded_range_categories = ExpandDownwardsCategoryQueryset(
+                    self.excluded_categories.values("id")
+                )
+                _filter |= Q(categories__in=expanded_range_categories)
+                # extend filter for parent categories, exclude parent = None
+                if (
+                    Product.objects.exclude(parent=None)
+                    .filter(parent__categories__in=expanded_range_categories)
```
viggo-devries, line 1135: why do we check for this? can't we just always add the extra parent query to the exclude?

viggo-devries (changes requested):

specialunderwear (approved):

Issue #4263, opened 2024-03-29
Must shown the Option value instead of Option code in thank you page showed products.


svandeneertwegh:
Found a bug? Please fill out the sections below.

### Issue Summary
When you are on the '/checkout/thank-you/' page, you see a list of products you bought. Normally when you used an option, the product name has '( option='optionvalue') in the string.
Problem is that the option= must be the option.value and not option.code.

For example on the basket page it does show good but only on the summary/thankyou page it does this issue.

### Technical details

* Python version: 3.11.8.
* Django version: 4.2.7.
* Oscar version: 3.2.4.

SarthakNikhal:
Can I work on this issue?

svandeneertwegh:
You may.

PR #4269, opened 2024-04-01
Fix get rules from order with option value instead of option code

svandeneertwegh:
This fixes showing order lines with the option value instead of the code.

#4263

### DIFF
--- a/src/oscar/apps/order/abstract_models.py
+++ b/src/oscar/apps/order/abstract_models.py
@@ -698,10 +698,10 @@ class AbstractLine(models.Model):
             value = attribute.value
             if isinstance(value, list):
                 ops.append(
-                    "%s = '%s'" % (attribute.type, (", ".join([str(v) for v in value])))
+                    "%s = '%s'" % (attribute.option.name, (", ".join([str(v) for v in value])))
                 )
             else:
-                ops.append("%s = '%s'" % (attribute.type, value))
+                ops.append("%s = '%s'" % (attribute.option.name, value))
         if ops:
             desc = "%s (%s)" % (desc, ", ".join(ops))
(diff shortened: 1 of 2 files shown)


src/oscar/apps/order/abstract_models.py
```diff
@@ -698,10 +698,10 @@ def description(self):
             value = attribute.value
             if isinstance(value, list):
                 ops.append(
-                    "%s = '%s'" % (attribute.type, (", ".join([str(v) for v in value])))
+                    "%s = '%s'" % (attribute.option.name, (", ".join([str(v) for v in value])))
```
viggo-devries, line 701: I think this should be attribute.value because the option could be empty

viggo-devries (changes requested):

src/oscar/apps/order/abstract_models.py
svandeneertwegh: When you look at another abstract_models. For example at baskets module, it is als like i changed.
https://github.com/django-oscar/django-oscar/blob/c20ef16274642dbc95d7ef26d9c6316ce26a08dc/src/oscar/apps/order/abstract_models.py#L690

tests/unit/catalogue/test_product_attributes.py
```diff
@@ -55,7 +55,7 @@ def setUp(self):
             upc="1234",
         )
         product.attr.weight = 3
-        product.full_clean()
+        # product.full_clean()
```
viggo-devries, line 58: oeps

viggo-devries (changes requested):

viggo-devries (approved):

PR #4271, opened 2024-04-04
Fixes exception thrown when pickling/unpickling ProductAttributesContainer

crgwbr:
After #4167 and prior to this patch, trying to use pickle with a Product or ProductAttributesContainer results in a recursion error.

### DIFF
--- a/tests/unit/catalogue/test_product_attributes.py
+++ b/tests/unit/catalogue/test_product_attributes.py
@@ -1,3 +1,4 @@
+import pickle
 import unittest
 from copy import deepcopy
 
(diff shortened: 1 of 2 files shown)


crgwbr:
@specialunderwear The test failure here seems unrelated to my change, and a very similar issue seems to be occurring in master[^1]. Any ideas on why? 

[^1]: https://github.com/django-oscar/django-oscar/actions/runs/8510200510/job/23307293806

tests/unit/catalogue/test_product_attributes.py
```diff
@@ -332,6 +333,32 @@ def test_entity_attributes(self):
...
 
+    def test_can_be_pickled_when_initialized(self):
+        """Should be able to pickle and unpickle an initialized ProductAttributesContainer"""
+        product = Product.objects.get(pk=self.product.pk)
+
+        self.assertEqual(product.attr.weight, 3)
+        self.assertTrue(product.attr.initialized)
+
+        pickled_attrs = pickle.dumps(product.attr)
+        print(pickled_attrs)
```
specialunderwear, line 344: I think this has been left here by accident?

specialunderwear (changes requested):

specialunderwear:
> 

> @specialunderwear The test failure here seems unrelated to my change, and a very similar issue seems to be occurring in master[1](#user-content-fn-1-d65c59dc7dec8f57bb4ecf9a1d37be47). Any ideas on why?
> 
> ## Footnotes
> 1. https://github.com/django-oscar/django-oscar/actions/runs/8510200510/job/23307293806 [↩](#user-content-fnref-1-d65c59dc7dec8f57bb4ecf9a1d37be47)

Yeas we have some ideas and we're working on it, but we can't reproduce this locally unfortunately, can you?

tests/unit/catalogue/test_product_attributes.py
crgwbr: Fixed

crgwbr:
> Yeas we have some ideas and we're working on it, but we can't reproduce this locally unfortunately, can you?

Unfortunately, no. The tests all pass for me locally.

Issue #4219, opened 2024-01-02
TypeError in OrderSearchForm


pgcd:
https://github.com/django-oscar/django-oscar/blob/872e42a106d731b3bbef85cd37de7492d606ad13/src/oscar/apps/customer/forms.py#L211

The above line causes a TypeError when the form includes date_from and date_to with an empty order number.

Thomasvdh92:
Bump

PR #4274, opened 2024-04-11
Fix a typo which caused a typeerror on string formatting

Thomasvdh92:
Fix a typo where the string formatting caused a typeerror when only providing a datefrom and dateto, without an order number. See https://github.com/django-oscar/django-oscar/issues/4219


viggo-devries:
@Thomasvdh92 tests failing unfortunately.

samar-hassan (approved):

viggo-devries (approved):

PR #4280, opened 2024-04-22
fix: call _fetch_item from _process_image for easier override

Chadys:
In `catalogue.Importer`, we have the `_fetch_item` function as a practical override entry point.
But instead of using it, `_process_image` is reimplementing the same logic directly in the function.
I fixed it so that `_fetch_item` is used and I removed the now unnecessary lookup argument in `_process_image`.

With those changes, in `handle` L.41, the `lookup_value` is retrieved but is only used to construct some error logs. I think we should change the text of the logs so as not to need to retrieve this value just for them, but I wanted to discuss about it first.

(PS: I have been working using Oscar for more than 2 years now, building a marketplace from scratch. I finally have some times to give back to that awesome project by providing some fixes for issues I found along the way so I'll be pretty active in the following days!)

### DIFF
--- a/src/oscar/apps/catalogue/utils.py
+++ b/src/oscar/apps/catalogue/utils.py
@@ -39,7 +39,7 @@ class Importer(object):
             for filename in filenames:
                 try:
                     lookup_value = self._get_lookup_value_from_filename(filename)
-                    self._process_image(image_dir, filename, lookup_value)
+                    self._process_image(image_dir, filename)
                     stats["num_processed"] += 1
                 except Product.MultipleObjectsReturned:
                     self.logger.warning(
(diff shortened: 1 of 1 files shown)


Chadys:
I never used the `Importer` in production either, still it was useful to create some fixtures that I used in development, and I actually have needed to override `_fetch_item` and ended up overriding `_process_image` for that. Hence this little fix. But I agree that it is not that important.

I'm not working for that marketplace anymore (that's the reason why I finally have some time to myself haha), but some utilities to map catalogue data would definitely have been useful!
For the marketplace what I did was to implement a whole bunch of utility classes to update the catalogue in bulk and easily handle data from different sources (API, FTP, ...) by only needing to declare the specific format for a seller in a pydantic model and have a generic way to handle them all. That took a lot of time to implement so your new project is definitely a welcome addition to the ecosystem, I'll check it!

PR #4284, opened 2024-04-23
fix: typo in has_children value and tests of annotated list extra info

Chadys:
There was a typo in `has_children` computation, the wrong variable is used for the comparison. I also added a test for `get_annotated_list`'s `num_to_close` and `has_children` values.


samar-hassan (approved):

specialunderwear:
thanks!

PR #4289, opened 2024-04-23
fix: prevent checkout button display in basket

Chadys:
Some actions make the basket notification message display the "Checkout now" while we are on the basket page. This is a fix


samar-hassan (changes requested):
Thanks for all your contributions @Chadys. Please see if you can outline the steps to reproduce it. I tested this and found out include_buttons should be set to False in `basket.views.SavedView`. Kindly add it there as well.

PR #4316, opened 2024-06-11
Missed django 4.2 upgrade DeleteView override fix

maziar-dandc:


samar-hassan (changes requested):
Thanks for PR. You can return the delete method from form_valid and leave the delete method as it is, because earlier django versions still use this method.

samar-hassan (approved):

PR #4339, opened 2024-08-29
Fix typo max_basket_applications

benb10:
Fix a typo - I think it should be max_basket_applications instead of max_user_applications


viggo-devries:
wow, please add a test to check this specific condition.

PR #4368, opened 2024-10-07
When a notification is archinved, it should nolonger contribute the t…

specialunderwear:
…he unread notifications.

Therefor we mark it as read


gwaidacher:
Hi!

I've got the same `AttributeError: 'Settings' object has no attribute 'STORAGES'` in #4376, how can I fix this?

Issue #4370, opened 2024-10-17
Building sandbox fails on M1 Mac with recent node/npm, unless `node-sass` package is removed


TibsAtWork:
### Issue Summary

Building the sandbox with `make sandbox` (which runs `npm install`) fails when it tries to install `node-sass` on a recent M1 Mac with a recent version of node.

[node-sass](https://www.npmjs.com/package/node-sass) is deprecated and in maintenance mode.  [sass](https://www.npmjs.com/package/sass) is suggested as a replacement, and that's already present in the `package.json` file.

Removing the requirement for `node-sass` from the `package.json` file allows `make sandbox` to complete successfully.

### Steps to Reproduce

I'm on Mac M1 Sonoma 14.7, with Python 3.11.3, node v22.9.0 (using `brew install npm`) and with the xcode command line tools installed (using `xcode-select --install`).

Using either the Oscar branch `releases/3.2.4` or `master`, `make sandbox`.

```
$ cd django-oscar
$ python3 -m venv venv
$ source venv/bin/activate
$ make sandbox
```

Specifically, the `npm install` step fails to install [node-sass](https://www.npmjs.com/package/node-sass), first being unable to find a suitable binary:
```
npm error code 1
npm error path /Users/tony.ibbs/temp/django-oscar-2024-10-16/django-oscar/node_modules/node-sass
npm error command failed
npm error command sh -c node scripts/build.js
npm error Binary found at /Users/tony.ibbs/temp/django-oscar-2024-10-16/django-oscar/node_modules/node-sass/vendor/darwin-arm64-127/binding.node
npm error Testing binary
npm error Binary has a problem: Error: dlopen(/Users/tony.ibbs/temp/django-oscar-2024-10-16/django-oscar/node_modules/node-sass/vendor/darwin-arm64-127/binding.node, 0x0001): tried: '/Users/tony.ibbs/temp/django-oscar-2024-10-16/django-oscar/node_modules/node-sass/vendor/darwin-arm64-127/binding.node' (not a mach-o file), '/System/Volumes/Preboot/Cryptexes/OS/Users/tony.ibbs/temp/django-oscar-2024-10-16/django-oscar/node_modules/node-sass/vendor/darwin-arm64-127/binding.node' (no such file), '/Users/tony.ibbs/temp/django-oscar-2024-10-16/django-oscar/node_modules/node-sass/vendor/darwin-arm64-127/binding.node' (not a mach-o file)
```

[...]

PR #4371, opened 2024-10-17
Remove the deprecated package node-sass to allow `make sandbox` to succeed

TibsAtWork:
Fixes issue https://github.com/django-oscar/django-oscar/issues/4370

[node-sass](https://www.npmjs.com/package/node-sass) is deprecated, does not have binary versions for my M1 Mac (Sonoma 14.7, although I think it's for any `arm64`), and will not build for me (with node v22.9.0) due to a compiler error.

The `package.json` file already requires [sass](https://www.npmjs.com/package/sass), the recommended replacement. Removing the requirement for `node-sass` allows `npm install` (and thus `make sandbox`) to succeed.

**Note** changing `package.json` also leads to changes in `package-lock.json` (in fact, making it significantly shorter). I am not able to judge if the changes make sense for the project as a whole, or to what extent they reflect the particularities of my own setup.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -24,7 +24,6 @@
         "jquery-mousewheel": "^3.1.13",
         "jquery-sortable": "^0.9.13",
         "nan": "^2.22.0",
-        "node-sass": "^9.0.0",
         "popper.js": "^1.16.1",
         "pump": "^3.0.0",
         "sass": "^1.83.4",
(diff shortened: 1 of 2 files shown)


jmb:
This also fixed my sandbox setup - Mac M2.

samar-hassan:
@TibsAtWork 
Thanks for the contribution. package-lock.json in this PR appears to be out of date now. Could you please regenerate it against the current master branch and update the PR?

Thanks!

PR #4405, opened 2024-12-10
setup: make npm build optional

kunkku:
The distribution package does not contain `package.json` nor `gulpfile.js`. Ideally, building from the distribution package would not depend on access to the npm registry.


viggo-devries:
why would you build from the distribution package?

PR #4427, opened 2025-01-22
WIP: Codecov requires a token now for uploading?

joeyjurjens:


joeyjurjens:
This is not needed. Codecov 5 came with changes regarding uploading and we just didn't adopt them properly.

PR #4436, opened 2025-02-17
FEAT: Add postgres materialised view to the ProductCategory model to increase performance on the range queries

viggo-devries:


joeyjurjens (approved):

alstr:
I've just come around to updating my project with several partially forked Oscar apps and the `ProductCategoryHierearchy` model caused me a whole host of problems.

In the end I've had to fully fork the entire `offer` app and remove all references to `ProductCategoryHierearchy`, because Django just keeps complaining it's not registered even after running migrations. I tried changing around the order of some imports but this just created more issues.

I'm not sure about others' experiences, but as this is Postgres specific functionality maybe it would have been better as a separate app/addon.

joeyjurjens:
> I've just come around to updating my project with several partially forked Oscar apps and the `ProductCategoryHierearchy` model caused me a whole host of problems.
> 
> 
> 
> In the end I've had to fully fork the entire `offer` app and remove all references to `ProductCategoryHierearchy`, because Django just keeps complaining it's not registered even after running migrations. I tried changing around the order of some imports but this just created more issues.
> 
> 
> 
> I'm not sure about others' experiences, but as this is Postgres specific functionality maybe it would have been better as a separate app/addon.

Have you forked the catalogue app? And if so, did you do it like explained in the docs? Sounds like the model is not registering, which can be related to that. We didn't have any of these problems and have been running it since it was merged.

alstr:
Yep, these are all my Oscar app forks:

Screenshot 2026-06-03 at 14 11 22

And in `models.py` I have `from oscar.apps.catalogue.models import *  # noqa isort:skip` at the bottom.

I am sure it probably does work for other people. Given the highly customised nature of my project there must be something causing it somewhere but I don't know what. I spent ages trying to figure it out but just gave up in the end and removed `ProductCategoryHierarchy`, which fixed everything. I don't want to divert this but making customisations more maintainable would be my number one request for Oscar, as right now I have essentially what is a separate project and incorporating updates is always tricky.

joeyjurjens:
> Yep, these are all my Oscar app forks:
> 
> Screenshot 2026-06-03 at 14 11 22
> And in `models.py` I have `from oscar.apps.catalogue.models import * # noqa isort:skip` at the bottom.
> 
> I am sure it probably does work for other people. Given the highly customised nature of my project there must be something causing it somewhere but I don't know what. I spent ages trying to figure it out but just gave up in the end and removed `ProductCategoryHierarchy`, which fixed everything. I don't want to divert this but making customisations more maintainable would be my number one request for Oscar, as right now I have essentially what is a separate project and incorporating updates is always tricky.

Yeah not sure what the issue is without seeing the code. I agree that upgrades can be painful when you have lots of customisations. We try to include related information in release notes. However oscar's way of customisations is just a bit different than other systems.

alstr:
> > Yep, these are all my Oscar app forks:
> > Screenshot 2026-06-03 at 14 11 22
> > And in `models.py` I have `from oscar.apps.catalogue.models import * # noqa isort:skip` at the bottom.
> > I am sure it probably does work for other people. Given the highly customised nature of my project there must be something causing it somewhere but I don't know what. I spent ages trying to figure it out but just gave up in the end and removed `ProductCategoryHierarchy`, which fixed everything. I don't want to divert this but making customisations more maintainable would be my number one request for Oscar, as right now I have essentially what is a separate project and incorporating updates is always tricky.
> 
> Yeah not sure what the issue is without seeing the code. I agree that upgrades can be painful when you have lots of customisations. We try to include related information in release notes. However oscar's way of customisations is just a bit different than other systems.

Understood. Anyway it's working for now so I can live with it. My desire to debug further was pretty low after updating the site to Bootstrap 5. Time for a lie down.

PR #4438, opened 2025-02-19
Issa

issasalmans:


viggo-devries:
This does not really feel like a core function to me, you could add this as a plugin. but I don't see this being merged into our core. Thank you tho!

PR #4446, opened 2025-03-02
fixed zipcode ordering

svandeneertwegh:
Changed everywhere the ordering of fields in an address.

Maybe changing line4 to city name in model?
Maybe in the settings we can specify one field lists and use it everywhere? Maybe something like this settings: 

OSCAR_REQUIRED_ADDRESS_FIELDS = (
    "first_name",
    "last_name",
    "line1",
    "postcode",
    "line4",
    "country",
)


svandeneertwegh:
#4446

svandeneertwegh:
Following ticket #4446

viggo-devries:
I think this is really country specific is it not? In the Netherlands, yes we want this order, but 

what could be useful is if we add settings that dictate the order of fields on the forms. 

something like

OSCAR_CHECKOUT_SHIPPING_ADDRESS_FORM_FIELD_ORDER = []

PR #4455, opened 2025-04-02
[Fix] Resolve rounding errors

samar-hassan:
Calculate basket total, tax and discounts upto 4 decimal places, resolving rounding errors.

### DIFF
--- a/src/oscar/apps/partner/strategy.py
+++ b/src/oscar/apps/partner/strategy.py
@@ -277,7 +277,7 @@ class FixedRateTax(object):
     """
 
     rate = D("0")  # Subclass and specify the correct rate
-    exponent = D("0.01")  # Default to two decimal places
+    exponent = D("0.0001")  # Default to two decimal places
 
     def pricing_policy(self, product, stockrecord):
         if not stockrecord or stockrecord.price is None:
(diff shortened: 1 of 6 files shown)


src/oscar/apps/partner/strategy.py
```diff
@@ -277,7 +277,7 @@ class FixedRateTax(object):
     """
 
     rate = D("0")  # Subclass and specify the correct rate
-    exponent = D("0.01")  # Default to two decimal places
+    exponent = D("0.0001")  # Default to two decimal places
```
viggo-devries, line 280: change comment, default to four decimal places

src/oscar/core/utils.py
```diff
@@ -174,12 +174,30 @@ def round_half_up(money):
...
     >>> should_not_be_one.quantize(exponent)
     Decimal('1.00')
-    >>> round_half_up(should_not_be_one)
+    >>> round_half_up_two_dec(should_not_be_one)
     Decimal('1.01')
     """
     return money.quantize(decimal.Decimal("0.01"), decimal.ROUND_HALF_UP)
 
 
+round_half_up = round_half_up_two_dec
```
viggo-devries, line 183: I think we should make this deprecated.

viggo-devries (changes requested):
Please add some docs or alter existing docs for this aswell.

viggo-devries (approved):

PR #4457, opened 2025-04-07
[Fix] Resolve multiple backend error in race condition

samar-hassan:


joeyjurjens (changes requested):
I'd rather have us pass backend to the login function than setting it as attribute on the user.
https://github.com/django-oscar/django-oscar/blob/21d66d1a8eb14aecb8ed1a3d80023e6aeb13337c/src/oscar/apps/customer/mixins.py#L81

samar-hassan:
> I'd rather have us pass backend to the login function than setting it as attribute on the user.
> 
> https://github.com/django-oscar/django-oscar/blob/21d66d1a8eb14aecb8ed1a3d80023e6aeb13337c/src/oscar/apps/customer/mixins.py#L81

`authenticate` function in the try block already sets backend attribute of the user. Using it in the except block keeps it consistent.

PR #4461, opened 2025-04-30
WIP: Replace join query with CTE for the range category lookup.

joeyjurjens:
**WIP** because I want to test this on a shop with lots of categories first.

This PR replaces the range query for the category lookup with a CTE query.

CTE's are known to be powerful for querying hierarchical data (and recommended so).

This should improve the performance a lot for shops with large categorie trees, since the CTE processes the hierarchy level-by-level instead of comparing all possible category combinations in a single massive join when dealing with large category trees.

It's similar to [FEAT: Add postgres materialised view to the ProductCategory model to increase performance on the range queries](https://github.com/django-oscar/django-oscar/pull/4436) and this would probably still be best to implement if you can as it's precomputed (but needs refreshes) but this PR provides a better and more performant fallback.

### DIFF
--- a/src/oscar/apps/catalogue/expressions.py
+++ b/src/oscar/apps/catalogue/expressions.py
@@ -1,36 +1,59 @@
 from django.db.models.expressions import Subquery
 
-EXPAND_UPWARDS_CATEGORY_QUERY = """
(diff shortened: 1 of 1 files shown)


joeyjurjens:
I just tested this, and it seems like the JOIN method IS actually faster.

PR #4480, opened 2025-06-25
Admin logout button should use POST as django 5 no longer supports is through GET.

joeyjurjens:
Django 5 removed support for logging out users through a GET request.

The frontend uses a custom logout view, which we might need to deprecate too. But this should be done gracefully as it could break existing sites.

So this should be done in another PR.

It's not really a security risk, as an attacker could only logout the user and that's it. But it's best to do it the django way.

### DIFF
--- a/src/oscar/templates/oscar/dashboard/catalogue/attribute_option_group_delete.html
+++ b/src/oscar/templates/oscar/dashboard/catalogue/attribute_option_group_delete.html
@@ -44,7 +44,7 @@
         <div class="table-header">
             <h2>{% trans "Delete Attribute Option Group" %}</h2>
         </div>
-        <form method="post" class="card card-body">
+        <form id="attribute_delete_form" method="post" class="card card-body">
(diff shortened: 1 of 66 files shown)


samar-hassan (approved):

alstr:
The normal customer account page still uses a GET request for logout. Should this be updated too for consistency?

joeyjurjens:
> The normal customer account page still uses a GET request for logout. Should this be updated too for consistency?

We need to deprecate it first. Otherwise it might silently break existing sites that don't directly uses oscar templates.

See:
https://github.com/django-oscar/django-oscar/issues/4481

alstr:
> > The normal customer account page still uses a GET request for logout. Should this be updated too for consistency?
> 
> We need to deprecate it first. Otherwise it might silently break existing sites that don't directly uses oscar templates.
> 
> See: #4481

Makes sense, just seeing where it's at before I build my own.

Issue #4482, opened 2025-06-28
Error when edit offer benefit with empty range.


Aimage:
Way to reproduce.
Go to offer list and choose an offer with range benefit required (ex: Discount is a fixed amount off of the basket's total). Click on edit button to edit the benefit.

![Image](https://github.com/user-attachments/assets/455d33f5-b8f3-459b-9c43-ca4071032084)

Set an empty range, then save.
The bellow error appears. I think the model validation is not called when updating.
I reproduce the problem locally and on sandbox online.

<html>
<body>
<h1 style="padding: 0px; margin: 0px; font-weight: normal; color: rgb(0, 0, 0); font-family: sans-serif; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">AttributeError at /dashboard/offers/1/incentive/</h1><pre class="exception_value" style="padding: 0px; margin: 10px 0px; font-size: 1.5rem; white-space: pre-wrap; word-break: break-word; font-family: sans-serif; color: rgb(87, 87, 87); font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">'NoneType' object has no attribute 'name'</pre>
Request Method: | POST
-- | --
http://localhost:8000/dashboard/offers/1/incentive/
5.2
AttributeError
'NoneType' object has no attribute 'name'

</body>
</html>

PR #4489, opened 2025-07-08
Avoid crash when Benefit.range is None

mfarhan943:
## Description
This PR resolves: https://github.com/django-oscar/django-oscar/issues/4482

When saving an edited Incentive offer form without providing a "Range" value, a 500 Internal Server Error is raised. This happens because the name property in the PercentageDiscountBenefit class tries to access:
`self._description % {"value": self.value, "range": self.range.name}`

If self.range is None, this raises an AttributeError, since None has no .name.
The error originates from the `description` property in `BaseOfferMixin`, which calls self.name. Because of the missing range, it fails before validation logic (like clean()) can run and display a user-friendly error.

## Fix
Update the `description` property in `BaseOfferMixin` to safely handle attribute errors by wrapping the call in a try block. This allows `clean()` to execute and return a proper validation error message instead of a server error:

## Result
Screenshot 2025-07-08 at 1 59 55 PM

### DIFF
--- a/src/oscar/apps/offer/abstract_models.py
+++ b/src/oscar/apps/offer/abstract_models.py
@@ -82,7 +82,10 @@ class BaseOfferMixin(models.Model):
         A description of the benefit/condition.
         Defaults to the name. May contain HTML.
         """
-        return self.name
+        try:
+            return self.name
+        except AttributeError:
+            pass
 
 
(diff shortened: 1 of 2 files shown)


awais786:
@samar-hassan Please review this PR.

PR #4493, opened 2025-07-22
Add custom dashboard permission checks

mfarhan943:


mfarhan943:
@samar-hassan I have added conditional rendering using user.has_perm. Could you confirm if this is the correct approach?Thanks!

samar-hassan:
> @samar-hassan I have added conditional rendering using user.has_perm. Could you confirm if this is the correct approach?Thanks!

You have to use the permissions from DashboardPermission class, so that there is no inconsistency between the permissions in the view and the templates.

mfarhan943:
@samar-hassan Could you please take a look when you get a chance?

samar-hassan:
We don’t want to hide stats based on dashboard permissions — those should remain visible. However, we do want to hide navigation links to other dashboard pages if the user lacks the required permission.

This should be implemented after breaking down the dashboard permissions for individual views.

For example:

https://github.com/django-oscar/django-oscar/blob/9c8a8fca258e803ee15b29ae5d917ccaad103e85/src/oscar/apps/dashboard/offers/apps.py#L21

this should be updated to:

```"offer-list": DashboardPermission.get("view-offer"),```

The "view-offer" permission should be added to the DashboardPermission class.

Finally, any buttons or links on any dashboard page that point to `dashboard:offer-list` should be conditionally rendered based on whether the user has the appropriate permission.

File: src/oscar/apps/dashboard/permissions.py [hunk 1]
```diff
@@ -140,6 +140,23 @@ class DashboardPermission:
...
+        ],
+        "edit-order": [
+            "is_staff",
+            "order.change_order",
+        ],
+        "view-product": [
+            "is_staff",
+            "catalogue.view_product",
+        ],
+        "view-stock": [
```
samar-hassan, line 156: This should be updated to view-stock-alert. `partner.view_stockalert`. is_staff seems redundant. Please add a default permission attribute and then add merge it in the get method.

Issue #4492, opened 2025-07-18
Improve EmailUserCreationForm to validate or filter invalid redirect_url to prevent bot registrations


dismine:
On my website, I noticed that bots are submitting spam values to the hidden redirect_url field in EmailUserCreationForm, which leads to errors such as: `Reverse for 'kynifhmd' not found. 'kynifhmd' is not a valid view function or pattern name.`
These bots are able to register successfully, even though the submitted redirect URL is clearly invalid or meaningless.

The current implementation of EmailUserCreationForm (and potentially similar forms) does not strictly validate that the redirect_url field points to a valid and resolvable view. 

**Proposed improvement:**
* The form should more strictly validate redirect_url, for example by:
* Ensuring it passes url_has_allowed_host_and_scheme
* Validating that it resolves to a real Django view using django.urls.resolve()
* Raising a ValidationError or silently falling back to settings.LOGIN_REDIRECT_URL if the input is invalid
* Optionally logging suspicious values

This would improve the form’s resilience against bots and reduce avoidable server errors and log noise.

Thanks for your work on Oscar!

mfarhan943:
I am interested in working on this. Kindly assign it to me.

PR #4495, opened 2025-07-29
Validate redirect_url to prevent broken redirects

mfarhan943:
###Description:
Adds stricter validation for the hidden `redirect_url` field in EmailUserCreationForm.

- Attempts to resolve internal paths via `resolve()`
- Falls back to `LOGIN_REDIRECT_URL` if invalid or unresolvable
- Adds tests to cover unresolvable internal and unsafe external redirects

This improves form resilience against bots and prevents server errors due to invalid redirect targets.

Resolves: https://github.com/django-oscar/django-oscar/issues/4492

### DIFF
--- a/src/oscar/apps/customer/forms.py
+++ b/src/oscar/apps/customer/forms.py
@@ -8,6 +8,7 @@ from django.contrib.auth.forms import AuthenticationForm
 from django.contrib.auth.password_validation import validate_password
 from django.contrib.sites.shortcuts import get_current_site
 from django.core.exceptions import ValidationError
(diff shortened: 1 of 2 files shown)


samar-hassan:
Thanks for your PR. Reducing unnecessary server logs makes sense. However, could you clarify how this change improves the form’s resilience against bots? Bots can still register.

mfarhan943:
@samar-hassan No, this change does not block bot registrations. Rather, it improves the form’s resilience by preventing invalid or unresolvable redirect targets from causing server-side errors.

PR #4513, opened 2025-09-05
Added a fix for ordering VoucherSets on properties aggregated from re…

deemka:
…lated Vouchers

### DIFF
--- a/src/oscar/apps/dashboard/vouchers/views.py
+++ b/src/oscar/apps/dashboard/vouchers/views.py
@@ -4,7 +4,7 @@ import csv
 from django.conf import settings
 from django.contrib import messages
 from django.db import transaction
-from django.db.models import Count
+from django.db.models import Count, Sum
 from django.http import HttpResponse
 from django.shortcuts import get_object_or_404, redirect
 from django.urls import reverse, reverse_lazy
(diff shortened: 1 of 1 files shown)


src/oscar/apps/dashboard/vouchers/views.py
```diff
@@ -311,7 +311,10 @@ class VoucherSetListView(generic.ListView):
     paginate_by = settings.OSCAR_DASHBOARD_ITEMS_PER_PAGE
 
     def get_queryset(self):
-        qs = self.model.objects.all().order_by("-date_created")
+        qs = self.model.objects.all()
```
samar-hassan, line 314: Thanks for the changes! 🙌 Could you also add a test to cover this case?

deemka: @samar-hassan Fair point. I had to set up the venv first, and the changes seemed trivial.
However, I realized, that I had to also add "fake" property setters for properties in question. Because aggregation tries to assign those and fails if no setter is defined. The setters look odd, but it seemed the cleanest way to add aggregation without changing the model or `sort_queryset()` to me.

I also changed the default values for `VoucherFactory.num_basket_additions`, `VoucherFactory.num_orders` to random numbers, explicitly setting them to zero where zero is expected or required. Let me know if it should the other way around - 0 by default and setting values where necessary. I was in doubt.

samar-hassan: @deemka please skip updating any existing tests. No need for the setters as well. Just call the view and check if the queryset is ordered correctly.

PR #4537, opened 2025-11-18
[FEAT] Bulk update public visibility in product and category listing

samar-hassan:
image


joeyjurjens:
> Bulk actions can actually be added anywhere on the page. For example, the order listing page already includes two separate bulk-action components—one in the header and another at the bottom of the table.
> 
> In the same way, custom bulk actions can be introduced wherever they make the most sense in the UI. Given that, placing a simple radio-button bulk action for is_public feels consistent with how similar actions (like the “make active” option in the customer listing) are already handled.
> 
> image

Hm okay, if the order does it already I guess it's fine for now, it's not like it's going to improve the rest of the dashboard if you're doing it the way I'm saying. Some day, the dashboard hopefully gets a good rewrite and things like this are properly solved 👍🏻

PR #4556, opened 2026-02-11
deps: compatibility with django-treebeard 5.0

solarissmoke:
Treebeard v5 is now available. This PR:

- Adds compatibility for Treebeard 5.0
- Raises minimum version of Treebeard to 4.7 (earliest version to declare support for Django 4.2)

Note that the tests on the main branch will fail until this is applied, because Oscar doesn't impose an upper limit on the version dependency for treebeard.

### DIFF
--- a/src/oscar/apps/catalogue/abstract_models.py
+++ b/src/oscar/apps/catalogue/abstract_models.py
@@ -264,8 +264,8 @@ class AbstractCategory(MP_Node):
         return children.filter(is_public=True)
 
     @classmethod
-    def fix_tree(cls, destructive=False, fix_paths=False):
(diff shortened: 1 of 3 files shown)


src/oscar/apps/catalogue/abstract_models.py
```diff
@@ -333,10 +333,10 @@ class Meta:
         verbose_name_plural = _("Categories")
 
     def has_children(self):
-        return self.get_num_children() > 0
+        return self.get_children().exists()
```
solarissmoke: Making this query more efficient - we don't need a count to know if children exist.

Issue #2705, opened 2018-05-21
Oscar should stop using inline Javascript to allow use of CSP policies


solarissmoke:
Oscar currently inserts inline Javascript [everywhere on the front-end](https://github.com/django-oscar/django-oscar/blob/master/src/oscar/templates/oscar/base.html#L50-L54). Inline scripts are considered a vector for XSS attacks, and increasingly we are using [content security policies](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) to mitigate against this threat.

In order to facilitate the use of CSP Oscar should remove all use of inline scripts in its codebase. This will also mean removing the `onbodyload` template block and refactoring page-specific JS to be stored in standalone JS files.

I'd propose to make this change in Oscar 2 as it will not be backward compatible.

sasha0:
It might also make sense for deeper client-code revision. Thoughts?

specialunderwear:
I will simply mark this a bug. But note that dashboard is much worse in this regard and must also be fixed.

PR #4561, opened 2026-03-04
fix: Sanitize HTML in description fields to prevent stored XSS ( GHSA-rvcc-4w2x-g7xv)

hofi199410:
## Summary
This PR fixes a stored XSS vulnerability in the dashboard by sanitizing 
HTML input in description fields using the `nh3` library.

## Problem
An authenticated staff user could inject arbitrary JavaScript into the 
`description` field of Products and Categories. The payload was persisted 
and executed in the browser of every user viewing the affected page.

## Solution
Added `clean_description()` methods to `ProductForm` and `CategoryForm` 
that sanitize HTML input using `nh3`. Safe HTML tags like `<p>`, `<b>`, 
`<a>` etc. are preserved so rich text content continues to work as expected.

## Changes
- `src/oscar/apps/dashboard/catalogue/forms.py`: Added `clean_description()` 
  to `ProductForm` and `CategoryForm`
- `tests/integration/dashboard/catalogue/test_forms_xss.py`: Added tests 
  to verify XSS payloads are stripped and safe HTML is preserved

## Testing
- Verified `<script>alert(1)</script>` is stripped after fix
- Verified `<p><b>Bold</b></p>` is preserved
- All tests passing

Fixes  GHSA-rvcc-4w2x-g7xv

### DIFF
--- a/src/oscar/apps/dashboard/catalogue/forms.py
+++ b/src/oscar/apps/dashboard/catalogue/forms.py
@@ -6,6 +6,8 @@ from treebeard.forms import movenodeform_factory
 from oscar.core.loading import get_class, get_classes, get_model
 from oscar.core.utils import slugify
 from oscar.forms.widgets import DateTimePickerInput, ImageInput
+import nh3
+
 
 Product = get_model("catalogue", "Product")
 ProductClass = get_model("catalogue", "ProductClass")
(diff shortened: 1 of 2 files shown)


joeyjurjens (commented):
I don't think we see this as vulnerability, because you need admin access and certain permissions to be able to set product descriptions.

We've had reports like this before, but it's always been that we by default trust the people that have admin access.

What exactly does NH3 do, when you don't pass any params?

Because I think stripping all JS isn't a good call either. People might want to load something some JS for their descriptions.

If we do this here, we have to do it in multiple places.

Maybe it can be made as optional feature, but you can always escape it in the template (don't render it as |safe). Also, if you're really worried about XSS you can setup CSP headers in your Django project.

@samar-hassan what do you think?

Also see related issues / reports:

https://github.com/django-oscar/django-oscar/issues/2705
https://huntr.com/bounties/c6ed0ec6-bcad-41a4-8f9a-5b0db7859db6

PR #4563, opened 2026-03-17
Fix #3056: Remove incorrect stock rules docs and add proper documentation

### LINKED ISSUE #3056: Documentation for enforcing stock rules is incorrect
The documentation at https://django-oscar.readthedocs.io/en/latest/howto/enforcing_stock_rules.html is incorrect - doing that will cause a 500 error because that exception isn't caught anywhere.

I think we just need to remove this, and instead enhance the [documentation about availability policies](https://django-oscar.readthedocs.io/en/latest/topics/prices_and_availability.html#the-strategy-class), which is where this sort of logic should happen.

lindsey-comet:
Fixes #3056

The 'enforcing_stock_rules' documentation was incorrect because it suggested raising `InvalidBasketLineError` in a `pre_save` signal handler, which isn't caught anywhere and causes 500 errors when customers exceed quantity limits.

## Changes:
- Removed `docs/source/howto/enforcing_stock_rules.rst` (the incorrect documentation)
- Updated `docs/source/howto/index.rst` to remove the reference to the deleted file
- Added proper documentation in `prices_and_availability.rst` explaining how to enforce stock rules using availability policies, which is the correct Oscar pattern

## The Correct Approach
The new documentation explains that stock validation rules should be implemented via custom availability policies by overriding the `is_purchase_permitted()` method. This integrates with Oscar's validation flow, provides proper error messages to customers, and avoids the 500 error issue.

## Before
Following the old documentation would cause 500 errors because `InvalidBasketLineError` is not caught by any middleware or view - it's designed for internal validation, not customer-facing limits.

## After
The new documentation explains how to create a custom `Availability` policy subclass that enforces limits properly, returning an `AvailabilityResult` with appropriate messaging instead of raising an exception.


docs/source/topics/prices_and_availability.rst
```diff
@@ -230,6 +230,59 @@ There are also several predefined availability policies:
    :members: Unavailable, Available, StockRequired
    :noindex:
 
+Enforcing stock rules
+~~~~~~~~~~~~~~~~~~~~~
+
+To enforce custom stock validation rules (such as limiting the maximum quantity
```
samar-hassan, line 236: @lindsey-comet thanks for your PR, please update the text without mentioning signal thing. Might be confusing for new users.

docs/source/topics/prices_and_availability.rst
```diff
@@ -230,6 +230,59 @@ There are also several predefined availability policies:
...
+rather than raising exceptions in basket signals.
+
+The correct approach is to subclass the appropriate availability policy and
+override the ``is_purchase_permitted`` method:
+
+.. code-block:: python
+
+    from oscar.apps.partner import availability
+
+    class LimitedStock(availability.StockRequired):
```
samar-hassan, line 247: `class LimitedStock(availability.Base):`
please add an example with availability Base class

docs/source/topics/prices_and_availability.rst
```diff
@@ -230,6 +230,59 @@ There are also several predefined availability policies:
...
+        """
+        Availability policy that enforces a maximum purchase quantity.
+        """
+        max_purchase_quantity = 4
```
samar-hassan, line 251: please add code and message.

docs/source/topics/prices_and_availability.rst
```diff
@@ -230,6 +230,59 @@ There are also several predefined availability policies:
...
+
+Then use this policy in your custom strategy:
+
+.. code-block:: python
+
+    from oscar.apps.partner import strategy
+
+    class LimitedStockStrategy(strategy.UseFirstStockRecord,
+                               strategy.Structured):
+        availability_policy = LimitedStock
```
samar-hassan, line 278: `def availability_policy(self, product, stockrecord):
     return LimitedStock()`

PR #4564, opened 2026-03-19
fix #4557: support Django 6 by replacing StartsWith subclass with Lookup

### LINKED ISSUE #4557: Support Django 6
I tried running django-oscar tests with a Django 6.0.2, for the moment tests are not passing due to a missing fix in their 6.0.x branch, cf. https://code.djangoproject.com/ticket/36922

jobcase-ebrake:
Django 6 changed internals of StartsWith in a way that broke the ReverseStartsWith subclass. Replace it with a direct Lookup subclass that implements as_sql() with swapped LHS/RHS, avoiding the dependency on StartsWith internals entirely. Also extend the Django version constraint to <=6.0.4.


JulienPalard:
I'm not fluent with Django ORM internals but it looks good to me. And the tests that were failing are now passing.

jvzammit (approved):

joeyjurjens (changes requested):
Thanks for working on Django 6 support! I think a smaller fix works here.

Per Django ticket [#36922](https://code.djangoproject.com/ticket/36922) and the [6.0 release notes](https://docs.djangoproject.com/en/6.0/releases/6.0/#custom-orm-expressions-should-return-params-as-a-tuple), the crash is a list-vs-tuple mismatch introduced by Django 6.0's mid-migration of lookup params. In 6.0.x, `PatternLookup.process_rhs` returns a tuple while `BuiltinLookup.as_sql` still calls `.extend()` on the lhs params — with the swap, the tuple ends up in the lhs position and breaks.

I think this works too, and it's less code to maintain since we keep depending on Django's `PatternLookup` for wildcard escaping and backend-correct `LIKE` handling:

```python
@Field.register_lookup
class ReverseStartsWith(StartsWith):
    lookup_name = "rstartswith"

    def process_lhs(self, compiler, connection, lhs=None):
        if lhs is not None:
            raise Exception("Flipped process_lhs does not accept lhs argument")
        sql, params = super().process_rhs(compiler, connection)
        return sql, list(params)

    def process_rhs(self, compiler, connection):
        return super().process_lhs(compiler, connection)
```

Should work on Django 4.2, 5.x, 6.0.x, and 6.1+.

joeyjurjens (approved):

PR #4570, opened 2026-05-21
rework bulk actions in the catalogue dashboard

mr-sphere:
Rework the bulk action mechanism. Now bulk actions can be made with special bulk action objects. also add a new kind of bulk action that lets you implement an intermediate step with a form. this form is then passed into the action execution which allows the enter parameters into the action.

useful in the case of updating child products' prices for example.  we can now select which children we want, and input a price as well. then all the prices of the selected children will then be updated to the entered price.


src/oscar/apps/dashboard/catalogue/bulk_actions.py
```diff
@@ -0,0 +1,120 @@
...
+Product = get_model("catalogue", "Product")
+StockRecord = get_model("partner", "StockRecord")
+
+
+class MakePublicAction(BulkAction):
+    label = _("Make public")
+
+    @atomic
+    def execute(self, request, records):
+        for record in records:
```
joeyjurjens: This isn't performant. It's doing N+1 queries for saving. Use django's bulk_update instead.

src/oscar/apps/dashboard/catalogue/bulk_actions.py
```diff
@@ -0,0 +1,120 @@
...
+            % {"count": len(records)},
+        )
+
+
+class MakeNonPublicAction(BulkAction):
+    label = _("Make non-public")
+
+    @atomic
+    def execute(self, request, records):
+        for record in records:
```
joeyjurjens: This isn't performant. It's doing N+1 queries for saving. Use django's bulk_update instead.

src/oscar/apps/dashboard/catalogue/bulk_actions.py
```diff
@@ -0,0 +1,120 @@
...
+    def execute(self, request, child_ids, form):
+        raise NotImplementedError
+
+
+class MakeChildrenPublicAction(ChildBulkAction):
+    label = _("Make children public")
+
+    @atomic
+    def execute(self, request, child_ids, form):
+        count = Product.objects.filter(
```
joeyjurjens: Please double check if count is always returned. I think some DB drivers might not return the count after an update call.

src/oscar/apps/dashboard/catalogue/bulk_actions.py
```diff
@@ -0,0 +1,120 @@
...
+            % {"count": count},
+        )
+
+
+class MakeChildrenNonPublicAction(ChildBulkAction):
+    label = _("Make children non-public")
+
+    @atomic
+    def execute(self, request, child_ids, form):
+        count = Product.objects.filter(
```
joeyjurjens: Please double check if count is always returned. I think some DB drivers might not return the count after an update call.

src/oscar/apps/dashboard/catalogue/bulk_actions.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+class SetChildrenPriceAction(ChildBulkAction):
+    label = _("Update children price")
+    form_class = SetChildrenPriceForm
+    template = "oscar/dashboard/catalogue/product_children_bulk_action_set_price.html"
+
+    @atomic
+    def execute(self, request, child_ids, form):
+        count = StockRecord.objects.filter(
```
joeyjurjens: Please double check if count is always returned. I think some DB drivers might not return the count after an update call.

src/oscar/views/generic.py
```diff
@@ -1,13 +1,37 @@
...
+
+
+class IntermediateBulkAction(BulkAction):
+    """Base class for two-step bulk actions. Intended to be used with IntermediateBulkEditMixin."""
+
+    form_class = None
+    template = "oscar/dashboard/intermediate_bulk_action.html"
+    context = {}
+
+    def execute(self, request, child_ids, form):  # pylint: disable=arguments-differ
```
joeyjurjens: If you're adding an additional arg to this function here, what's the point of the base BulkAction then?

src/oscar/apps/dashboard/catalogue/forms.py
```diff
@@ -448,3 +448,31 @@ class Meta:
...
+        error_messages={"required": _("Select at least one child product.")},
+    )
+
+    def __init__(self, *args, children_queryset=None, **kwargs):
+        super().__init__(*args, **kwargs)
+        if children_queryset is not None:
+            self.fields["selected_children"].queryset = children_queryset
+
+
+class SetChildrenPriceForm(ChildrenBulkActionForm):
```
joeyjurjens: Can we add things like: increase by (percentage, amount). Or have possible override per child possible? Childs aren't always the same price this is way too limited.

joeyjurjens (changes requested):

src/oscar/apps/dashboard/catalogue/bulk_actions.py
mr-sphere: "Performs an SQL update query for the specified fields, and returns the number of rows matched (which may not be equal to the number of rows updated if some rows already have the new value)."

from https://docs.djangoproject.com/en/6.0/ref/models/querysets/#django.db.models.query.QuerySet.update

seems to always return a number

src/oscar/views/generic.py
mr-sphere: bulk actions without an intermediate step do not have a form passed through from the intermediate step.

src/oscar/apps/dashboard/catalogue/bulk_actions.py
mr-sphere: the old bulk action did it this way, updated it to do 1 query

src/oscar/views/generic.py
joeyjurjens: Yeah I get that. But what's the point of the base `BulkAction` then? It defines an `execute` method, which is unimplemented and implementors like this change the method signature. To me this sounds like a wrong abstracted base class, which is pointless because implementations differ and thus you can never trust how to call . execute() on an bulk action.

src/oscar/views/generic.py
```diff
@@ -1,13 +1,37 @@
...
+    def execute(self, request, records):
+        raise NotImplementedError
+
+
+class IntermediateBulkAction(BulkAction):
+    """Base class for two-step bulk actions. Intended to be used with IntermediateBulkEditMixin."""
+
+    form_class = None
+    template = "oscar/dashboard/intermediate_bulk_action.html"
+    context = {}
```
joeyjurjens: I don't think you use this so far, and if you are, it's prone to bugs because dicts are mutable in Python. Meaning that if you ever somewhere do something like:

```python
class MyAction(IntermediateBulkAction):
    def execute(self, request, child_ids, form):
        self.context["foo"] = "bar"
        ...
```

Every other IntermediateBulkAction subclass will have that "foo" thing.

So either remove it (as you dont use it) or set it to None as default.

src/oscar/apps/dashboard/catalogue/bulk_actions.py
```diff
@@ -0,0 +1,179 @@
...
+        if not records:
+            return
+        model = type(records[0])
+        count = model.objects.filter(pk__in=[r.pk for r in records]).update(
+            is_public=True
+        )
+        messages.success(
+            request,
+            ngettext(
+                "Public status was successfully updated for %(count)d record.",
```
joeyjurjens: Maybe state what the status was set to? Same for the non public one. A bit clearer then.

src/oscar/views/generic.py
```diff
@@ -98,6 +137,150 @@ def get_object_dict(self, ids):
...
+            "Subclasses of IntermediateBulkEditMixin must implement get_intermediate_url()"
+        )
+
+    def post(self, request, *args, **kwargs):
+        action = request.POST.get(self.action_param, "").lower()
+        if action in self.intermediate_actions:
+            return self._handle_intermediate_action(request, action, *args, **kwargs)
+        return super().post(request, *args, **kwargs)
+
+    def _handle_intermediate_action(self, request, action, *args, **kwargs):
```
joeyjurjens, line 218: In this method you trust user input directly, eg:

```python
                    request.POST.getlist(
                        "selected_%s" % self.get_checkbox_object_name()
                    ),
```

If a certain partner wants to do harm to another partner, they could change checkbox, POST it and then your view happily passes it on to the next view.

In the next view you can then change prices / whatever from another partner.

Should be something like this:

```python
def _handle_intermediate_action(self, request, action, *args, **kwargs):
    select_across = request.POST.get(self.select_across_param, "").lower()
    if select_across == "1":
        parent_pks = list(self.get_queryset().values_list("pk", flat=True))
    else:
        posted_pks = request.POST.getlist(
            "selected_%s" % self.get_checkbox_object_name()
        )
        parent_pks = list(
            self.get_queryset()
            .filter(pk__in=posted_pks)
            .values_list("pk", flat=True)
        )
```

Changing items to a int isn't needed either, Django handles that automatically. Plus if the PK wasn't an int it would crash too. So just do pk__in=values directly without converting it to int

joeyjurjens (changes requested):

src/oscar/templates/oscar/dashboard/catalogue/product_children_bulk_action_set_price.html
```diff
@@ -0,0 +1,33 @@
...
+{% endif %}
+{% endblock %}
+
+{% block extra_column_headers %}
+<th>{% trans "Lowest current price" %}</th>
+<th>{% trans "Specific price" %}</th>
+{% endblock %}
+
+{% block extra_column_values %}
+{% with sr=child.cheapest_stockrecord price_field=form|child_price_field:child.pk %}
```
joeyjurjens: Do we only allow updating the cheapest stockrecord? What if you have more?

src/oscar/views/generic.py
mr-sphere: i have refactored it because this logic should also be in the regular BulkEditMixin.

src/oscar/templates/oscar/dashboard/catalogue/product_children_bulk_action_set_price.html
mr-sphere: this is only for display. because we cannot display all stockrecords in a single table cell. the action itself updates all stockrecords that fit within the selected partner filter.

src/oscar/apps/dashboard/catalogue/forms.py
mr-sphere: you can now set a specific price per product, modify by amount and modify by percentage.

src/oscar/views/generic.py
```diff
@@ -63,41 +121,258 @@ def get_error_url(self, request):
...
+            return redirect(self.get_error_url(request))
+
+        objects = self.get_objects(ids)
 
-            objects = self.get_objects(ids)
+        action_handler = actions[action]
 
-        return getattr(self, action)(request, objects)
+        if hasattr(action_handler, "execute"):
+            action_handler.execute(request, objects)
```
samar-hassan, line 167: Instead of turning the ids into a list of loaded objects just so execute() can pull the pks back out, pass the scoped queryset straight to execute() and .update() on it.

src/oscar/apps/dashboard/catalogue/bulk_actions.py
```diff
@@ -0,0 +1,178 @@
...
+                product_id__in=override_ids,
+                **partner_filter,
+            ).update(
+                price=Case(
+                    *[
+                        When(product_id=pk, then=Value(p))
+                        for pk, p in specific.items()
+                        if pk in override_ids
+                    ],
+                    output_field=DecimalField(),
```
samar-hassan, line 150: plz add `default=F("price")`

src/oscar/static_src/oscar/js/oscar/dashboard.js
```diff
@@ -557,6 +557,134 @@ var oscar = (function(o, $) {
...
+                    });
+
+                    syncType(t);
+                });
+
+                // Sync the select all/across 
+                $('.select-all-children').on('change', function() {
+                    var pid = $(this).data('parent');
+                    $('input[name="selected_products"][data-parent="' + pid + '"]')
+                        .prop('checked', this.checked);
```
samar-hassan, line 676: This should toggle selections.

src/oscar/apps/dashboard/catalogue/bulk_actions.py
```diff
@@ -0,0 +1,178 @@
...
+            # Never trust the submitted partners as-is: a non-staff user must
+            # only ever be able to touch stockrecords of their own partners.
+            partners = partners.filter(pk__in=request.user.partners.all())
+            if not partners:
+                raise PermissionDenied(
+                    "You do not have permission to update prices for the"
+                    " selected partners."
+                )
+
+        partner_filter = {"partner__in": partners}
```
samar-hassan, line 135: unnecessary variable

src/oscar/apps/dashboard/catalogue/views.py
```diff
@@ -181,6 +192,224 @@ def apply_search(self, queryset):
...
+
+    intermediate_actions = ProductBulkActionMixin.intermediate_actions
+
+    max_display_products = 200
+
+    def get_cancel_url(self):
+        return reverse("dashboard:catalogue-product-list")
+
+    def get_success_url(self):
+        return reverse("dashboard:catalogue-product-list")
```
samar-hassan, line 212: Can we redirect to the same page, so that users can easily verify that the prices have updated?

src/oscar/apps/dashboard/catalogue/views.py
```diff
@@ -181,6 +192,224 @@ def apply_search(self, queryset):
...
+            qs = Product.objects.filter(
+                Q(pk__in=self._selected_ids) | Q(parent_id__in=self._selected_ids)
+            )
+            if not self.request.user.is_staff:
+                # Variants stocked by another partner must never be selectable.
+                # Parents aren't stock-bearing themselves, so leave them be.
+                user = self.request.user
+                qs = qs.filter(
+                    Q(structure=Product.PARENT)
+                    | Q(stockrecords__partner__users__pk=user.pk)
```
samar-hassan: plz verify if this filter is required in publicbulkaction, it might have been missed there previously

src/oscar/apps/dashboard/catalogue/views.py
```diff
@@ -181,6 +192,224 @@ def apply_search(self, queryset):
...
+                    "indent": False,
+                }
+            )
+            displayed += 1
+            selectable_displayed += 1
+
+        for parent in parents:
+            if displayed >= max_display:
+                break
+            children = list(parent.children.all())
```
samar-hassan, line 364: This looks inefficient and slow. See if you can fetch all children separately in a single query and map them to parents. Maybe a separate `get_children_queryset`.

mr-sphere: wouldnt a simple prefetch also do the trick?

src/oscar/apps/dashboard/catalogue/views.py
```diff
@@ -181,6 +192,224 @@ def apply_search(self, queryset):
...
+        Build the list of table rows and count how many selectable products
+        were included.
+        """
+        parents = self.get_parent_queryset()
+        standalone = self.get_standalone_queryset()
+
+        display_rows = []
+        displayed = 0
+        selectable_displayed = 0
+        max_display = self.max_display_products
```
samar-hassan: why do we have both `max_bulk_intermediate_ids` and `max_display_products`

src/oscar/apps/dashboard/catalogue/views.py
samar-hassan: yes, that should work

src/oscar/apps/dashboard/catalogue/views.py
mr-sphere: max_bulk_intermediate_ids is the maxmimum amount of ids you can select on the list page.
max_display_products is the maximum amount of products the form on the intermediate page is allowed to display.

i renamed max_display_products to max_displayable_products for a bit more clarity

src/oscar/apps/dashboard/catalogue/bulk_actions.py
mr-sphere: its used twice right below

src/oscar/apps/dashboard/catalogue/views.py
samar-hassan: why is max_bulk_intermediate_ids set to 500 and max_displayable_products to 200

src/oscar/apps/dashboard/catalogue/views.py
mr-sphere: thats going to make it difficult to delete the session key. im not sure if there is a good way around that.

src/oscar/apps/dashboard/catalogue/views.py
mr-sphere: max_bulk_intermediate_ids contains data that is not rendered. and max_displayable_products is.
max_displayable_products also contains children of parent products, so if you select 10 parents (which then go into max_bulk_intermediate_ids as 10 ids) then the intermediate step may have to render hundreds of products, depending on the amount of variants in those 10 parents.

samar-hassan (approved):

PR #4572, opened 2026-06-01
Fix/get applicable lines null range check

### LINKED ISSUE #4506: get_applicable_lines in AbstractBenefit crashes when self.range is NULL for custom benefits
I use a custom discount benefit in Django Oscar. By default, Oscar forbids setting a range for custom benefit classes. But default method `get_applicable_lines` assumes that a valid range is always present. 

It contains the logic:

```python
if range is None:
    range = self.range
```

If do not override `get_applicable_lines` this will crash the code with error `'NoneType' object has no attribute 'contains_product'`.

At list the code should check that `self.range` also not None. And log an error to warn about an issue in configuration.

In my case, I would like to have possibility to fill the Range filed together with custom benefit. In my case I cannot provide a range dynamically with `oscar.apps.offer.custom.create_range`. Or at least I don't understand how.

yachikadev:
When a custom benefit class is used without a range, `get_applicable_lines` 
crashes with `'NoneType' object has no attribute 'contains_product'`.

This fix adds a null check after `range = self.range` and raises 
`ImproperlyConfigured` with a clear error message instead of crashing silently.

Fixes #4506

### DIFF
--- a/src/oscar/apps/offer/abstract_models.py
+++ b/src/oscar/apps/offer/abstract_models.py
@@ -3,6 +3,7 @@ import csv
 import operator
 from decimal import ROUND_DOWN
 from decimal import Decimal as D
+from django.core.exceptions import ImproperlyConfigured
 
 from django.conf import settings
 from django.core import exceptions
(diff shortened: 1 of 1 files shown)


yachikadev:
@janheini please review

janheini:
Well this PR contains two unrelated commits from your other PR which you will need to remove.

Looks fine to me otherwise. But I can neither officially approve nor merge :)

yachikadev:
@janheini Thank you for the review! Since the unrelated commits have been removed and the fix looks good, could you help ping a maintainer who can officially review and merge this?

yachikadev:
Closing this PR as I'm no longer actively contributing to this project. Thanks for the review.

PR #4594, opened 2026-07-06
[Fix] Prevent deadlock between concurrent order placements

samar-hassan:
Two orders placed at the same time could deadlock while allocating stock: each `place_order` locked its basket's stock records in basket-line order (`all_lines()` is ordered by line pk), so two baskets sharing the same products in different line orders could acquire the row locks in opposite order and deadlock.

`place_order` now allocates stock in a deterministic **stock-record** order, so every concurrent placement takes the locks in the same order and they serialise instead of deadlocking.

### DIFF
--- a/tests/integration/order/test_stockrecord_locking.py
+++ b/tests/integration/order/test_stockrecord_locking.py
@@ -1,0 +1,131 @@
+import threading
+import time
+from unittest import skipUnless
+
+from django.contrib.auth.models import AnonymousUser
+from django.db import connection, connections
+from django.db.utils import OperationalError
+from django.test import TransactionTestCase
+
+from oscar.apps.checkout import calculators
+from oscar.apps.order.models import Order
+from oscar.apps.order.utils import OrderCreator
+from oscar.apps.shipping.methods import Free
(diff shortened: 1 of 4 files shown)


tests/integration/order/test_stockrecord_locking.py
```diff
@@ -0,0 +1,130 @@
...
+
+        errors = self._run_concurrently(worker, num_threads=4)
+
+        deadlocks = [
+            error
+            for error in errors
+            if isinstance(error, OperationalError)
+            and getattr(error.__cause__, "pgcode", None) == DEADLOCK_SQLSTATE
+        ]
+        self.assertEqual(deadlocks, [], "concurrent placements deadlocked")
```
joeyjurjens: The reason string no longer makes sense after you fixed the deadlock. I think the two things you should assert after you fixed it is:

errors = empty
4 orders created.

The whole "deadlocks =" and DEADLOCK_SQLSTATE can be gone now after u fixed it.

Maybe create some asserts regarding the allocated count too.

joeyjurjens (changes requested):

joeyjurjens (approved):

PR #4595, opened 2026-07-08
allow all headings in tinymce editor

mr-sphere:
liberation!


joeyjurjens (commented):
You're missing h6, so it's not "all". Also, do you think h1 is a good idea? Cause that can have SEO impact, normally h1 is things like product title / category tittle.

joeyjurjens (approved):

PR #4617, opened 2026-08-25
Fix switching an offer benefit/condition back from a pre-defined one

### LINKED ISSUE #3889: Cannot switch back from pre-defined benefit/condition to a non-pre-defined one
### Issue Summary

The current benefit and condition forms don't work properly when switching back to using a non-pre-defined benefit/condition from a pre-defined one.

### Steps to Reproduce

1. Create an offer with a non-pre-defined benefit or condition.
2. Go to the benefit/condition form for that offer and change it to a pre-defined one.
3. Go back to the form and try to change it back to a non-pre-defined one.  
  Instead of switching back, the benefit/condition is still referencing the previous pre-defined one.

#### Notes
- On the public sandbox, the form only has pre-defined benefits, but the saving logic is identical and the bug is also present in the `ConditionForm` as well:
    https://github.com/django-oscar/django-oscar/blob/04dd391c900537a61dbb0ae5250ca5c2df6bbc4b/src/oscar/apps/dashboard/offers/forms.py#L137-L143

- After switching to a pre-defined benefit/condition, the form also displays the other field data that had to be removed back in step 2.

### Technical details

* Reproducable on the public sandbox: https://latest.oscarcommerce.com/en-gb/dashboard/offers/

ChrisJr404:
Fixes #3889.

When an offer's benefit or condition is set to a pre-defined one, and you then edit it again to switch back to a non-pre-defined (custom) benefit/condition, the dashboard form is bound to the shared pre-defined instance. Saving it wrote the new range/type/value straight onto that shared row and left `proxy_class` set, so the offer kept referencing the pre-defined benefit/condition. It also mutated the shared instance for every other offer using it.

Now, when the form is switching away from a pre-defined benefit/condition (i.e. no custom benefit/condition selected but the bound instance has a `proxy_class`), it detaches from that instance and creates a fresh one, so the original is left untouched.

Tests: added `test_switch_from_custom_benefit_creates_new_benefit` and `test_switch_from_custom_condition_creates_new_condition` to `tests/integration/dashboard/test_offer_forms.py`, which assert a new non-pre-defined instance is returned and the shared pre-defined one is unchanged.


samar-hassan (approved):

samar-hassan:
@ChrisJr404 Thanks for your contribution! 🙌

PR #4618, opened 2026-08-28
Fix Unicode product slug redirect loop (#4599)

### LINKED ISSUE #4599: Possible redirect loop with Unicode product slugs in ProductDetailView
Hello,

I would like to report a possible issue with Unicode slugs in `ProductDetailView.redirect_if_necessary()`.

## Environment

- Python: 3.11.15
- Django: 5.1.3 (also tested with Django 4.2.30)
- django-oscar: 4.1

## Problem

When a product has a Unicode slug (for example Persian characters), the product detail page enters a redirect loop.

Example product slug:

`گوشی-موبایل-موتورولا-مدل-edge-60-fusion`

The generated URL from `get_absolute_url()` is already percent encoded:

`/catalogue/%DA%AF%D9%88%D8%B4%DB%8C-%D9%85%D9%88%D8%A8%D8%A7%DB%8C%D9%84_13/`

However, `ProductDetailView.redirect_if_necessary()` currently compares:

```python
expected_path = product.get_absolute_url()

if quote(expected_path) != quote(current_path):
    return HttpResponsePermanentRedirect(expected_path)
```

Since `expected_path ` is already encoded, applying `quote()` again causes double encoding:

Before:

`%DA%AF...`

After:

`%25DA%25AF...`

The browser request path contains the Unicode representation:

`/catalogue/گوشی-موبایل_13/`

so the comparison never matches and the view keeps returning a permanent redirect.

### Debug output

Example:

```
request.path:

/catalogue/گوشی-موبایل_13/


product.get_absolute_url():

/catalogue/%DA%AF%D9%88%D8%B4%DB%8C-%D9%85%D9%88%D8%A8%D8%A7%DB%8C%D9%84_13/
```

```
Comparison:

quote(expected_path):

/catalogue/%25DA%25AF...

quote(current_path):

/catalogue/%DA%AF...
```

### Workaround

I solved this locally by overriding the method in my Oscar catalogue fork and comparing decoded paths instead:

```python
from urllib.parse import unquote

def redirect_if_necessary(self, current_path, product):
    if self.enforce_parent and product.is_child:
        return HttpResponsePermanentRedirect(product.parent.get_absolute_url())

    if self.enforce_paths:
        expected_path = unquote(product.get_absolute_url())

        if expected_path.rstrip("/") != current_path.rstrip("/"):
            return HttpResponsePermanentRedirect(product.get_absolute_url())
```

This makes both sides Unicode strings before comparison.

### Question

Is this expected behavior for Unicode slugs, or should `redirect_if_necessary()` avoid quoting `expected_path because get_absolute_url()` already returns a URI-safe encoded path?

Thank you for your work on django-oscar.

keyanskv:
# Fix Unicode product slug redirect loop (#4599)

## Description

Fixes a possible redirect loop when a product has a Unicode slug.

When `ProductDetailView.redirect_if_necessary()` compares the current request path with `product.get_absolute_url()`, URL-encoding both paths with `quote()` can cause already percent-encoded Unicode characters to be encoded a second time.

For example:

```text
/catalogue/%DA%AF%D9%88%D8%B4%DB%8C-.../
```

can become:

```text
/catalogue/%25DA%25AF%25D9%258...
```

This causes the paths to be considered different even when they represent the same URL, resulting in repeated permanent redirects.

## Changes

* Updated `ProductDetailView.redirect_if_necessary()` in `src/oscar/apps/catalogue/views.py`.

  * Replaced `quote()` comparison with `unquote()` comparison.
  * Both expected and current paths are decoded before comparison.

* Updated `ProductCategoryView.redirect_if_necessary()` in `src/oscar/apps/search/views/catalogue.py`.

  * Applied the same decoded-path comparison to avoid inconsistent URL encoding behavior.

* Added a regression test in `tests/functional/catalogue/test_catalogue.py`.

  * Added `test_does_not_go_into_redirect_loop_with_unicode_slug`.
  * The test uses a Persian Unicode product slug and verifies that the product detail request returns HTTP 200 rather than entering a redirect loop.

## Testing

Ran:

```bash
DATABASE_PORT=5433 python -m pytest tests/functional/catalogue/test_catalogue.py -k unicode_slug
```

Result:

```text
1 passed, 18 deselected, 2 warnings
```

The warnings are unrelated to this change:

* `RemovedInDjango70Warning` for `select_related()` without arguments.
* A `http.cookiejar`/WebOb Unicode warning encountered while processing the Unicode request.

## Related Issue

Fixes #4599

## Checklist

* [x] Added a regression test for the reported issue
* [x] Tested the affected functional test
* [x] No unrelated changes included
* [x] Working tree is clean

### DIFF
--- a/scratch.py
+++ b/scratch.py
@@ -1,0 +1,2 @@
+def test():
+    pass
--- a/src/oscar/apps/voucher/abstract_models.py
+++ b/src/oscar/apps/voucher/abstract_models.py
@@ -204,7 +204,7 @@ class AbstractVoucher(models.Model):
         """
         is_available, message = False, ""
         if self.usage == self.SINGLE_USE:
-            is_available = not self.applications.exists()
+            is_available = self.num_orders == 0
             if not is_available:
                 message = _("This voucher has already been used")
         elif self.usage == self.MULTI_USE:
--- a/src/oscar/apps/voucher/abstract_models.py
+++ b/src/oscar/apps/voucher/abstract_models.py
@@ -1,7 +1,7 @@
 from decimal import Decimal
 
 from django.core import exceptions
-from django.db import models
+from django.db import models, transaction
 from django.db.models import Sum
 from django.utils import timezone
 from django.utils.translation import gettext_lazy as _
@@ -204,7 +204,7 @@ class AbstractVoucher(models.Model):
         """
         is_available, message = False, ""
         if self.usage == self.SINGLE_USE:
-            is_available = self.num_orders == 0
+            is_available = not self.applications.exists()
             if not is_available:
                 message = _("This voucher has already been used")
         elif self.usage == self.MULTI_USE:
@@ -248,12 +248,19 @@ class AbstractVoucher(models.Model):
         """
         Records a usage of this voucher in an order.
         """
-        if user.is_authenticated:
-            self.applications.create(voucher=self, order=order, user=user)
-        else:
-            self.applications.create(voucher=self, order=order)
-        self.num_orders += 1
-        self.save()
+        with transaction.atomic():
+            voucher = type(self).objects.select_for_update().get(pk=self.pk)
+            is_available, message = voucher.is_available_to_user(user)
+            if not is_available:
+                raise exceptions.ValidationError(message)
+
+            if user.is_authenticated:
+                voucher.applications.create(voucher=voucher, order=order, user=user)
+            else:
+                voucher.applications.create(voucher=voucher, order=order)
+            voucher.num_orders += 1
+            voucher.save()
+            self.num_orders = voucher.num_orders
 
     record_usage.alters_data = True
 
--- a/tests/integration/voucher/test_race_conditions.py
+++ b/tests/integration/voucher/test_race_conditions.py
@@ -32,7 +32,6 @@ from django.test import TestCase, TransactionTestCase
 from oscar.apps.voucher.models import Voucher
 from oscar.test.factories import OrderFactory, UserFactory, VoucherFactory
 
-
 # ---------------------------------------------------------------------------
 # Helpers
 # ---------------------------------------------------------------------------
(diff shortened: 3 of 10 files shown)


anooshiravan-G1:
Thanks for working on this and for linking it to #4599!

This looks like the issue I originally reported. Unfortunately, I’m not able to go back to the project/environment to test the fix right now, so I can’t personally verify it at the moment.

The regression test for the Unicode/Persian slug case looks useful, and I hope it helps with validating the fix.

src/oscar/apps/voucher/abstract_models.py
```diff
@@ -248,12 +248,19 @@ def record_usage(self, order, user):
         """
         Records a usage of this voucher in an order.
         """
-        if user.is_authenticated:
-            self.applications.create(voucher=self, order=order, user=user)
-        else:
-            self.applications.create(voucher=self, order=order)
-        self.num_orders += 1
-        self.save()
+        with transaction.atomic():
```
joeyjurjens, line 251: This is unrelated to the issue

tests/integration/voucher/test_race_conditions.py
```diff
@@ -0,0 +1,318 @@
+"""
```
joeyjurjens, line 1: This is unrelated to the issue

scratch.py
```diff
@@ -0,0 +1,2 @@
+def test():
```
joeyjurjens, line 1: This is unrelated to the issue

joeyjurjens (changes requested):

scratch.py
keyanskv: Thanks for the review. I realized that my branch included unrelated commits from the voucher race-condition work.

I've created a new clean branch based on `master` containing only the changes for #4599, and I'll submit the fix as a separate PR.

I'll close this PR to avoid mixing the two issues.

src/oscar/apps/voucher/abstract_models.py
keyanskv: Thanks for the review. I realized that my branch included unrelated commits from the voucher race-condition work.

I've created a new clean branch based on `master` containing only the changes for #4599, and I'll submit the fix as a separate PR.

I'll close this PR to avoid mixing the two issues.

tests/integration/voucher/test_race_conditions.py
keyanskv: Thanks for the review. I realized that my branch included unrelated commits from the voucher race-condition work.

I've created a new clean branch based on `master` containing only the changes for #4599, and I'll submit the fix as a separate PR.

I'll close this PR to avoid mixing the two issues.

PR #4620, opened 2026-09-04
Raise ImproperlyConfigured when a custom Benefit has no range

### LINKED ISSUE #4506: get_applicable_lines in AbstractBenefit crashes when self.range is NULL for custom benefits

wakqasahmed:
Custom benefit classes that don't set a range crash inside `get_applicable_lines` with `'NoneType' object has no attribute 'contains_product'`, since the method assumes `self.range` is always present. Oscar already forbids setting a range on some custom benefit types, so this is easy to hit.

This adds a check right after the existing `if range is None: range = self.range` fallback and raises `ImproperlyConfigured` with a message pointing at the actual problem, instead of the confusing AttributeError. Also added a regression test in `tests/integration/offer/test_benefit.py`.

Ran the full `tests/integration/offer/` suite locally (259 tests) plus the new test, all green.

Fixes #4506


samar-hassan:
@wakqasahmed Thanks for your contribution! 🙌

samar-hassan (approved):

PR #4622, opened 2026-09-08
Prevent voucher usage limit bypass under concurrent redemptions

### LINKED ISSUE #4580: [Security] CRITICAL voucher usage-limit race bypass + email-change account takeover + refund over-payment + partner IDOR

joeyjurjens:
Supersedes https://github.com/django-oscar/django-oscar/pull/4612

Fixes #4580


joeyjurjens:
Currently the UI does not show a clear error message. It just says something went wrong.

This could be solved, but not as clean as I'd like.

So for now; Not mergable yet.

PR #4624, opened 2026-09-11
Fix MultipleObjectsReturned when reordering a product with multiple option variants

### LINKED ISSUE #3789: Reorder a product with Option causes MultipleObjectsReturned exception
### Issue Summary

The request to `reorder` a basket line from the previous order crashes with `MultipleObjectsReturned` exception. 
This is caused by getting the basketline using product_id in code, while DB has two products with the same product_id but different options. 

### Steps to Reproduce

1. In dashboard, create an `Option`. Add the Option to Product Class Book.  
2. In shorefront, order a product `P1` with the Option. Then checkout and pay the order `O1`.
3. Add a product with the same product_id and Option_1 to basket. Then add another product with same product_id but Option_2 to basket. 
4. Go to order history and go to order `O1`.  Click `Reorder` button the product `P1`.  
5. Page crash with `MultipleObjectsReturned` exception. 

Stacktrace
```python
Traceback (most recent call last):
  File "django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "django/contrib/auth/decorators.py", line 21, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "django/views/generic/base.py", line 70, in view
    return self.dispatch(request, *args, **kwargs)
  File "django/views/generic/base.py", line 98, in dispatch
    return handler(request, *args, **kwargs)
  File "/django-oscar/src/oscar/views/generic.py", line 26, in post
    getattr(self, method_name)(model)
  File "/django-oscar/src/oscar/apps/customer/views.py", line 574, in do_reorder
    line_available_to_reorder, reason = line.is_available_to_reorder(
  File "/django-oscar/src/oscar/apps/order/abstract_models.py", line 773, in is_available_to_reorder
    basket_line = basket.lines.get(product=self.product)
  File "django/db/models/manager.py", line 85, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "django/db/models/query.py", line 439, in get
    raise self.model.MultipleObjectsReturned(

Exception Type: MultipleObjectsReturned at /en-gb/accounts/orders/100003/7/
Exception Value: get() returned more than one Line -- it returned 2!
```

#### Expected result:
* No crash. 

### Root Cause
In BasketLine model, the `product_id` and `basket_id` are not unique_together, so cannot use `.get` to retrieve the record. 
* Only the `basket_id` and `line_reference` is unique_together. 
* Comments for `line_reference` in class AbstractLine says: 
> This is to determine which products belong to the same line. 
We can't just use product.id as you can have customised products
which should be treated as separate lines. 

```python
File: src/oscar/apps/order/abstract_models.py
class AbstractLine(models.Model):
    def is_available_to_reorder(self, basket, strategy):
        if not self.product:
            return False, (_("'%(title)s' is no longer available") %
                           {'title': self.title})

        try:
            basket_line = basket.lines.get(product=self.product)
        except basket.lines.model.DoesNotExist:
            desired_qty = self.quantity
        else:
            desired_qty = basket_line.quantity + self.quantity
```

### Technical details

* Python version: 3.9.7
* Django version: 3.2.7
* Oscar version: master

wakqasahmed:
This fixes the `MultipleObjectsReturned` crash on reorder described in #3789.

`AbstractLine.is_available_to_reorder` looked up the matching basket line with `basket.lines.get(product=self.product)`, but basket lines aren't unique per product — they're unique per `(basket, line_reference)`, where `line_reference` also factors in options and stockrecord. So if a customer's basket already has two lines for the same product with different option values (e.g. two different engravings), that `.get()` raises `MultipleObjectsReturned` instead of finding (or not finding) the one line that actually matches.

I switched it to use `Basket.line_quantity(product, stockrecord, options)`, which already exists and does the correct product+options+stockrecord lookup via `_create_line_reference`, so this reuses the same logic baskets already use elsewhere instead of reinventing it.

Added a test (`LineReorderingTests` in `tests/integration/order/test_models.py`) that reproduces the exact scenario from the issue — a basket with two lines for the same product but different options — and asserts `is_available_to_reorder` no longer raises and correctly only counts the quantity of the matching option variant. I confirmed this test fails with `MultipleObjectsReturned` against the old code and passes against the fix (ran via `pytest tests/integration/order/test_models.py -k Reorder` against a Postgres 16 test DB).

Closes #3789


wakqasahmed:
Hi @samar-hassan @joeyjurjens — not sure who owns review for this one — it's been a little while, green and mergeable. Could you take a look or redirect me?

samar-hassan:
Thanks for finding the root cause — keying on `line_reference` via `line_quantity` instead of product is the right call. One issue before it's correct: it uses the order line's historical `self.stockrecord`, but a reorder uses the current one. `do_reorder` calls `add_product(product, qty, options)` (no stockrecord), and `add_product` resolves the live stockrecord via the strategy and builds the line_reference from that.

Two problems follow:

1. Crash: order.Line.stockrecord is null=True, SET_NULL. If it was deleted, line_quantity(self.product, None, options) raises `AttributeError` on None.id — and since `do_reorder` loops all lines, it 500s the whole reorder instead of skipping one.
2. Wrong result when the stockrecord changed: if it was deleted-and-replaced (or just recreated with a new id), the product is purchasable again and should reorder against the new stockrecord (what master does today). Here the historical stockrecord builds a line_reference that won't match the line add_product creates, so line_quantity returns 0 and undercounts.

The method already fetches the live stockrecord (PurchaseInfo = (price, availability, stockrecord)), so please use it:
```
result = strategy.fetch_for_product(self.product)
stockrecord = result.stockrecord
if stockrecord is None:
    return False, (_("'%(title)s' is no longer available") % {"title": self.title})
......
desired_qty = self.quantity + basket.line_quantity(self.product, stockrecord, options)
.....
```
This keeps the line_reference fix, matches the stockrecord `add_product` uses, reorders correctly against a replaced stockrecord, and doesn't crash when there's none.

Please add a case where the stockrecords differ (delete/recreate the stockrecord) asserting it doesn't raise and checks availability against the current stockrecord.

